PR #31, opened 2015-12-29
Ui external

matthiasbeyer:
Tracking branch for implementation to call external editor and edit file contents with it.


matthiasbeyer:
@homu r+

PR #32, opened 2015-12-29
Add command: show

matthiasbeyer:
Add a command "show" which is a module-independent way to print content of a file to the user.

Uses the modules behind the scenes to print the contents module-specific, but all the user has to know is the ID of a file.

Later this can also include search options to grep through the content of the files, or to grep through the header.


matthiasbeyer:
Waiting for #37

PR #34, opened 2015-12-29
Add some documentation

matthiasbeyer:
Still does not produce the documentation I want, though it basically works now.


matthiasbeyer:
This builds fine, though the documentation is not generated properly. @TheNeikos can you tell why?

matthiasbeyer:
@TheNeikos is 1647352 the right approach?

matthiasbeyer:
I assume so now.

@homu r+

PR #36, opened 2015-12-30
BM::command_list() add filter functionality

matthiasbeyer:
Adds missing filter functionality for BM subcommand `list`.


matthiasbeyer:
@homu r+

PR #37, opened 2015-12-30
Refactor filter building into types which can be chained up

matthiasbeyer:
State before was to generate functions which can be used as filter()
arguments.

State now is that there are types which can be lined up and be executed
as filters. And/Or/Not-Filters are provided and so are helper functions
to initialize objects of these filter types.

The BM module is adapted for this.

@TheNeikos can you have a look, please?


matthiasbeyer:
@homu r+

PR #42, opened 2016-01-02
ImagLogger can be configured through the CLI only now

matthiasbeyer:
The logger instance cannot be configured through the configuration file, as this is kinda bulky to use and we can log to debug much earlier now.


matthiasbeyer:
@homu r+

PR #43, opened 2016-01-02
Config refactor

matthiasbeyer:
Refactoring of the internal representation of the configuration file.

Also removes the possibility to set verbosity and debugging flag in the config file, as I consider this bad anyways and it allows us now to setup the logger earlier in the startup phase, which means more code which can use debugging output.


matthiasbeyer:
@homu r+

PR #49, opened 2016-01-02
Test store

matthiasbeyer:
Test store interface.

### DIFF
--- a/src/storage/mod.rs
+++ b/src/storage/mod.rs
@@ -451,5 +451,33 @@ mod test {
         finalize_store(store);
     }
 
+    #[test]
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@TheNeikos can you explain why this test sometimes fails with one failing test function, sometimes with two failing test functions and now and then succeeds?

TheNeikos:
That seems weird... I have to admit I don't know. Where are you writing per default?

matthiasbeyer:
Default is `~/.imag/store/`, though I'm using `/tmp/test-store` here.

PR #55, opened 2016-01-03
Fix: [warn]: Malformed Header, expected Array

matthiasbeyer:
We did not check whether we got the TAGS array in here.


matthiasbeyer:
@homu r+

PR #62, opened 2016-01-05
Module parser

matthiasbeyer:
Move the 

```
let parser = ...
```

Foo in each function to a module-global parser object in the module object.


matthiasbeyer:
@homu r+

PR #64, opened 2016-01-05
Add JsonPrinter for printing files Jsonified

### LINKED ISSUE #21: [output]: Output format: JSON
We should provide an output module so all output is in JSON.

The header/content should be wrapped:

```JSON
[ ...
{ "HEADER": <header>,
  "CONTENT": <content>,
}
...]
```
The `<header>` contains the header of an file in JSON format, of course. The `<content>` contains the content part of the file in plain text. All of this is in an Array of Objects.

matthiasbeyer:
Closes #21

### DIFF
--- a/etc/cli.yml
+++ b/etc/cli.yml
@@ -326,6 +326,13 @@ subcommands:
                         required: false
                         takes_value: false
 
+                    - json:
+                        long: json
+                        short: j
+                        help: Show notes as json
+                        required: false
+                        takes_value: false
+
                     - id:
                         long: id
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Not tested yet, so I won't merge this.

matthiasbeyer:
Works, though this is a really naiive implementation and does not yield correct results/json for files which contain a single single quote (`'`) or a single double quote (`"`), for example.

This will remain unmerged until this feature is actually needed. This is just a showcase how one would implement it.

PR #65, opened 2016-01-06
Remove parser variable from Notes::command_show()

matthiasbeyer:
Was still there because of the merge order.


matthiasbeyer:
@homu r+

PR #66, opened 2016-01-06
Report success

matthiasbeyer:
Report success (as simple "Ok"/"Error" string) to the user, if the user wants that.


matthiasbeyer:
@homu r+

PR #70, opened 2016-01-07
Paper setup

matthiasbeyer:
This branch aims for a paper setup so we can start writing down where this project should go.

The paper build is not yet included in travis, though I'd really like to do that, so @TheNeikos will you have a look? I'd like to see something like `if changes in doc/, then cd && make; else cargo...` - is this possible?

If not, how to proceed?


matthiasbeyer:
dfc1262 represents my idea of how we could do this. Lets see whether it builds or not...

matthiasbeyer:
It fails because the version of `pandoc` is too old on travis. I need `1.10+` and we have `1.9` on travis. Let's see whether we are able to install newer versions in travis...

matthiasbeyer:
@TheNeikos It seems I need `sudo` now as I need to install a new pandoc version. I'm not sure what effects this has, but I guess because travis builds this on another infrastructure now, the builds take longer.

Any ideas how to optimize?

matthiasbeyer:
Theoretically, I'm ready for squashing.

TheNeikos:
I don't think this is a good idea. Dropping the container infrastructure from travis means that it will run seriously slow and I don't believe it is worth the trade off of having it build your markdown.

matthiasbeyer:
@TheNeikos So what would you do? Exclude the doc from the travis builds? Just build them locally?

TheNeikos:
I would build it locally, since it is markdown we can still comment on it on Github as it should be rendered somewhat correctly!

matthiasbeyer:
Yeah... It's pandoc markdown, so it has embedded latex and such, but I think this is okay. I will rewrite the branch to exclude the travis builds then...

matthiasbeyer:
@TheNeikos But one thing we _could_ do is check whether the PR changes are only in the `doc/` subfolder, and don't call cargo at all if there are only changes in this folder, right? Would speed up travis for only-doc-changing-PRs!

TheNeikos:
Yup, that should work

matthiasbeyer:
@TheNeikos Not sure about 1730717, as I'm not sure whether we want to build with cargo if the `travis.yml` changes. I'm rather sure we don't want that commit, right?

TheNeikos:
Hm, I think you should keep the `.travis.yml` in. If anything, it won't change often so it shouldn't really be an issue

matthiasbeyer:
Still nice to know that the build on 1730717 succeeded as expected, so my script seems to work.

I will try to fixup the math setup now and then this seems to be ready for squash+merge.

TheNeikos:
Awesome! :+1:

matthiasbeyer:
This seems to be ready. I compiled the `pandoc-reference` example and it worked for me, with pandoc `1.15.2.1`.

TheNeikos:
Compiles with pandoc 1.16 over here as well

matthiasbeyer:
Did you include the reference file by moving it to (for example) `src/0001.md`, so you see the example code?

TheNeikos:
Just did, and it works just as well

matthiasbeyer:
@homu r+

matthiasbeyer:
Just for future reference:

If required, I might be able to integrate `graphviz` or `mscgen` into the pandoc workflow, if this is neccessary. This way we can build the paper with scripts graphics to some extend. If everything fails, I could also integrate `gnuplot` or `metapost`, I guess.

I won't do this for now, as I don't think we need graphics.

PR #71, opened 2016-01-08
doc/default.nix: List dependencies for building the paper explicitely

matthiasbeyer:
Alters the `default.nix` in the `doc/` folder to list the (latex-)dependencies for building the paper explicitly, rather than listing some random collections which may or may not contain the required packages.

Just relevant for nix(os) users. Anyways helpful for others, as they know which latex packages to install for building the doc.


matthiasbeyer:
@homu r+

PR #88, opened 2016-01-09
Init rewrite

matthiasbeyer:
Start the rewrite.

@TheNeikos tell me what you think about the structure I create in this PR.


matthiasbeyer:
@TheNeikos , @neithernut more libraries to divide into? Shall I also create the appropriate subdirectories for the binaries (the frontends for the libs which have one)?

Should I maybe even change the structure to:

```
/
/store
/store/lib/
/store/lib/src/
/store/lib/Cargo.toml
/store/bin/
/store/bin/src/
/store/bin/Cargo.toml
```

Or just (as initially thought) create `./imag-store/` as binary crate directory? I would vote for the latter one, as the more nesting levels, the more complexity in the project setup, so I'd go for the flat approach here.

neithernut:
More libraries will emerge automatically if needed. The structure looks good.

TheNeikos:
The structure is correct and should be sufficient at first, however you will have to be careful once you start having dependencies.

Let's imagine `imagrt` depends on `imagutil`, how do you resolve the dependency?
There are two ways:

```
[dependencies.imagutil]
path = '../libimagutil'
```
or
```
[dependencies.imagutil]
version = "<1"
```

They are not compatible though, and if you plan on creating crates on crates.io you will run into issues regarding this.

There are two ways to resolve this:

1. If there will be no dependencies between these crates, this problem doesn't apply.
2. Do not use path dependencies, but rather always upload the latest version to crates, thus allowing the other crates to build. This however can be quite tedious if debugging cross-crate.

matthiasbeyer:
@TheNeikos How'd you do it? Why am I not allowed to

```
[dependencies.imagutil]
path = '../libimagutil'
version = "<1"
```

I'd vote against putting things on crates.

TheNeikos:
The problem with that is simple, you may get conflicting input in a system that was designed to be precise and easily understood.

What if you bump the version above 1 in one of the crates? What is supposed to happen? There are two outcomes: One gets ignored, but which one? You would have to make a pick. You get an error. I'm not sure what cargo does in this instance, going to check now.

matthiasbeyer:
@TheNeikos So what is the solution to our problem? Using paths only and trying not to break things? I mean, we have CI, so that shouldn't be that hard, right?

TheNeikos:
I might have not been all to clear about what I was trying to say, so here goes hopefully a better version:


1. You have one big crate called for example `imag` it would have `imag::rt` for example for runtime stuff, under the hood it is actually composed of `libimagrt`, `libimagutil` etc...

2. You create a seperate *crate* for each of the given parts and publish them and depend on those.

TheNeikos:
The structure right now:

```
neikos ~/p/r/imag ((5784142...))> tree
.
├── Cargo.lock
├── Cargo.toml
├── default.nix
├── doc
│   ├── bin
│   │   ├── html
│   │   │   └── index.html
│   │   └── pdf
│   │       └── paper.pdf
│   ├── default.nix
│   ├── Makefile
│   ├── README.md
│   ├── src
│   │   ├── 0000.md
│   │   ├── 0001.md
│   │   └── pandoc-reference
│   └── templates
│       ├── default.html5
│       └── default.latex
├── etc
│   └── cli.yml
├── libimagmodule
│   ├── Cargo.lock
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── libimagrt
│   ├── Cargo.lock
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── libimagstore
│   ├── Cargo.lock
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── libimagutil
│   ├── Cargo.lock
│   ├── Cargo.toml
│   └── src
│       └── lib.rs
├── LICENSE
├── README.md
└── src
    └── main.rs

16 directories, 29 files
```

I think this looks good

matthiasbeyer:
@homu r+

Issue #83, opened 2016-01-09
Rewrite Runtime


matthiasbeyer:
The `Runtime` type could be reused almost completely, anyways the `CLI` type would be partly rewritten, which is what the `Runtime` type depends on.

- [ ] CLI-Type rewrite
  - [ ] Flag handling
    - [ ] verbosity
    - [ ] debugging
  - [ ] configuration file setter
- [ ] Configuration file parsing (Move to TOML)
- [ ] Should hold the Store interface object

PR #89, opened 2016-01-10
Libimagrt init

matthiasbeyer:
Initial code for `libimagrt`.


matthiasbeyer:
Targets #83

neithernut:
Just quickly read over this PR.
About `Runtime::init_logger()`: why not just have a factory of the logger inside `Runtime`, which returns a logger and performs initialization on the first request?

matthiasbeyer:
Because you don't need to. The `log` crate sets a global logger and you can then use the `debug!()`, `info!()` (etc) macros...

neithernut:
This would still require the user to "do the right thing" and call `init_logger()` at some point.

neithernut:
Maybe this should be called during the config/argument parsing automatically.

matthiasbeyer:
Yeah, we should do that ASAP. Though, I'm not sure whether you want this to happen implicitely, that's why I exported it as a function.

libimagrt/src/configuration.rs
```diff
@@ -0,0 +1,77 @@
...
+
+pub struct Configuration {
+    pub verbosity: bool,
+    pub editor: Option<String>,
+    pub editor_opts: String,
+}
+
+impl Configuration {
+
+    pub fn new(rtp: &PathBuf) -> Option<Configuration> {
```
matthiasbeyer: We might want to implement `Default` instead of returning an Option and return the `default()` if the file cannot be found.

neithernut: Yep, this would make sense.

matthiasbeyer: Would even make more sense to implement `Default` but anyways return `Option<Configuration>`, so the callee would do `Configuration::new(...).unwrap_or(Configuration::default())`. Is less convenience for the callee, but more flexibility and control.

matthiasbeyer:
I will do some rebasing now.

matthiasbeyer:
Waiting on #119.

libimagrt/src/configuration.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        xdg_basedir::get_data_home().map(|data_dir| gen_vars(data_dir, variants.clone(), &modifier))
+                                    .unwrap_or(vec![]),
+    ].iter()
+        .flatten()
+        .filter(|path| path.exists())
+        .map(|path| from_file(&path).ok())
+        .filter(|loaded| loaded.is_some())
+        .nth(0)
+        .map(|inner| inner.unwrap())
```
TheNeikos: What will happen if the iter is empty?

matthiasbeyer: `None`?

matthiasbeyer:
Review please.

TheNeikos:
I see nothing wrong so far

matthiasbeyer:
> so far

does it mean I can merge? Or does it mean that there are things missing or you'd expect more things?

neithernut:
Additional features may still be implemented afterwards. You get an OK from me.

matthiasbeyer:
@neithernut ping me when you're done.

neithernut:
The interface looks acceptable. However, after reading over it again I wonder:
Why did you inline the "error" module? And why didn't you put `Runtime` in it's own module?

matthiasbeyer:
`Runtime` is the point of the whole library, so it is exported as-is.
`configuration` is a sub-module to that, which is for config file parsing. `error` is a sub-module to `configuration` which handles errors in the configuration only.

neithernut:
> `Runtime` is the point of the whole library, so it is exported as-is.

...and `store` is the point of the whole `libstore`. So isn't this a bit inconsistent?

matthiasbeyer:
@TheNeikos , @neithernut Review again pls. I want to merge this, then we can start parallelizing implementation.

TheNeikos:
Looks good to me

matthiasbeyer:
@homu r+

Issue #76, opened 2016-01-09
Redefine the Store interface


matthiasbeyer:
- [ ] Create
- [ ] Read
- [ ] Update
- [ ] Delete
- [ ] exported types

Each of these will be tracked in own issues:

#77 , #78 , #79 , #80 and #81

matthiasbeyer:
@neithernut Feel free to submit PRs or comments on this and the related issues for defining the store {file format, interface} with me.

PRs to the document are welcome as well.

matthiasbeyer:
When we implement the store interface, we should take multithreading into account and export the types appropriately.

Changes to the store should be done as transactions and the store should have an internal cache, writing to FS at the latest possible time.

TheNeikos:
If you put the issue number into the checklist to their corresponding number it should get checked once that issue is closed!

neithernut:
Currently thinking about this, but I currently lack time. You can assign this to me, if you want (and blame me if I didn't come up with anything yet on Thursday).

matthiasbeyer:
@neithernut I cannot assign you, as github does not offer this as an option for non-organization repositories, at least not as long as I don't give you push access. And of course I won't give you push access.

neithernut:
Yep, I forgot...
Well, I'll try to take some time to work on this issue.

neithernut:
So after thinking about this, I hit some things which definitely should be considered when defining an interface:

### File-system synchronization

We don't want to allow messing things up when running multiple `imag` commands in parallel. Hence, we definitely want to `flock()` the file when altering it, ensuring that the file experiences only one edit at a time. Locking it when doing a read is also generally a good idea. Alternatively, one can use a "shared lock" when retrieving an item, knowing it will only be read.
Now for read-only access of items, it would be nice if we could, at some point, release the file, leaving us with a read-only version of the item. Use-cases include, of course, commands which will retrieve some information without altering the item itself (most commands, I guess), without the chance of reading the file while someone else is writing to it.
This could be done by defining two traits: one trait describing read-only items and one trait, which extends (not quite sure about the rust semantics) the first one by read-write-access and a function which will create a read-only item, destroy the read-write item and release the file, preferably in that order.

### In-process locking

[...]

PR #90, opened 2016-01-10
Define file format

matthiasbeyer:
Related to #76.

@TheNeikos , @neithernut Would like to hear suggestions.

### DIFF
--- a/doc/src/02000-store.md
+++ b/doc/src/02000-store.md
@@ -38,3 +38,23 @@ configured in the header of the file.
 
 ## File organization {#sec:thestore:fileorganization}
 
+The "Entries" are stored as files in the "Store", which is a directory the
+user has access to.
+The store MAY exist in the users Home-directory or any other directory the
+user has Read-Write-Access to.
+
+The Path of each File is shown as absolute path in this paper, while the root
+is always the store directory.
+This Path is named "Storepath".
+So if the store exists in `/home/user/store/`, a file with the Storepath
+`/example.file` is (on the filesystem) located at
+`/home/user/store/example.file`.
+
+A Storepath contains one predefined part: The module name of the Module the
+Entry belongs to, in lowercase letters.
+So if a Module named "ExampleModule" stores a file in the Store, the Storepath
+for a file with the name "example.file" is "/examplemodule/example.file".
+
+Any number of subdirectories MAY BE used, so creating folder hierarchies is
+possible and valid.
+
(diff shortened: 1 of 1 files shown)


neithernut:
Having an `imag` section is a good idea, but maybe we should explicitly allow modules to store metadata in other sections, _if_ that data is of general interest to a user.

TheNeikos:
@neithernut Sections can be nested arbitrarily deep, so a module would be free to for example do this:

```toml
[imag]
id = 'asdasdfasdf12334555'
module = 'random_names_generator'

[imag.names_generated]
proper = ["Suzy", "Allen"]
first = []
```

matthiasbeyer:
@neithernut what @TheNeikos suggested is exactly the way I would do it. I'll describe it in the document, as it seems this should be stated...

neithernut:
I do know that you TOML supports nesting, but it seems a bit weird if general purpose data is stored under the `imag` section. Then again, most users will probably use a tool for extracting metadata anyway...

matthiasbeyer:
No, I think this is a good way to do it. At least, it is not _general_ purpose data, it is _module specific_ data, so it makes sense to me to store it in a section `imag.<modulename>`.

neithernut:
Still not quite sure, but for the sake of progress I agree to use this convention, at least for now.

TheNeikos:
We could reserve a [content] area, thus not allowing a 'content' module, making that the default content section, and allowing other modules thus to make their own section, not inside imag.

matthiasbeyer:
@TheNeikos You mean that the content of a file goes into the `[content]` section of the `TOML`-Part?

I propose a structure like this in this PR:

```
---
[imag]
<bla>
[imag.somemodule]
<module specific bla>
---

<content>
```

So the content has its own area in this case... (`content` is user input in this case. For some modules there is no user input, of course). I'm not quite sure what you mean with your comment.

TheNeikos:
Ah ok, in that case then, both of our arguments do not apply I think.

matthiasbeyer:
> both of our arguments

@TheNeikos what `both` arguments? You mean the one from @neithernut on "general purpose data"?

Anyways, do you ACK this?

TheNeikos:
Please do add an example, since there was a misunderstanding on my part, and I think with an example that would have been avoided.

TheNeikos:
Alright, now, I think that we can drop the imag prefix for the tables, since the content will always be seperate. I see no reason for it being nested from the get go.

matthiasbeyer:
@TheNeikos see 535c2f4

TheNeikos:
That seems good to me!

:+1:

matthiasbeyer:
@homu r+

PR #95, opened 2016-01-12
Init libstore

matthiasbeyer:
Initial interface proposal for the store library.

@TheNeikos , @neithernut RFC.

### DIFF
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -10,7 +10,6 @@ use std::io::Error as IOError;
 
 #[derive(Clone)]
 pub enum StoreErrorType {
-    Unknown,
     IdNotFound,
     OutOfMemory,
     // maybe more
@@ -28,7 +27,6 @@ impl<'a> From<&'a StoreErrorType> for String {
 
     fn from(e: &'a StoreErrorType) -> String {
         match e {
-            &StoreErrorType::Unknown     => String::from("<Unknown>"),
             &StoreErrorType::IdNotFound  => String::from("ID not found"),
             &StoreErrorType::OutOfMemory => String::from("Out of Memory"),
         }
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -8,7 +8,7 @@ use std::convert::From;
 
 use std::io::Error as IOError;
 
-#[derive(Clone)]
+#[derive(Clone, Copy)]
 pub enum StoreErrorKind {
     IdNotFound,
     OutOfMemory,
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -58,37 +58,24 @@ impl Display for StoreErrorType {
 
 pub struct StoreError {
     err_type: StoreErrorType,
-    expl: Option<&'static str>,
+    expl: &'static str,
     cause: Option<Box<Error>>,
 }
 
 impl StoreError {
 
-    pub fn new() -> StoreError {
+    pub fn new(errtype: StoreErrorType, expl: &'static str, cause: Option<Box<Error>>)
+        -> StoreError
+    {
         StoreError {
-            err_type: StoreErrorType::Unknown,
-            expl:  None,
-            cause: None,
+            err_type: errtype,
+            expl:  expl,
+            cause: cause,
         }
     }
 
     pub fn err_type(&self) -> StoreErrorType {
         self.err_type.clone()
-    }
-
-    pub fn with_type(mut self, t: StoreErrorType) -> StoreError {
-        self.err_type = t;
-        self
-    }
-
-    pub fn with_expl(mut self, e: &'static str) -> StoreError {
-        self.expl = Some(e);
-        self
-    }
-
-    pub fn with_cause(mut self, e: Box<Error>) -> StoreError {
-        self.cause = Some(e);
-        self
     }
 
 }
@@ -107,9 +94,7 @@ impl Display for StoreError {
 
     fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
         let e : String = self.err_type.clone().into();
-        try!(write!(fmt, "[{}]: {}",
-                    e,
-                    self.expl.unwrap_or("")));
+        try!(write!(fmt, "[{}]: {}", e, self.expl));
         Ok(())
     }
 
@@ -118,7 +103,7 @@ impl Display for StoreError {
 impl Error for StoreError {
 
     fn description(&self) -> &str {
-        self.expl.unwrap_or("")
+        self.expl
     }
 
     fn cause(&self) -> Option<&Error> {
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -9,20 +9,20 @@ use std::convert::From;
 use std::io::Error as IOError;
 
 #[derive(Clone)]
-pub enum StoreErrorType {
+pub enum StoreErrorKind {
     IdNotFound,
     OutOfMemory,
     // maybe more
 }
 
-fn store_error_type_as_str(e: &StoreErrorType) -> &'static str {
+fn store_error_type_as_str(e: &StoreErrorKind) -> &'static str {
     match e {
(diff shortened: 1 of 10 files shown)


neithernut:
Before looking at it in detail: I'll write down some results of my thinking in #76, which _may_ have some significant impact.

matthiasbeyer:
@neithernut I don't intend to merge this without heavy discussion and detailed review. So yes, please go ahead!

neithernut:
I commented in said issue. Feel free to comment.

matthiasbeyer:
@TheNeikos Implemented your suggestions from #98, feel free to comment my code in this PR from now on.

matthiasbeyer:
@TheNeikos review pls!

libimagstore/src/error.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
+        try!(write!(fmt, "{}", store_error_type_as_str(self)));
+        Ok(())
+    }
+
+}
+
+pub struct StoreError {
+    err_type: StoreErrorType,
+    expl: &'static str,
```
TheNeikos: Why keep the pointer in the struct? You can just call `store_error_type_as_str` to get the `&'static str`

matthiasbeyer: This code is already removed, your comment is already outdated.

libimagstore/src/error.rs
```diff
@@ -0,0 +1,95 @@
...
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Error as FmtError;
+use std::clone::Clone;
+use std::convert::From;
+
+use std::io::Error as IOError;
+
+#[derive(Clone)]
```
TheNeikos: Also make it copy! This way you don't have to clone it further down.

libimagstore/src/error.rs
```diff
@@ -0,0 +1,96 @@
...
+#[derive(Copy)]
```
TheNeikos: Please write it like this:  `#[derive(Copy, Clone)]`

libimagstore/src/error.rs
```diff
@@ -0,0 +1,95 @@
...
+}
+
+fn store_error_type_as_str(e: &StoreErrorKind) -> &'static str {
+    match e {
+        &StoreErrorKind::IdNotFound  => "ID not found",
+        &StoreErrorKind::OutOfMemory => "Out of Memory",
+    }
+}
+
+impl Debug for StoreErrorKind {
```
TheNeikos: I think the standard debug is more than enough, when I develop I would rather see the Typename rather than an explanation string.

libimagstore/src/error.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub fn err_type(&self) -> StoreErrorKind {
+        self.err_type.clone()
+    }
+
+}
+
+impl Debug for StoreError {
+
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
```
TheNeikos: Same here

TheNeikos:
LGTM!

matthiasbeyer:
@TheNeikos Thanks.

We should argue on the CRUD-Interface. The point you made in #98 sounds good to me and I guess we could change the interface to something like this:

```rs
fn read<P>(&self, &P) -> Result<Arc<RwLock<Entry>>
    where P: Into<PathBuf>;
```

For example. We also might discuss about the idea of having a wrapper around `RwLock` which does filesystem locking, as I played around with in https://github.com/matthiasbeyer/playground.rs/blob/flocking/src/flocking.rs.

libimagstore/src/fsstore.rs
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+}
+
+impl Drop for FSStore {
+
+    /**
+     * Unlock all files on drop
+     *
+     * TODO: Error message when file cannot be unlocked?
```
TheNeikos: You cannot return error messages in a drop, but you can panic.

TheNeikos:
I see no reason to make it a Into<PathBuf>, why would you?

libimagstore/src/fsstore.rs
matthiasbeyer: I mean to _print_ an error message, not to return one.

matthiasbeyer:
@TheNeikos 

> I also suggest to use a custom trait for the id, which could be called Id
> The idea is that you have a trait that you can use to convert into whatever you need. Thus you can easily add more ways of searching for files. It would implement a single method called matches. It takes a Lazy File/Filename as it's arguments and would allow one to implement it for &'a str or &'a String to match against the filename, or a closure for more advanced matching.

I was thinking of alternatives to what you proposed in the quote. But after re-reading, I guess I have to confess that I did not understand how you'd design this... We do not need a filter/matcher, but an identifier, so why would we need a trait which gives us the possibility to filter on something?

matthiasbeyer:
@TheNeikos I would make it into a trait, so we keep flexibility. A `PathBuf` might be sufficient, though.

libimagstore/src/fsstore.rs
TheNeikos: And then ignore it? What happens to the data then? Does it get silently not written? Or is the file never unlocked? Or only once the process locking it exits? What happens if you try to lock it again afterwards?

TheNeikos:
Perhaps I misunderstood, so let's me ask this: Retrieve gets a specific file, but through what mechanism? Exact path? Id? Metadata?

libimagstore/src/fsstore.rs
matthiasbeyer: Yep, you are right. I'm not sure how common unlocking-problems are, so I'm not sure whether we actually want to `pandic!()` here.

Though if we agree on using something like I proposed (with my playground.rs/flocking), this piece of code would get removed anyways.

matthiasbeyer:
In our current model: `ID == exact path`, so no, not metadata. :smile:

neithernut:
Yep, but you're screwed as soon as you add segmentation of items through subfolders.

TheNeikos:
Okay, in that case a PathBuf is sufficient.

matthiasbeyer:
Come _on_ now. We agreed on this days ago. A file is identified by its path and links work through pathes. If you have better ideas, go ahead and propose something better which offers the same level of convenience.

neithernut:
Ok, sry. I have both a daytime job now and work on my hands in my free time, because someone else fucked up just before going on vacation. I just didn't get the chance to keep track of everything.

matthiasbeyer:
@neithernut Sorry, it was not my intension to argue in an aggressive way or something. The point is, with pathes, the user can simply link content like this:

```md
some text in [another file in the wiki](/wiki/some_topic_folder/foo)
```

With IDs, the user would need to know the ID of the content he wants to link, this way you could almost use editor-autocompletion, you have a link which you can read (when printing the content in plaintext) where it points to and everything.

If the target file gets moved, we could auto-update the links - this is less effort (in manner of IO, Disk access and so on) and more convenience for the user than keeping track of IDs all the time. We could even export a library+binary for exactly this: Changing URIs in `$content` from `a->b`, where `$content in {Markdown, Textile, WhateverElse}`. URIs might also appear in the header part of files (for no-user-input-content things like for example todo-lists), of course. We need a set of utilities (speaking of binaries) to work with the headers anyways, so that's a thing which would go into this bin+lib bundle.

neithernut:
Ok, I agree. I'd propose we wait for @TheNeikos to come up with a suitable id-type (which may be, in fact, be based on paths) and incorporate those in our PRs.

TheNeikos:
I would still go with a Trait that compiles to PathBuf, this way you could easily namespace things for example. Or use strings (because PathBuf::new(), etc is annoying) etc... 

Is this something wanted? I wouldn't mind implementing it.

neithernut:
Yes. I think we would benefit from it.

matthiasbeyer:
Closed in favour of #99

PR #97, opened 2016-01-14
travis: exit(0) if there are only changes in ./doc

matthiasbeyer:
Optimize the CI-time for `./doc` changes by not doing anything if there are only changes in `./doc`.

@TheNeikos careful review would be appreciated!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,17 @@ matrix:
     allow_failures:
         - rust: nightly
 
+before_install:
+    - |
+      c=$(git diff $TRAVIS_BRANCH..$TRAVIS_COMMIT --name-only | cut -d "/" -f 1 | uniq)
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
It seems that the `before_install` does not run before the installation of the deps. @TheNeikos ideas?

TheNeikos:
https://travis-ci.org/matthiasbeyer/imag/jobs/102321928#L176

?

matthiasbeyer:
Yes, but the install phase is already over at that point...

TheNeikos:
Uhm, no? What you see beforehand is the Rust configuration.

matthiasbeyer:
So you think I can merge?

TheNeikos:
I think so

matthiasbeyer:
@homu r+

PR #98, opened 2016-01-16
Init libstore alt

neithernut:
This alternative version to @matthiasbeyer's pr #95 demonstrates some design points brought up in the store interface discussion. The PR is based on work by @matthiasbeyer.
Relates to #76.

### DIFF
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -1,0 +1,74 @@
+use std::error::Error;
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Error as FmtError;
+use std::clone::Clone;
+
+use std::io::Error as IOError;
+
+pub struct StoreError {
+    name: Option<&'static str>,
+    expl: Option<&'static str>,
+    cause: Option<Box<Error>>,
+}
+
+impl StoreError {
+
+    pub fn new() -> StoreError {
+        StoreError {
+            name: None,
+            expl:  None,
+            cause: None,
+        }
+    }
+
+    pub fn with_name(mut self, n: &'static str) -> StoreError {
+        self.name = Some(n);
+        self
+    }
+
+    pub fn with_expl(mut self, e: &'static str) -> StoreError {
+        self.expl = Some(e);
+        self
+    }
+
+    pub fn with_cause(mut self, e: Box<Error>) -> StoreError {
+        self.cause = Some(e);
+        self
+    }
+
+}
+
+impl Debug for StoreError {
+
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
+        try!(write!(fmt, "[{:?}]: {:?}, caused: {:?}", self.name, self.expl, self.cause));
+        Ok(())
+    }
+
+}
+
+impl Display for StoreError {
+
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
+        try!(write!(fmt, "[{}]: {}",
+                    self.name.unwrap_or("StoreError"),
+                    self.expl.unwrap_or("")));
+        Ok(())
+    }
+
+}
+
+impl Error for StoreError {
+
+    fn description(&self) -> &str {
+        self.expl.unwrap_or("")
+    }
+
+    fn cause(&self) -> Option<&Error> {
+        self.cause.as_ref().map(|e| &**e)
+    }
+
+}
+
--- a/libimagstore/src/single_use_lock.rs
+++ b/libimagstore/src/single_use_lock.rs
@@ -1,0 +1,5 @@
+pub trait SingleUseLock<T> {
+    fn access(&self) -> &T;
+    fn unlock(self) -> T;
+}
+
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -3,6 +3,7 @@ use std::string;
 
 pub use entry::Entry;
 pub use error::StoreError;
+pub use single_use_lock::SingleUseLock;
 
 pub type Result<T> = RResult<T, StoreError>;
 pub type LockedEntry = SingleUseLock<Entry>;
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -1,0 +1,17 @@
+use std::result::Result as RResult;
+use std::string;
+
+pub use entry::Entry;
+pub use error::StoreError;
+
+pub type Result<T> = RResult<T, StoreError>;
+pub type LockedEntry = SingleUseLock<Entry>;
+
+pub trait Store {
+    fn create(&self, entry : Entry) -> Result<()>;
+    fn retrieve(&self, id : string) -> Result<LockedEntry>;
+    fn retrieve_copy(&self, id : string) -> Result<Entry>;
+    fn update(&self, LockedEntry) -> Result<()>;
+    fn delete(&self, id : string) -> Result<()>;
+}
+
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -12,7 +12,7 @@ pub trait Store {
     fn create(&self, entry : Entry) -> Result<()>;
     fn retrieve(&self, id : String) -> Result<LockedEntry>;
     fn retrieve_copy(&self, id : String) -> Result<Entry>;
-    fn update(&self, LockedEntry) -> Result<()>;
+    fn update(&self, &LockedEntry) -> Result<()>;
     fn delete(&self, id : String) -> Result<()>;
 }
 
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -1,5 +1,5 @@
 use std::result::Result as RResult;
-use std::string;
+use std::string::String;
 
 pub use entry::Entry;
 pub use error::StoreError;
@@ -10,9 +10,9 @@ pub type LockedEntry = SingleUseLock<Entry>;
 
 pub trait Store {
     fn create(&self, entry : Entry) -> Result<()>;
-    fn retrieve(&self, id : string) -> Result<LockedEntry>;
-    fn retrieve_copy(&self, id : string) -> Result<Entry>;
+    fn retrieve(&self, id : String) -> Result<LockedEntry>;
+    fn retrieve_copy(&self, id : String) -> Result<Entry>;
     fn update(&self, LockedEntry) -> Result<()>;
-    fn delete(&self, id : string) -> Result<()>;
+    fn delete(&self, id : String) -> Result<()>;
 }
 
--- a/libimagstore/Cargo.lock
+++ b/libimagstore/Cargo.lock
@@ -1,4 +1,20 @@
 [root]
 name = "libimagstore"
 version = "0.1.0"
+dependencies = [
+ "toml 0.1.25 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "rustc-serialize"
+version = "0.3.16"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+
+[[package]]
+name = "toml"
+version = "0.1.25"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "rustc-serialize 0.3.16 (registry+https://github.com/rust-lang/crates.io-index)",
+]
 
--- a/libimagstore/Cargo.toml
+++ b/libimagstore/Cargo.toml
@@ -2,3 +2,8 @@
 name = "libimagstore"
 version = "0.1.0"
 authors = ["Matthias Beyer <mail@beyermatthias.de>"]
+
+[dependencies]
+
+toml = "0.1.25"
+
--- a/libimagstore/src/content.rs
+++ b/libimagstore/src/content.rs
@@ -1,0 +1,1 @@
+pub type EntryContent = String;
--- a/libimagstore/src/entry.rs
+++ b/libimagstore/src/entry.rs
@@ -1,0 +1,11 @@
+use std::path::PathBuf;
+
+use header::EntryHeader;
+use content::EntryContent;
+
+pub struct Entry {
+    location: PathBuf,
+    header: EntryHeader,
+    content: EntryContent,
+}
+
--- a/libimagstore/src/header.rs
+++ b/libimagstore/src/header.rs
@@ -1,0 +1,19 @@
+use toml::Table;
+
+pub struct EntryHeader {
+    toml: Table,
+}
+
+impl EntryHeader {
+
+    pub fn new(toml: Table) -> EntryHeader {
+        EntryHeader {
+            toml: toml,
+        }
+    }
+
+    pub fn toml(&self) -> &Table {
+        &self.toml
+    }
+
+}
--- a/libimagstore/src/lib.rs
+++ b/libimagstore/src/lib.rs
@@ -1,3 +1,9 @@
-#[test]
-fn it_works() {
-}
+extern crate toml;
+
+pub mod content;
+pub mod entry;
+pub mod error;
+pub mod header;
+pub mod single_use_lock;
+pub mod store;
+


neithernut:
Note that I don't have a cargo-installation and I'm still unfamiliar to `rustc`. However, I start to get the hang on the syntax of rust.

libimagstore/src/single_use_lock.rs
```diff
@@ -0,0 +1,5 @@
+pub trait SingleUseLock<T> {
+    fn access(&self) -> &T;
+    fn unlock(self) -> T;
+}
```
matthiasbeyer: What is the semantical difference between a `SingleUseLock` and a `std::sync::RwLock`?

libimagstore/src/store.rs
```diff
@@ -0,0 +1,17 @@
...
+
+pub type Result<T> = RResult<T, StoreError>;
+pub type LockedEntry = SingleUseLock<Entry>;
+
+pub trait Store {
+    fn create(&self, entry : Entry) -> Result<()>;
+    fn retrieve(&self, id : string) -> Result<LockedEntry>;
+    fn retrieve_copy(&self, id : string) -> Result<Entry>;
+    fn update(&self, LockedEntry) -> Result<()>;
+    fn delete(&self, id : string) -> Result<()>;
```
matthiasbeyer: First: It is `String` (uppercase). I'm not sure why travis doesn't moan on this...

Second: What is an ID here? Is it a Path? Is it a Hash? I assume it is a path, as we agreed on this scheme already afaik, so please use `PathBuf`.

TheNeikos: That's because none of the source files are being referenced from the lib right now, meaning that they do not get compiled.

Use `mod <filename>;` to add them

libimagstore/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+    name: Option<&'static str>,
+    expl: Option<&'static str>,
+    cause: Option<Box<Error>>,
+}
+
+impl StoreError {
+
+    pub fn new() -> StoreError {
+        StoreError {
+            name: None,
```
TheNeikos: `name` and `expl` should be put into enums, allowing one to handle different errors differently.

For example:

```rust
enum StoreErrorType {
    IdNotFound,
    OutOfSpace,
    <Other kinds of errors that one might encounter>
}
```

This is a very clean /and/ typesafe way of handling errors.

libimagstore/src/single_use_lock.rs
TheNeikos: The RwLock stays around, but the SingleUseLock consumes itself and thus you can use T, but then it disappears once you stop using it.

libimagstore/src/error.rs
matthiasbeyer: Sounds sane, yes. This is actually my code, so I will fix this in my PR. I guess I can remove the "name" part for this and replace it with some `Type -> String` helper function which translates `IdNotFound` (The Type) into `"ID not found"` (The String) for printing the error messages...

libimagstore/src/store.rs
TheNeikos: I also suggest to use a custom trait for the id, which could be called `Id` and works similarly to the `Pattern` here: https://doc.rust-lang.org/stable/std/primitive.str.html#method.rfind

The idea is that you have a trait that you can use to convert into whatever you need. Thus you can easily add more ways of searching for files. It would implement a single method called `matches`. It takes a Lazy File/Filename as it's arguments and would allow one to implement it for &'a str or &'a String to match against the filename, or a closure for more advanced matching. 

I would note that this is a more advanced way of writing Rust but it makes it seriously fun to use and also **easy**. I would be willing to implement it afterwards, if you use a String right now it will *not* change the API in a backwards incompatible way so we could do this whenever.

libimagstore/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Error as FmtError;
+use std::clone::Clone;
+
+use std::io::Error as IOError;
+
+pub struct StoreError {
+    name: Option<&'static str>,
```
TheNeikos: `name` and `expl` should be put into enums, allowing one to handle different errors differently.

For example:

```rust
enum StoreErrorType {
    IdNotFound,
    OutOfSpace,
    // <Other kinds of errors that one might encounter>
}
```

This is a very clean /and/ typesafe way of handling errors.

libimagstore/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+    expl: Option<&'static str>,
+    cause: Option<Box<Error>>,
+}
+
+impl StoreError {
+
+    pub fn new() -> StoreError {
```
TheNeikos: I do not think it is a good idea to create an empty struct this way, rust doesn't allow incomplete structs, putting everything behind Options is very un-rust like.

If you wanted to change the fields afterwards that will not work as the fields themselves are not pub. And you wouldn't want them to be pub either. If you implement my suggestion of an enum you will have a very clean interface and just 1 constructor:

```rust
pub fn new(type: StoreErrorType, cause: Option<Box<Error>>) -> StoreError {
    StoreError {
        kind: type,
        cause: cause,
    }
}

**Comment by TheNeikos** [comment]
As general advice I will give you the same I gave @matthiasbeyer, try to encode *everything* with types. Even empty ones! Rust's type system is very powerful, and using it allows you to create expressive and useful API's that are forwards compatible.

File: libimagstore/src/single_use_lock.rs (hunk 1 above)
Comment (TheNeikos) [inline libimagstore/src/single_use_lock.rs]: I would like to note that it does not allow you to tell when T is not used anymore. If you wish to do that I suggest to make a wrapper type that dereferences to T.

File: libimagstore/src/error.rs (hunk 4 above)
Comment (matthiasbeyer) [inline libimagstore/src/error.rs]: Sounds sane, yes. This is actually my code, so I will fix this in my PR. I guess I can remove the "name" part for this and replace it with some Type -> String helper function which translates IdNotFound (The Type) into "ID not found" (The String) for printing the error messages...

File: libimagstore/src/error.rs (hunk 5 above)
Comment (matthiasbeyer) [inline libimagstore/src/error.rs]: Will add that as well.

File: libimagstore/src/store.rs (hunk 2 above)
Comment (matthiasbeyer) [inline libimagstore/src/store.rs]: I'd like to see code on this. Feel free to submit a PR to #95 for implementing this.

File: libimagstore/src/single_use_lock.rs (hunk 1 above)
Comment (neithernut) [inline libimagstore/src/single_use_lock.rs]: In the overall picture, the idea was to have distinct instances of `Enty`, which do not share state. This way, one can pull both read-only and write-back-enabled items from the store without worrying about conflicts.

File: libimagstore/src/store.rs (hunk 2 above)
Comment (neithernut) [inline libimagstore/src/store.rs]: Mh... pushed a new version of `lib.rs`. Since I'm still missing cargo I have to (ab)use the CI for now...
I also agree with @TheNeikos about the idea of an id type.

**Comment by neithernut** [comment]
That was my intention, but I try to apply it to more aspects of the code.

File: libimagstore/src/single_use_lock.rs (hunk 1 above)
Comment (matthiasbeyer) [inline libimagstore/src/single_use_lock.rs]: AFAICT this is exactly what an `RwLock` would do.

File: libimagstore/src/store.rs (hunk 2 above)
Comment (neithernut) [inline libimagstore/src/store.rs]: @matthiasbeyer about your third point: there are, essentially, two use-cases which I considered:
 * An item is fetched from the store for modification: in this case, the programmer will, in most cases, modify in one place, or in one thread. If she does not, she is free to use any storage solution for items _appropriate for her use_case_.
 * An item is fetched for reading-only. In this case, you may want to modify the item in another context, a already explained in the associated issue. Having shared state obstructs this use-case. Is one _expects_ to pull and temporarily multiple items, one can, again, use an appropriate storage solution.

File: libimagstore/src/single_use_lock.rs (hunk 1 above)
Comment (neithernut) [inline libimagstore/src/single_use_lock.rs]: An RWLock can be locked after being, unlocked, no?

Comment (matthiasbeyer) [inline libimagstore/src/single_use_lock.rs]: Yes, of course!

Comment (neithernut) [inline libimagstore/src/single_use_lock.rs]: The whole purpose of this type is to avoid the loop-holes opening up when using `RWLock`s. If you'd like to have an explanation on these loop-holes _again_, I'd propose to meet in a room with a black-board.

Comment (TheNeikos) [inline libimagstore/src/single_use_lock.rs]: Loopholes? I would be very interested which ones you talk about, do you mean some sort of race condition?

Comment (neithernut) [inline libimagstore/src/single_use_lock.rs]: `RWLock`s can be logged by the user. Hence, if receiving an entry in an `update()` call, we have no way of knowing whether the item was, in fact, unlocked and locked again later. If an item can be temporarily unlocked, the item can be altered by another thread/process in the meantime, thus leading to data loss or inconsistencies.
The type I propose can only be `unlocked()` once. Theoretically, it _Can_ be unwrapped by the user and put into a new `SingleUseLock`, but if we make the implementation private, for use in the store only, the user has to implement the trait herself, which takes far more effort and _probably_ will not be done by accident.
Btw: is there a way to disallow the implementation of a trait outside the crate is was declared?

Another problem, which the interface of `RWLock` does not cover is that elevating a read-lock to a write-lock potentially leads to dead-locks. Locks may only be _relaxed_ after acquiring them.

Comment (TheNeikos) [inline libimagstore/src/single_use_lock.rs]: Regarding traits, yes, you just do not export it.

I see what you mean. What about update taking a closure, taking the file as an input. Then inside the closure you can mutate the file as much as you want, and once the closure exits it gets written.

This means that you do never get access to the file directly, you can still write to it with your own data, and others can still read while you do not update.

Comment (matthiasbeyer) [inline libimagstore/src/single_use_lock.rs]: @TheNeikos Sounds sane.

I'd propose to change the interface to take the lock results:

```rs
fn locked_update(&self, LockResult<RwLockWriteGuard<Entry>);
fn try_locked_update(&self, TryLockResult<RwLockWriteGuard<Entry>);
fn try_update(&self, Arc<RwLock<Entry>>) -> Result<(), PoisonError<Guard>>;
```

Would result in the same behaviour as the proposal of @TheNeikos , though with more interface-noise. I'm not sure which proposal is better, though I like what @TheNeikos proposed.

neithernut: Mh... this option may be viable, but one has to ensure that:
 * the closure is bundled with the `Entry` in a way does not allow retrieval and re-bundling of the closure with another (or the same) `Entry`.
 * a lock can be transported together with that bundle.

Not quite sure whether this makes things any better, but if you have any ideas, I'd be interested in them.

neithernut: Ah, there is a `RwLockWriteGuard`? Perfect. Looks like what I wanted.
@matthiasbeyer: can you explain why you want an `try_update()` call? Optimally, the user should never see a `RwLock<Entry>`, unless she creates one for context-specific storage imho. It just reintroduces the caveats I tried to avoid.

neithernut: Argh, still not usable as-is, since it does support allow file-locking.

matthiasbeyer: From the stdlib docs: `RwLock::write() -> LockResult<RwLockWriteGuard<T>>`.

@neithernut I'm not completely sure on `try_update()`. Its semantics would be that it would aquire a try_lock on the passed entry internally, if that fails returns and if not it calls `update()` with the locked Entry (so it is a wrapper around `update()`). This way one could go on with processing things if locking fails. I know this can be done with `update()` as well, but I'm hiding a bit of code behind the `try_update()` wrapper here, so it is just for convenience.

TheNeikos: Closures in rust are really just anonymous functions that are type checked at compile time and may use their parent stack. So it's not something that usually gets 'moved' around. Which makes me a bit confused by what you mean with 'bundled with an entry'.

Regarding file locking, you could do that before calling the closure for example.

matthiasbeyer: Let me add: The user of the library needs to see the `RwLock<Entry>` part. Actually, she sees `Arc<RwLock<Entry>>`, there's no way around that IMO.

If we apply my proposal with my `playground.rs/flocking`, we'd replace the `RwLock` part with a `FSLock` maybe, but that's all to it.

matthiasbeyer: > Regarding file locking, you could do that before calling the closure for example.

So the closure would get an `RwLockWriteGuard`! :smile_cat:

TheNeikos: No, it would get an entry or simply not be called. Why would you make it more complicated?

matthiasbeyer: Legit.

TheNeikos: If you do not want it to block, you can make it return an UpdateResult which will tell you if it succeeded or where it failed.

neithernut: So here is the thing: the file-lock _must_ live/be held from the point the item is pulled from the store until the user is done with modifying it. The `update()` call _must not_ try to lock the file, but rely on the file being locked. It shouldn't touch the lock itself _at all_.
@matthiasbeyer so that speaks against `try_update()`, at least with the current semantics.
@TheNeikos that's why locking the file before executing the closure doesn't work.

About the `update()` call getting a closure: what sense does it make in this context to have a closure which can, essentially, be applied on any item? How does this help enforce the file being locked until the user is done with editing/updating the item?

neithernut: Giving the user a lock which she can unlock and lock again at any time does break data consistency.

TheNeikos: In that case, no that won't work. 
But what is possible is to have the store, (which is a trait? Do we plan on having multiple store implementations?) is to return a FileLockEntry, that can only be created by locking the file and getting the corresponding RwLock. On drop it unlocks. It would be bound to the store.

1. No way for the user to unlock it except by dropping it.
2. It is thread safe.
3. It is process safe (as so far as flock is process safe)
4. Lifetimes are also guaranteed (since Rust)

neithernut: That was, in fact, my original proposal made in #95.

matthiasbeyer: > Do we plan on having multiple store implementations

I did this mainly to be able to implement a in-memory store for testing purposes, though we might want to have different caching technologies in the background, which would result in different store implementations, yes. 

Overall: This is for future flexibility by now.

neithernut: My original proposal contained two traits, one which behaves similar to what you described and one for mere data representation (used when you don't intend to write an item back to disk)

matthiasbeyer: > It is process safe (as so far as flock is process safe)

FS-locks are advisory, and that's what the crate `fs2` tells you as well, so "kindof" is the right term here.

TheNeikos: @neithernut I don't see anything in #95 pertaining to that, can you link to it directly?

@neithernut I do think that you should still lock the file even if only reading, what if you read between two write calls of another process and become a malformed file? That throw an error for no reason.

@matthiasbeyer Alright

neithernut: Bah. No wonder you couldn'T find anything. I keep linking to @matthiasbeyer's PR instead of the issue https://github.com/matthiasbeyer/imag/issues/76#issuecomment-171078539.

Yes, you should lock files for reading. But there is no reason at all to keep a file locked if you already loaded everything into RAM. Hence, the file needs only to be read-locked within `Store::retrieve()` in this proposal.

TheNeikos: In that case it should be no problem to just `clone` the entry after locking it and returning that. Since it is not bound to the store there is no way to write it back as it is (without acquiring a lock again and putting the struct back in, which is pretty much legit).

Unless you disagree, I think we should proceed then with this:

- A new type is to be made, a FileLock. It dereferences to an Entry (Check out deref and derefmut), thus allowing one to freely modify it. Once the file is dropped it is written and returned to the store and unlocked. 
- This type is returned by 'borrowing' from the store. It has similar semantics to the rust  interpretation, and additionally has safeguards for process interactions.
- The store itself will take care of locking and creating the FileLock, making sure that it's invariants are held up.

This is very possible to write in Rust and is also clean in my opinion.

matthiasbeyer: @TheNeikos go ahead. Finally we get some progress here.

Try to reuse as many commits from both our PRs (or even base on one of them) and ensure the documentation is updated for this as well (my PR contains documentation foo, actually)!

neithernut: @TheNeikos yep, that would be fine.

neithernut: Maybe you should base your branch on the same commit I have, and cherry-pick any commits you need.

TheNeikos: Will take yours and rebase it on the updated branch of @matthiasbeyer and work off that.

matthiasbeyer: @TheNeikos sounds awesome to me!

neithernut:
Closed in favour of #99.

PR #99, opened 2016-01-16
Add libimagstore

TheNeikos:
Alright, do check the file and feel free to post comments/ask questions


TheNeikos:
I just noticed that your travis file does not run the tests of each sub-crate, I think we want that to be added.

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+     */
+    entries: Arc<RwLock<HashMap<StoreId, (File, Option<Entry>)>>>,
+}
+
+impl Store {
+
+    fn create(&self, entry: Entry) -> Result<()> {
+        unimplemented!();
+    }
+    fn retrieve<'a>(&'a self, path: StoreId) -> Result<FileLockEntry<'a>> {
```
matthiasbeyer: Should call the argument `id` instead of `path` here.

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+        unimplemented!();
+    }
+    fn update<'a>(&'a self, entry: FileLockEntry<'a>) -> Result<()> {
+        unimplemented!();
+    }
+    fn retrieve_copy(&self, path: StoreId) -> Result<Entry> {
```
matthiasbeyer: ... and here.

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+        unimplemented!();
+    }
+    fn delete(&self, path: StoreId) -> Result<()> {
```
matthiasbeyer: ... and here.

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.entry
+    }
+}
+
+impl<'a> Drop for FileLockEntry<'a> {
+    fn drop(&mut self) {
+        let mut map = self.store.entries.write().unwrap();
+        let (_, ref mut en) = *map.get_mut(&self.key).unwrap();
+        *en = Some(self.entry.clone());
```
matthiasbeyer: Can't this be written as

```rs
*(map.get_mut(&self.key).unwrap()) = Some(self.entry.clone());
```

:smile: ?

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+
+impl<'a> ::std::ops::Deref for FileLockEntry<'a> {
+    type Target = Entry;
+
+    fn deref(&self) -> &Self::Target {
+        &self.entry
+    }
+}
+
+impl<'a> ::std::ops::DerefMut for FileLockEntry<'a> {
```
matthiasbeyer: Care to import `DerefMut` to have less noise here?

libimagstore/src/store.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn new(store: &'a Store, entry: Entry, key: StoreId) -> FileLockEntry<'a> {
+        FileLockEntry {
+            store: store,
+            entry: entry,
+            key: key,
+        }
+    }
+}
+
+impl<'a> ::std::ops::Deref for FileLockEntry<'a> {
```
matthiasbeyer: Care to import `Deref` to have less noise here?

matthiasbeyer:
Looks good to me, besides:

You create a cyclic reference between the `Store` object and the `FileLockEntry`. I know this is something we want, as such an lock must be bound to the store it comes from... but this might blow up in complexity at some point, so make sure you add some documentation here.

---

Can you please fix these minor points:

* Add blank lines between the functions in `impl Store`
* Export the `FileLockEntry` type to a new file
* Export the `StoreId` type to a new file

matthiasbeyer:
Also, as far as I can see: There is no way that two threads can read one entry at the same time without actually duplicating an `Entry` instance, am I right with this?

libimagstore/src/store.rs
TheNeikos: No, as it's a tuple.

libimagstore/src/store.rs
TheNeikos: To have more noise at the top?

TheNeikos:
No, two threads cannot `retrieve` a file at the same time, since that would be inconsistent. However, one could `receive_copy` it which is why it's there.

TheNeikos:
Regarding putting FileLockEntry into into another module. No, you cannot. Store's fields are private, you want them to stay private.

libimagstore/src/store.rs
matthiasbeyer: Nah, I meant:

```rs
*((map.get_mut(&self.key).unwrap()).1) = Some(self.entry.clone());
```

libimagstore/src/store.rs
matthiasbeyer: Two times noise vs. one time noise.

matthiasbeyer:
> Regarding putting FileLockEntry into into another module. No, you cannot. Store's fields are private, you want them to stay private.

Ah, I didn't notice that you access private content there. Nothing said.

neithernut:
Mh... kind of expected that you'd let the user _borrow_ a `FileLockEntry` from your previous comments. IMO this would solve a few problems which you now try to avoid by the mechanism @matthiasbeyer pointed out as circular referencing. Did I misunderstand you or did you have a specific reason not to do it like this?

neithernut:
Mh... kind of expected that you'd let the user _borrow_ a `FileLockEntry` from your previous comments. IMO this would solve a few problems which you now try to avoid by the mechanism @matthiasbeyer pointed out as circular referencing. Did I misunderstand you or did you have a specific reason not to do it like this?

TheNeikos:
Yes, as you would have to keep the borrow on the Arc and the RwLock open as well to even get to the HashMap. So the only way is to move it out of the HashMap, which is also safe to do.

neithernut:
Good point.

TheNeikos:
Currently writing an alternative impl, that might ease the pain points a bit, tell me once I've pushed.

neithernut:
As far as I can tell the interface looks good.

matthiasbeyer:
I like the interface as well.

Maybe rename `retreive` to `fetch` or even `get`,... the shorter the better IMHO. But that's the only point, really.

TheNeikos:
@neithernut @matthiasbeyer Because I do not like Tuples in structs when you could use structs instead.

It also allows you make helper methods.

neithernut:
Do you want to rewrite the history to something sane?

matthiasbeyer:
I like it.

I will re-read the doc part now, annotate what has to be changed. After that, if everyone agrees, I will merge this.

doc/src/04000-lib-store.md
```diff
@@ -4,3 +4,67 @@
...
+## Types {#sec:libstore:types}
+
+The types in @tbl:libstore:types are exported by the library.
+
+| Type         | Meaning                                          |
+| :----------- | :----------------------------------------------- |
+| Entry        | Entity on the Filesystem, File                   |
+| EntryContent | User-Content of the Entry                        |
+| EntryHeader  | Header of the Entry                              |
+| Store        | Store interface                                  |
```
matthiasbeyer: @TheNeikos can you add the extra types we agreed on?

matthiasbeyer: Also in the sub-chapters, of course.

neithernut:
I'd also like to see documentation on the interface itself, apart from external documentation.

TheNeikos:
@neithernut I will add it once we've decided one thing.

doc/src/04000-lib-store.md
```diff
@@ -4,3 +4,67 @@
...
+- An instance of `EntryContent` as interface to the content of the file
+  (@sec:libstore:types:entrycontent).
+- An instance of `EntryHeader` as interface to the header of the file
+  (@sec:libstore:types:entryheader).
+
+The entry type MUST export functions to get
+
+- The content object
+- The header object
+- The path of the actual file
```
matthiasbeyer: @TheNeikos please add that the entry type also must export functions to set header and content, but not path, of course.

Or maybe it would even make sense to _set_ the path, so we can move files with great convenience. Would make the implementation of the store a bit harder, though. @neithernut what do you think about this?

neithernut:
Mh... ok. I have to go now. Would you mind waiting just until tomorrow morning, @matthiasbeyer --given that @TheNeikos considers the work done? I'd like to get another look at it tonight.

neithernut:
I'd also vote in favour of cleaning up the history btw.

matthiasbeyer:
@neithernut Mah... okay, but I guess @TheNeikos and I will do some more work on imag,... maybe at other places if we can.

Cleaning up history... yeah, partly. I don't care about the history beeing long... this way one can see that this was a long progress... I'll look at it a bit closer once this is completely ready.

libimagstore/src/store.rs
```diff
@@ -0,0 +1,210 @@
...
+
+    /// Set the entry to the given one. This will also overwrite the entry
+    /// and flush it to disk
+    fn set_entry(&mut self, entry: Entry) -> Result<()> {
+        self.entry = Some(entry);
+        unimplemented!()
+    }
+
+    /// We borrow the entry
+    fn get_entry(&mut self) -> Result<Entry> {
```
matthiasbeyer: Is this type signature what we want? Why do we need a `&mut self` to get the `Entry`? Also: Don't we want a `Result<&Entry>` back?

TheNeikos: Yes, you basically take out the entry out of... the entry. This is a private method and to be used in the constructor of the `FileLockEntry`

libimagstore/src/store.rs
```diff
@@ -0,0 +1,210 @@
...
+impl<'a> ::std::ops::DerefMut for FileLockEntry<'a> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.entry
+    }
+}
+
+impl<'a> Drop for FileLockEntry<'a> {
+    fn drop(&mut self) {
+        let mut map = self.store.entries.write().unwrap();
+        map.get_mut(&self.key).unwrap().set_entry(self.entry.clone());
```
neithernut: The entry should be written back in this case, the entry should be rather be removed from the map, freeing the lock on the file.
 * If I didn't miss anything, the lock isn't released anywhere else and it's most natural to assume a file would be free for further editing after the only reference visible to the user vanished. Keeping the file locked is counter-intuitive.
 * We requiring the user to call `Store::update()` to write stuff to disk. It is easy to understand to a user that changes which were committed previously end up in new versions pulled from the store. If we, however, keep touched, unwritten versions around in memory, those changes also end up in entries pulled from the store afterwards, which, again, is counter-intuitive.
 * When we do _not_ store intermediate changes in the store, one can easily implement a "discard" feature, where changes are rolled back to the last saved version or never written to disk in the first place.

neithernut: inb4: I am aware that this makes the store effectively not cache any items. But I do believe that the benefits from not caching everything automatically in the store, without the user's consent, is what should be done.
In the beginning, we'll have one-shot commands anyway, for which the caching is not important. Creators of longer-lived applications, which do may want some sort of caching will, on the other hand, most likely require more control over the caching.

doc/src/04000-lib-store.md
```diff
@@ -4,3 +4,67 @@
...
+filesystem which is represented by an object of this type.
+It also MUST contain a getter for this variable.
+It MUST NOT contain a setter for this variable, as changing the store while the
+programm is running is not allowed.
+
+The `Store` object MAY contain caching functionality.
+If the `Store` type contains caching functionality for filesystem operations,
+the interface MUST NOT differ from the non-caching interface.
+If the `Store` type contains caching functionality, these functionality MUST NOT
+BE visible to the user of the library.
```
neithernut: As already argued in another place, this entire section is bogus. We should not cache on this level.

neithernut:
> Cleaning up history... yeah, partly. I don't care about the history beeing long... this way one can see that this was a long progress...

Nobody has to justify the amount of time it took for this to get finished. On the other hand, if someone looks for information on why the interface or implementation looks the way it does, this person will be better of reading the design document.
Also, the commit texts will not reflect the current discussion in a satisfying manner, anyway. Fewer commits which introduce the interface on a direct route give a far better overview over design decisions.
It _is_ a bit of work though, considering how the history looks now...

libimagstore/src/store.rs
```diff
@@ -0,0 +1,210 @@
...
+        res.push(second.into_storeid());
+        res
+    }
+}
+
+/// A store entry, depending on the option type it is either borrowed currently
+/// or not.
+struct StoreEntry {
+    file: File,
+    entry: Option<Entry>,
```
neithernut: If the change requests I made regarding caching of entries are accepted, the `StoreEntry` doesn't need this member any more.

neithernut:
I've gone over the sources a few times and found an issue regarding the caching. Other than that, I didn't find anything odd.
Since my objection of caching in the store will probably lead to some discussion, I'll spare us one round-trip by posting some arguments off the top of my head:
 * Caching is not required for (most) single-shot commands.
 * Long-lived applications can easily create their own cache, over which they have control, if needed.
 * A cache requires some management. If we implement caching in the store, we also have to implement this management (maximum cache size, etc.), introducing complexity, which, at least for now, isn't required by _anything_.
 * A general caching policy without any possibility to influence it by the user of the interface will most likely lead to interference at some point. Especially when file-locks are involved. Especially for long-lived applications, for which the cache may have been intended originally.
 * Unix-philosophy: small, combinable tools instead of a tool which tries to do everything.

matthiasbeyer:
I agree with you on the caching part.

On the rebase-cleanup: @TheNeikos should cleanup a bit, yes, but as the PR here contains a merge of our two PRs, this gets rather hard at some point (at the merge commit, actually). So "light" cleanup would be okay, but I don't mind if @TheNeikos does not want to cleanup any further back than to the merge commit!

libimagstore/src/store.rs
TheNeikos: I don't think I understand what you mean, the current dataflow is as follows:

1. There are entries in the HashMap
2. A user borrows one and thus gets indirect access to it through the `FileLockEntry`, this will also lock the file.
3. After the `FileLockEntry` is dropped or 'updated' the whole thing is written back to the harddisk.
4. Then the entry is put back into the store, available to be taken again, and the file is unlocked.

If you mean that there could be another change made before the entry is borrowed again, I agree and I will fix that now.

neithernut: That way one of my points, yes.

TheNeikos:
I've changed the interface, is this somewhat what you though?

neithernut:
Yes, now it's fine.

TheNeikos:
Alright, gonna try to clean up the history a bit

matthiasbeyer:
:+1:

TheNeikos:
I'm not sure I can do much better due to the fact that we've all three changed things.

matthiasbeyer:
@TheNeikos Looks really good to me. Can you update the documentations (the things me and @neithernut annotated) and then I guess... we can merge!

TheNeikos:
@matthiasbeyer Something like this?

neithernut:
Yep.

matthiasbeyer:
Looks good.

Everyone ACK on merge?

neithernut:
ACK.

TheNeikos:
I suppose if you want mine as well: AC

matthiasbeyer:
@homu r+

PR #100, opened 2016-01-16
travis: Run cargo chain for all cargo files

matthiasbeyer:
Test sub-crates if there are changes in them, as pointed out by @TheNeikos in #99, this wasn't happening before.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,8 +47,12 @@ script:
 
         for d in $(find -name "Cargo.toml" | grep -vE "^.$"); do
             dir=$(dirname $d)
-            changes_in $dir && travis_cargo_run_in $dir
+            changes_in $dir &&                \
+              echo -e "\nRunning in $d\n" &&  \
+              travis_cargo_run_in $dir
         done
+
+        echo -e "\nRunning on top-level crate...\n"
         travis_cargo_run_in "."
 
 addons:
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Hm. travis seems to run cargo on all tests with these changes... :astonished:

.travis.yml
```diff
@@ -44,6 +44,11 @@ script:
         }
 
         [[ $(changes_in "doc") ]] && echo "Changes in ./doc are not build by CI"
+
+        for d in $(find -name "Cargo.toml" | grep -vE "^.$"); do
+            dir=$(dirname $d)
+            changes_in $dir && travis_cargo_run_in $dir
+        done
         travis_cargo_run_in "."
```
neithernut: Wouldn't it be cleaner if you omit the `grep` and remove this line?

matthiasbeyer: No, as we want to build in `"."` always, not with the preconditions which are tested in `changes_in`.

matthiasbeyer:
Ah, it actually works and compiles everything because everything is a dep from the main crate. I wondered why the libs got build, but yeah... it's clearer with the output now...

matthiasbeyer:
@TheNeikos can we merge here? What do you think?

TheNeikos:
Ye, sounds good

matthiasbeyer:
@homu r+

PR #101, opened 2016-01-17
Libimagstore/impl store create

matthiasbeyer:
Implementation of `Store::create`.


TheNeikos:
Can you rebase so we get to see only your changes?

matthiasbeyer:
Already done.

matthiasbeyer:
I have to iterate over this, because of the caching part.

libimagstore/src/store.rs
```diff
@@ -116,7 +116,58 @@ pub struct Store {
...
     pub fn create(&self, entry: Entry) -> Result<()> {
-        unimplemented!();
+        use std::fs::create_dir_all;
+        use std::io::Write;
+        use std::ops::Deref;
+
+        let dirpath = {
+            let path = entry.location();
+            if path.file_name().is_some() {
+                path.parent().unwrap().to_path_buf() // TODO: Safe here?
```
TheNeikos: What happens if I put in 'general.conf' ?

libimagstore/src/store.rs
```diff
@@ -116,7 +116,58 @@ pub struct Store {
...
+                file.write_all(entry.as_string().as_bytes())
+                    .map_err(|e| {
+                        StoreError::new(StoreErrorKind::IOWrite, Some(Box::new(e)))
+                    })
+                    .and_then(|_| self.cache(entry, file))
+
+            })
+    }
+
+    fn cache(&self, e: Entry, f: File) -> Result<()> {
```
TheNeikos: Cache?

neithernut:
For the record: targets #102.

libimagstore/src/store.rs
matthiasbeyer: You mean a non-store path?

neithernut: I guess you don't properly deal with paths targeting files outside the store.

TheNeikos: @neithernut That shouldn't happen

@matthiasbeyer The store itself has a path, you have to them together before doing this.

libimagstore/src/store.rs
```diff
@@ -116,7 +116,64 @@ pub struct Store {
...
+        let id = e.location().clone();
+        let se = StoreEntry {
+            file: f,
+            entry: StoreEntryPresence::Present,
+        };
+
+        self.entries
+            .write()
+            .map_err(|e| {
+                // StoreError::new(StoreErrorKind::CacheLocking, Some(Box::new(e)))
```
TheNeikos: Nah, you can't do that because the `WriteLockGuard` from the RwLock is not Send :)

neithernut: Won't this entire function be removed?

TheNeikos: I don't know, perhaps the naming is just misleading.

matthiasbeyer:
I did some fixup and reverted the caching-functionality commits. Review pls!

neithernut:
Ah, you introduced an IO error type. Good, because I'll need one.

libimagstore/src/error.rs
```diff
@@ -12,13 +12,23 @@ use std::io::Error as IOError;
 pub enum StoreErrorKind {
     IdNotFound,
     OutOfMemory,
+    IO,
+    IOMkdirP,
+    IOWrite,
```
neithernut: Wouldn't it make more sense to use the error `cause` instead of a bunch of new enum values?

matthiasbeyer:
@neithernut feel free to `cherry-pick` the commits you need. They vanish on rebase anyways.

libimagstore/src/error.rs
matthiasbeyer: No, that's not the intension of the `cause`. With `cause` you can encapsulate errors in errors and create a "chain of errors" - which is semantically different from the underlying `cause`!

matthiasbeyer: Its like:

* enum = What failed
* cause = Why did it fail

neithernut:
I'll do so once my version compiles.

matthiasbeyer:
@TheNeikos , @neithernut Review appreciated

libimagstore/src/store.rs
```diff
@@ -125,7 +125,38 @@ impl Store {
...
+            return Err(StoreError::new(StoreErrorKind::InvalidStorePath, None));
+        }
+
+        let dirpath = {
+            if path.file_name().is_some() {
+                let mut newpath = self.location.clone();
+                newpath.push(path.parent().unwrap().to_path_buf()); // TODO: unwrap() safe here?
+                newpath
+            } else {
+                unimplemented!()
```
TheNeikos: Do you plan on implementing something? Why not return an error?

libimagstore/src/entry.rs
```diff
@@ -9,3 +9,15 @@ pub struct Entry {
...
 
+impl Entry {
+
+    pub fn location(&self) -> &StoreId {
+        &self.location
+    }
+
+    pub fn as_string(&self) -> String {
+        unimplemented!();
+    }
```
neithernut: Could you elaborate on the semantics of this method?

matthiasbeyer: It simply returns the content which is written to file later. So the header is parsed to text, the seperators are inserted... etc.

neithernut: The complete file? Header + Content? In this case, I'd prefer a different name, e.g. `dump()`.

matthiasbeyer: Sounds reasonable. Will change after LUG meetup.

matthiasbeyer:
@TheNeikos ping.

libimagstore/src/store.rs
```diff
@@ -125,7 +125,38 @@ impl Store {
...
+
+        let path = entry.get_location().clone();
+        if !path.is_absolute() {
+            return Err(StoreError::new(StoreErrorKind::InvalidStorePath, None));
+        }
+
+        let dirpath = {
+            if path.file_name().is_some() {
+                let mut newpath = self.location.clone();
+                newpath.push(path.parent().unwrap().to_path_buf()); // TODO: unwrap() safe here?
```
neithernut: Better be paranoid.

libimagstore/src/entry.rs
```diff
@@ -36,5 +36,9 @@ impl Entry {
         &mut self.content
     }
 
+    pub fn dump(&self) -> String {
+        unimplemented!();
+    }
+
```
neithernut: This might serve as a base for a point of #134.

libimagstore/src/store.rs
matthiasbeyer: The point is, if we can get a `file_name().is_some()`, we should have a `parent()` here, right? So the `unwrap()` won't fail and we wouldn't need to do complicated error handling here...

neithernut: That's why I wrote about being paranoid. You could, at least, change it from a `TODO` to a comment explaining why you assume `unwrap()` to be save there, or omit it altogether.
Also, what if path happens to be `D:\` on a windows platform? (Not that I actually care about that platform)

matthiasbeyer:
Done by #142 ...

PR #110, opened 2016-01-17
Remove methods which are not needed anymore

neithernut:
These methods were intended for managing cached entries. Sice it was
decided not to serve caching in the store, these methods are no longer
needed.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -82,21 +82,6 @@ impl StoreEntry {
     fn is_borrowed(&self) -> bool {
         self.entry == StoreEntryPresence::Borrowed
     }
-
-    /// Flush the entry to disk
-    fn set_entry(&mut self, entry: Entry) -> Result<()> {
-        unimplemented!()
-    }
-
-    /// We borrow the entry
-    fn get_entry(&mut self) -> Result<Entry> {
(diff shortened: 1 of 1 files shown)


TheNeikos:
The idea was that you would simply create a `StoreEntry` and let that entry take care of everything.

neithernut:
Is this still the case? We're currently implementing everything in the store directly.

TheNeikos:
I don't think that's a good idea, you will basically duplicate functionality across all the functions.
All should you need to do is:

1. Check if Item is in store (Or if you create it, check that it isn't in the store)
2. Get the `StoreEntry`, tell the store entry to give you an `Entry`

One oversight we had was that the File itself is a member, although it should be a lazy file.

TheNeikos:
@matthiasbeyer ^

matthiasbeyer:
I'm not sure. :unamused:

neithernut:
Concentrating _some_ functionality in `StoreEntry` makes sense. It also simplifies things greatly, as realized while implementing `Store::delete()`.

TheNeikos:
Okay, we will merge this for now, removing the methods. Should we see the need for them during development adding them back in should not be a problem.

matthiasbeyer:
@homu r+

Issue #106, opened 2016-01-17
Implement: Store::delete()


matthiasbeyer:

matthiasbeyer:
@neithernut Wanna do this?

neithernut:
Sure.

neithernut:
This issue can now be closed (I'll use `resolves #...` next time)

matthiasbeyer:
@neithernut Just use "Closes #...", but in the _pull request description_. Not in the commits please!

neithernut:
Of course not.

PR #111, opened 2016-01-17
Implement `Store::delete()`

neithernut:
Targets #106.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::fs::File;
+use std::fs::{File, remove_file};
 use std::ops::Drop;
 use std::path::PathBuf;
 use std::result::Result as RResult;
@@ -9,7 +9,7 @@ use std::sync::{RwLock, Mutex};
 use fs2::FileExt;
 
 use entry::Entry;
-use error::StoreError;
+use error::{StoreError, StoreErrorKind};
 
 /// The Result Type returned by any interaction with the store that could fail
 pub type Result<T> = RResult<T, StoreError>;
@@ -147,7 +147,17 @@ impl Store {
 
     /// Delete an entry
     pub fn delete(&self, id: StoreId) -> Result<()> {
-        unimplemented!();
+        let mut entries_lock = self.entries.write();
+        let mut entries = entries_lock.unwrap();
+
+        // if the entry is currently modified by the user, we cannot drop it
+        if entries.get(&id).map( |e| e.is_borrowed() ).unwrap_or(false) {
+            return Err(StoreError::new(StoreErrorKind::IdLocked, None));
+        }
+
+        // remove the entry first, then the file
+        entries.remove(&id);
+        remove_file(&id).map_err( |e| { StoreError::new(StoreErrorKind::FileError, Some(Box::new(e))) } )
     }
 }
 
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -151,13 +151,13 @@ impl Store {
         let mut entries = entries_lock.unwrap();
 
         // if the entry is currently modified by the user, we cannot drop it
-        if entries.get(&id).map( |e| e.is_borrowed() ).unwrap_or(false) {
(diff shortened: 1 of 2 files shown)


libimagstore/src/store.rs
```diff
@@ -147,7 +147,14 @@ impl Store {
 
     /// Delete an entry
     pub fn delete(&self, id: StoreId) -> Result<()> {
-        unimplemented!();
+        let mut entries = self.entries.write();
+
+        // if the entry is currently modified by the user, we cannot drop it
+        if entries.unwrap().get().and_then( |entry| { entry.isBorrowed() } ) {
+            return Result<StoreError::new(IdLocked)>;
```
TheNeikos: [Result](https://doc.rust-lang.org/stable/std/result/enum.Result.html) is an enum that has two variants, `Ok` or `Err`. The constructor of each variant is the same as their name. In this case `Result::Err`, however for ease of use Rust already `use`s the `Err` and `Ok` variant for you, thus you just have to write `Err(StoreError::new(..))`.

neithernut: Yep, I already fixed the error locally. Thanks, anyway.

neithernut:
Mh... I should definitely take some time and get me all those crates so I can build locally.

neithernut:
I finally got something building here. I'll clean up this mess now.

neithernut:
I think this is now ready for the first round of inspection.

libimagstore/src/store.rs
```diff
@@ -147,7 +147,17 @@ impl Store {
 
     /// Delete an entry
     pub fn delete(&self, id: StoreId) -> Result<()> {
-        unimplemented!();
+        let mut entries_lock = self.entries.write();
+        let mut entries = entries_lock.unwrap();
+
+        // if the entry is currently modified by the user, we cannot drop it
+        if entries.get(&id).map( |e| e.is_borrowed() ).unwrap_or(false) {
```
matthiasbeyer: You're hiding an error here. If the `get()` call fails due to the idea not beeing in the store, the error is still `IdLocked`.

Also: Closures are without trailing whitespace: `map(|e| e.is_borrowed())`.

libimagstore/src/store.rs
```diff
@@ -147,7 +147,17 @@ impl Store {
...
+            return Err(StoreError::new(StoreErrorKind::IdLocked, None));
```
matthiasbeyer: You don't need the `return` if you put the things below in the `else{}` branch, which is IMHO a good idea for readability in this case.

libimagstore/src/store.rs
```diff
@@ -147,7 +147,17 @@ impl Store {
...
+        }
+
+        // remove the entry first, then the file
+        entries.remove(&id);
+        remove_file(&id).map_err( |e| { StoreError::new(StoreErrorKind::FileError, Some(Box::new(e))) } )
```
matthiasbeyer: Also here, spaces around the closure should be removed.

matthiasbeyer:
Mainly style issues.

AFAIK, the store is `write()`-locked, so if the entry is not borrowed, there is no possibility that it gets borrorwed before we `::remove()` it, right?

libimagstore/src/store.rs
neithernut: If the item is not in the store and get returns `None`, we jump over the block returning the error.

neithernut:
That's the idea of keeping around `entries_lock`, yes.

libimagstore/src/store.rs
neithernut: Well, conceptually, it's an early return.

libimagstore/src/store.rs
neithernut: I'm not sure if I get what you mean.

matthiasbeyer: Meh... sorry, you're right.

matthiasbeyer:
Looks good for me now.

For readers of this PR: We always argue with eachother about the early-return thing. So I don't mind merging this without this "fixed" :smile_cat: !

matthiasbeyer:
@homu r+

PR #114, opened 2016-01-17
Store::new(): Create path if nonexistent

matthiasbeyer:
`Store::new()` behaviour if the store does not exist yet.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -58,7 +58,11 @@ impl Store {
         use std::fs::create_dir_all;
 
         if !location.exists() {
-            create_dir_all(location.clone()).ok(); // TODO: Error handling?
+            let c = create_dir_all(location.clone());
+            if c.is_err() {
+                return Err(StoreError::new(StoreErrorKind::StorePathCreate,
+                                           Some(Box::new(c.err().unwrap()))));
+            }
         } else {
             if location.is_file() {
                 return Err(StoreError::new(StoreErrorKind::StorePathExists, None));
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -55,6 +55,14 @@ impl Store {
 
     /// Create a new Store object
     pub fn new(location: PathBuf) -> Store {
+        use std::fs::create_dir_all;
+
+        if !location.exists() {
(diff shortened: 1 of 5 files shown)


matthiasbeyer:
I consider this ready for review.

@TheNeikos , @neithernut

libimagstore/src/store.rs
```diff
@@ -116,7 +116,17 @@ pub struct Store {
...
+
+        if !location.exists() {
+            create_dir_all(location.clone()).ok(); // TODO: Error handling?
+        } else {
+            if location.is_file() {
+                return Err(StoreError::new(StoreErrorKind::StorePathExists, None));
+            }
+        }
+
         Store {
```
TheNeikos: Missing `Ok(`

libimagstore/src/store.rs
```diff
@@ -116,11 +116,21 @@ pub struct Store {
 impl Store {
 
     /// Create a new Store object
-    pub fn new(location: PathBuf) -> Store {
-        Store {
+    pub fn new(location: PathBuf) -> Result<Store> {
+        use std::fs::create_dir_all;
+
+        if !location.exists() {
+            create_dir_all(location.clone()).ok(); // TODO: Error handling?
```
neithernut: I would definitely do error handling in this case

matthiasbeyer:
@homu retry

neithernut:
ACK.

matthiasbeyer:
@homu r+

matthiasbeyer:
@TheNeikos Would appreciate your review. Especially on the cargo.lock update thing... I'm not sure whether I should commit this, ... ?

TheNeikos:
Why did you just update Cargo.lock?

matthiasbeyer:
What do you mean by 'just'? There was noting more to update!

TheNeikos:
Usually the lock only updates if you change the toml.

matthiasbeyer:
Maybe it was left out at some point?

TheNeikos:
Actually, leave it in. It won't matter as I will submit a PR that removes it since as a library one should not 
publish the lock as well.

matthiasbeyer:
Sounds good.

matthiasbeyer:
@TheNeikos ack?

TheNeikos:
A Cee kay

PR #118, opened 2016-01-18
Libimagstore/header utils

matthiasbeyer:
Tracking branch for header-parser code utilities in `EntryHeader` type.

### DIFF
--- a/libimagstore/Cargo.toml
+++ b/libimagstore/Cargo.toml
@@ -5,8 +5,9 @@ authors = ["Matthias Beyer <mail@beyermatthias.de>"]
 
 [dependencies]
 fs2 = "0.2.2"
-toml = "0.1.25"
+regex = "0.1.47"
 semver = "0.2"
+toml = "0.1.25"
 
 [dev-dependencies]
 tempdir = "0.3.4"
--- a/libimagstore/Cargo.toml
+++ b/libimagstore/Cargo.toml
@@ -8,6 +8,7 @@ fs2 = "0.2.2"
 regex = "0.1.47"
 semver = "0.2"
 toml = "0.1.25"
+version = "1.1.0"
 
 [dev-dependencies]
 tempdir = "0.3.4"
--- a/libimagstore/src/header.rs
+++ b/libimagstore/src/header.rs
@@ -1,4 +1,74 @@
+use std::error::Error;
+use std::result::Result as RResult;
+
 use toml::Table;
+
+pub mod error {
+    use std::fmt::{Debug, Display, Formatter};
+    use std::fmt;
+    use std::error::Error;
+    use toml;
+
+    #[derive(Clone)]
+    pub enum ParserErrorKind {
+        TOMLParserErrors,
+        MissingMainSection,
+    }
+
+    pub struct ParserError {
+        kind: ParserErrorKind,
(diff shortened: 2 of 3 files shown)


matthiasbeyer:
Blocked by #122 afaict.

TheNeikos:
#122 has been merged

matthiasbeyer:
@TheNeikos You mean I can merge?

neithernut:
I'm wondering whether the internal error module could be merged with the general libstore error module in a sensible way.

matthiasbeyer:
Nah, I don't like the idea. The error module I'm introducing is header related only (it is actually only header-parser related only), so it is a completely other domain, IMHO.

neithernut:
Wouldn't `MissingMainSection` and `MissingVersionInfo` also make sense in another context? E.g. when preparing/creating an entry?

matthiasbeyer:
@neithernut indeed. Though I'm not sure here how to proceed. We want `::new()` to return a type, not an `Option<type>`, so changing the `::new()` function to do such a check won't be good. I'm not sure how to do verification on creation...

neithernut:
@matthiasbeyer of what `::new()` are you talking about, exactly?

matthiasbeyer:
When creating an `Entry`, we want `Entry::new()` return `Entry`, not `Result<Entry>` ...

neithernut:
Ah, now I get what you problem is. Mh... I guess you don't have a choice here.
Another thing which just popped into my mind: should an `Entry` be created by the user (vs. created and returned by the store)? Think about a user creating an entry which already exists in the store.

matthiasbeyer:
Let us ask @TheNeikos

TheNeikos:
I am not sure about what?

matthiasbeyer:
> shoukd an Entry be created by the user vs. Created and returned by the store?

TheNeikos:
Returned by the store

matthiasbeyer:
Okay, so I will not implement `::new()` on `Entry`. :smile:

TheNeikos:
See #131

matthiasbeyer:
#131 is about `EntryHeader::new()`, not `Entry::new()`, ... or what do you mean?

TheNeikos:
Same reasoning, don't let the user create one explicitely.

matthiasbeyer:
I have not implemented any `::new()` in this PR, except one for an error type... So I'm not sure what you want from me...

TheNeikos:
Nothing :D But I just like to crossref things to keep discussions together

matthiasbeyer:
And what is the conclusion on this PR? Ack?

neithernut:
If the error types I mentioned are not being put into `libstore`'s general error module.

matthiasbeyer:
Nope, I wouldn't do that, as the domain of the errors is different. The error (and its `..Kind` will be exported through `StoreError.cause` anyways, so that wouldn't be an issue, but putting the error code into the main error module does not make sense to me in a semantical way.

matthiasbeyer:
@neithernut Do you ACK? Homu is dead, so I have to merge manually. Actually, I have to rebase first, but I think I won't change any content anymore.

neithernut:
ACK.

Issue #120, opened 2016-01-18
Handling of content


neithernut:
I do think @matthiasbeyer, @TheNeikos and I agree on that an entry may contain a chunk of textual content, maybe as a MIME container or something similar, as well as optionally refer to items outside the store via URIs. The rationale is that you can embed small datasets (e.g. appointments) in the store itself but refer to data which can not/should not be embedded into the store nicely (e.g. media files). Another possibility opening up would be to provide a "sync" address via an URI (e.g. to a webdav) for a piece of embedded data.
The current `entry` type defined in the store is not yet capable of serving any of these features. Things which have to be done are:
 * [x] define how URIs are provided in the header
 * [x] provide methods for retrieval of the URIs
 * [x] provide means to store the ~~MIME-type of the data~~ type of text embedded. ~~either by a similar declaration within the header or by using/writing a MIME-container for the `content`~~
 * [ ] clarify what the semantics are on some operations. When getting the content, for example, do we return an empty string if no data is embedded in the item file itself or do we try to fetch the data via one of the URIs?

TheNeikos:
I am not sure how much you want to tie the Store to the Data. I don't think it is a good idea to go down that road.

There is one exception I am willing to make though: The header should have a special table called 'content' that if present tells the store that it doesn't care about the content on disk as it is only a reference. (And the rest of the data stays in the header anyway.)

An example:

### An entry generated by an `imag-deincylinder` module

```toml
---
[imag]

[imag.content]
url = "https://www.deincylinder.com/watch?v=randomid"

mime = "video/mp4"
file = "/home/neikos/Videos/deincylinder/winter-wrap-up.mp4"

[youtube]
rating = 9 # Out of 10
viewed = 958 # Times
---
```

So there are two things going on, `imag.content.url` is the official location of the resource, it is the only __required__ entry. The other two are module specific, but I think that we now have the chance for good defaults to increase interoperability. For example in my use cases I will mostly want to use it to save pictures/Videos/Music I see, classify them and download them neatly so that I can easily retrieve them offline later on (Should the URL expire, it be taken down or just no internet connection).

# Advantages 

You liberate the Store of a lot of heavy lifting that might not be needed in the end. The store itself doesn't need to know the MIME type for example.

# Disadvantages

Each module would need to separately parse and check the entries. In my opinion this makes for more code, but also gives a lot of flexibility regarding the structure of the external content. However, the need for a 'best practices' guide is very apparent to make sure that common types are not fragmented.

[...]

Issue #58, opened 2016-01-05
Multiple binaries


matthiasbeyer:
I would like to split up the modules and the store into several modules, where one binary can use other binaries as libraries.

For example if we have one binary `imag-store`, this one implements an interface to the store. It can be used by the user from the commandline to inspect the store, but also exports a library which can be used by (for example) `imag-bm` to access the store.

This would be the git model of how the things are seperated. Additional thoughts must be put into this, as I'm not sure whether the binaries (`bm`, `notes`,...) should actually call the `store` binary or whether the `store` binary should be compiled into the others as a library.

Anyways, this is just an issue to remember this and collect some ideas. This is not achievable in the _near_ future.

TheNeikos:
I've thought about it and this is what got out in the end:

---

One thing one wants to avoid when making applications is to nedlessly repeat yourself. Thus, making a library first and then a binary wrapper around it makes sense. 

That way:

1. The important parts of the library are exported
2. One can easily manage dependencies between library versions, since you compile against another, the binary you get is _final_

**A problem appears though** The store is not static, items evolve, new ones get added, old ones renamed or removed. Thus there is a need for versioning, `Semver` seems to be a sane choic. The first idea would be to namespace the modules, so if you had the module `ferhnrohrer` at version `0.3.3`, it's items would be at `fernrohrer-v0.3.3`. 

However, this could easily fragment the whole database to obscene levels. Imagine having dozens of different libraries, each potentially talking to different versions of stores! This could work if the Store does some more heavy lifting. If one imagines that the store reserves the first level and only allows abstract paths from there on, one could imagine it like this:

```rust
    module_path!("fernrohrer", Semver::parse("0.3.2")); # Creates a struct called ModulePath    

    // Somewhere in a function

    let store = Store::new();
    let item = store.get_item(ModulePath::new("search/do_snake_ballons_hiss_when_scared"));
    let all_items = store.get_all_for(ModulePath::root());
```

This StorePath would also automatically implement `IntoStoreId` making it painless to use, but still gives enough flexibility for the next point.

The store, in `get_item` would take out the `module` part (the first argument to the macro) and search for all folders in the root folder using a regex akin to `{modulename}-[^-]*"`. It then filters out all *incompatible* files and matches those against the store path itself.

For `get_all_for` it would simply merge the folders together and return the array like that, the incompatible ones are marked as "NOT COMPTIBLE", using an enum would make sense.

[...]

PR #121, opened 2016-01-19
[RFC] Add content type

neithernut:
This PR makes `EntryContent` a container holding the type of the (textual) content.
Related to one aspect discussed in #120.


TheNeikos:
I do not think that there should be a content type as so much as a hint to the library that created it, thus leaving it up to the library to make such an entry.

So I do not think that that this belongs in the store itself.

neithernut:
I'm not sure I get your point. You mean the `EntryContentType` is in the wrong place? Or the content-type should be stored in a completely different way?

TheNeikos:
I think that the content type should not be stored by the store at all. It itself does not need any notion of what the file is it points to, so why encode it into the store library?

neithernut:
You may have missed something: both @matthiasbeyer and I pointed out that, for textual content stored in the item file itself, you need a way to tell how the text is formatted or whether it is formatted at all (hence `Plain`).
Viewers may rely on this because, in contrast to external files, there is no single hint of the text format, except looking at it and taking an educated guess.

matthiasbeyer:
@neithernut Wait for the doc-PR of @TheNeikos . We discussed and things got cleared up, just wait until the PR is there and then we can argue about details, but things are defined already (though not carved into stone).

TheNeikos:
Since store items should only be loaded through their respective modules (as we've discussed in #58) shouldn't that mean that whoever loads the file leave it to the respective library? 
I would wait until #58 iss resolved and then resume here.

neithernut:
Item files are loaded via the store only. At least that's what I thought. Waiting for the PR now...

matthiasbeyer:
I guess we can close this unmerged as this is done in a completely different way. @TheNeikos ?

TheNeikos:
I think so yes, hence my initial comments

PR #122, opened 2016-01-19
Doc/redefine header

matthiasbeyer:
As proposed in personal discussion with @TheNeikos

### DIFF
--- a/doc/src/02000-store.md
+++ b/doc/src/02000-store.md
@@ -16,28 +16,41 @@ The content follows the header-close marker (@sec:thestore:fileformat:content).
 ### Header Format {#sec:thestore:fileformat:header}
 
 The header format MUST BE "TOML".
-The contents of the header contain
+The sections which MAY or MUST be in the header are defined in the following
+chapters.
 
-1. A section called "imag", where the automatically by the program generated
-   data goes to.
-   The contents of these sections are edited via commandline calls or by the
-   program implicitely and SHOULD NOT be edited by the user.
-   Modules of the program are free to store arbitrary data here.
-   If a module stores data in the header of a file it MUST do that in a
-   dedicated section, as TOML supports it.
-   Exceptions are:
+#### Header section: "imag" {#sec:thestore:fileformat:header:imag}
 
-    * A module MAY collect URIs and put it into a flat list in the subsection
-      "imag.links".
-    * A section "imag.content" MAY BE used for refering to external content.
-      Valid keys in this section are "uri", "file" and "mime", whereas
-        - "uri" refers to the external content
-        - "file" refers to a local variant, this is optional
-        - "mime" stores the MIME-Type of the local variant, if present.
+The header MUST contain a section called "imag", where the automatically by the
+program generated data is stored in.
+The contents of this section is edited via commandline calls or by the
+program implicitely and SHOULD NOT be edited by the user.
 
-1. Other OPTIONAL sections which are named and edited by the user. The program
-   MUST NOT touch the contents of these sections, except explicitely asked by
-   the user to do so.
+This "imag" section MUST contain the following keys
+
+1. A "version" Key. The version stored here is the version of the Store, the
+   Entry was created with.
+
+The "imag" section MAY contain
+
+1. A section "imag.links" where a module is allowed to store URIs in a flat
+   list
+1. A section "imag.content", used for refering to external content.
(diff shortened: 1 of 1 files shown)


doc/src/02000-store.md
```diff
@@ -16,21 +16,41 @@ The content follows the header-close marker (@sec:thestore:fileformat:content).
...
+
+The "imag" section MAY contain
+
+1. A section "imag.links" where a module is allowed to store URIs in a flat
+   list
+1. A section "imag.content", used for refering to external content.
+   Valid keys in this section are "uri", "file" and "mime", whereas
+     - "uri" refers to the external content
+     - "file" refers to a local variant, this is optional
+     - "mime" stores the MIME-Type of the local variant, if present.
```
neithernut: This is redundant, as you can simply use the `file://` scheme to refer to local files.

doc/src/02000-store.md
```diff
@@ -16,21 +16,41 @@ The content follows the header-close marker (@sec:thestore:fileformat:content).
...
+
+#### Header section: "module" {#sec:thestore:fileformat:header:module}
+
+The header MAY contain a section named after a module.
+The corrosponding module is allowed to store arbitrary data in this section.
```
neithernut: The title of this section is misleading as it suggests that a TOML-section _named_ `mudule` does exist. I propose renaming this section to something like `Module sections` or `Header section: module sections`.

matthiasbeyer: Thanks. I searched for a better naming but failed. Will name appropriately and move below the "custom" section.

matthiasbeyer:
Ready for cleanup.

neithernut:
`file` and `uri` are still redundant.

matthiasbeyer:
No. If you reread the section, `uri` is for pointing _anywhere_ but `file` may point to a local variant.

I'm not that sure whether we want that. @TheNeikos what do you think?

neithernut:
In this case `refers to the external content` is ambogious.

neithernut:
Also: omit the `the`

matthiasbeyer:
Hm, no. I can see now what you mean. Maybe @TheNeikos can point out what he meant by `file` ...

doc/src/02000-store.md
```diff
@@ -16,21 +16,40 @@ The content follows the header-close marker (@sec:thestore:fileformat:content).
...
+
+The header MUST contain a section called "imag", where the automatically by the
+program generated data is stored in.
+The contents of this section is edited via commandline calls or by the
+program implicitely and SHOULD NOT be edited by the user.
+
+This "imag" section MUST contain the following keys
+
+1. A "version" Key. The version stored here is the version of the Store, the
+   Entry is compatible with.
```
TheNeikos: I would say that it was created with this version of the store, compability is not necessarily encoded in the version.

matthiasbeyer:
@TheNeikos What about the `file` member, @neithernut and I argued over above?

TheNeikos:
The idea was that you have an `url` field that is required, and then the module can put in w/e it wants there. Then you would put in optional fields with well-defined meanings to make sure libraries understand each other, like for example `file` and `mime` meant for example for a local representation of the entry.

matthiasbeyer:
@TheNeikos as suggested...

doc/src/02000-store.md
```diff
@@ -16,21 +16,40 @@ The content follows the header-close marker (@sec:thestore:fileformat:content).
...
+
+1. A "version" Key. The version stored here is the version of the Store, the
+   Entry was created with.
+
+The "imag" section MAY contain
+
+1. A section "imag.links" where a module is allowed to store URIs in a flat
+   list
+1. A section "imag.content", used for refering to external content.
+   Valid keys in this section are "uri" and "file", whereas
```
TheNeikos: only file is required, *valid* is everything needed by the library that it needs.

neithernut:
Would a URI pointing to something local still be valid?

TheNeikos:
Yup

neithernut:
In this case the semantics kinda make sense. Maybe I'll come back with some clarifications at some point. But for now I can ACK this.

matthiasbeyer:
@TheNeikos do you ACK as well?

TheNeikos:
Yes

matthiasbeyer:
@homu r+

PR #123, opened 2016-01-19
Doc/redefine store orga

matthiasbeyer:
As proposed in personal discussion with @TheNeikos

### DIFF
--- a/doc/src/02000-store.md
+++ b/doc/src/02000-store.md
@@ -97,4 +97,32 @@ sub-folders like this:
 /module/some/sub/folder/example~0.1
 ```
 
+## Store path links {#sec:thestore:links}
+
+Linking entries MUST BE version independent.
+
+This means if an entry "a" from a module "A" gets written to the store, it may
+link to an entry "b" from a module "B", which is in version "0.1" at the moment.
+If the module "B" gets updated, it might update its entries in the store as
+well.
+The link from the "a" MUST NOT get invalid in this case.
+
+This is accomplished by linking without the version number: So a link for the
+entry
+
+```
+/module/some/sub/folder/example~0.1
+```
+
+is
+
+```
+imag://module/some/sub/folder/example
+```
+
+As shown in the example, a link to imag-internal entries, the link is prefixed
+with a "imag://" identifier.
+A link to external content MUST NEVER be prefixed this way.
+The path of the internal link MUST NEVER be relative, but always absolute from
+the root directory of the store.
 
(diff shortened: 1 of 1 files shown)


neithernut:
ACK.

doc/src/02000-store.md
```diff
@@ -71,11 +71,50 @@ So if the store exists in `/home/user/store/`, a file with the Storepath
...
+
+is
+
+    "imag://module/some/sub/folder/example"
+
+As shown in the example, a link to imag-internal entries, the link is prefixed
+with a "imag://" identifier.
+A link to external content MUST NEVER be prefixed this way.
+The path of the internal link MUST NEVER be relative, but always absolute from
+the the root directory of the store.
```
TheNeikos: "the the"

matthiasbeyer:
Ready for cleanup.

TheNeikos:
There is one part that might bite us in the butt.

Semver allows identifiers with a dash, akin to `2.3-alpha` how do we handle that?

matthiasbeyer:
@TheNeikos Simply allow it. I mean... why not? The version information is a postfix, so there isn't anything wrong with that, is there? AFAIK, the first character of a semver is always a number...

neithernut:
@matthiasbeyer then what about `blink-182-2.0`?

TheNeikos:
What @neithernut said. I think we should use something else than a dash. A possible one would be `~`.

So we would just search for the last one and evverything after that is the version.

matthiasbeyer:
@TheNeikos Sounds reasonable. I'm not sure what characters are allowed by FS, so we can argue here on which character would fit. A tilde sounds good to me, though.

We have to restrict the interface so that a filename in the store cannot end with this character, then. This should go into the paper. Tell me where.

TheNeikos:
A tilde is an allowed character under btrfs, not sure about ntfs tho

matthiasbeyer:
[Tidle is in 7-bit-ascii](http://www.neurophys.wisc.edu/comp/docs/ascii/#table) and I bet [all major FS allow 7-bit ascii as filenames](https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations).

neithernut:
A tilde is also used by some editors and office suites for backups/swap files. Even Windows supports a tilde in filenames (not that I care).

TheNeikos:
ACK from me

neithernut:
ACK from me, too.

matthiasbeyer:
Squashed into two commits.

@homu r+

matthiasbeyer:
@homu r+

PR #131, opened 2016-01-22
Dont pass the table to the header object, but generate it in ::new()

### LINKED ISSUE #127: store: EntryHeader should build a standard-header
I guess the `EntryHeader::new()` function should not get a `table` passed but build its own one and set some default entities within.

What do you think?

This is an interface-change!

matthiasbeyer:
Closes #127 

I will not merge this yet, as we have to build the default-header. I will implement it as far as the docs say by now, but maybe we want more...

### DIFF
--- a/libimagstore/src/header.rs
+++ b/libimagstore/src/header.rs
@@ -1,7 +1,11 @@
+use std::collections::BTreeMap;
 use std::error::Error;
 use std::result::Result as RResult;
 
 use toml::{Table, Value};
+
+use self::error::ParserErrorKind;
+use self::error::ParserError;
 
 pub mod error {
     use std::fmt::{Debug, Display, Formatter};
@@ -68,10 +72,6 @@ pub mod error {
 
 }
 
-
-use self::error::ParserErrorKind;
-use self::error::ParserError;
-
 /**
  * EntryHeader
  *
@@ -92,10 +92,12 @@ impl EntryHeader {
 
     /**
      * Get a new header object with a already-filled toml table
+     *
+     * Default header values are inserted into the header object by default.
      */
-    pub fn new(toml: Table) -> EntryHeader {
+    pub fn new() -> EntryHeader {
         EntryHeader {
-            toml: toml,
+            toml: BTreeMap::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
It might be even better to implement `Default` and not providing a `::new()` method,... not sure.

@TheNeikos ?

TheNeikos:
Implementing `Default` here makes not a lot of sense in my opinion. I would just not make it public and instead let the entry create one (with the defaults)

matthiasbeyer:
So you would do it as I did in my PR? :smile:

TheNeikos:
Yes, except the 'hardcoded' part in the version

matthiasbeyer:
Yes, I really should change that to the `version!()` of the crate, right? :-)

libimagstore/src/header.rs
```diff
@@ -30,3 +36,20 @@ impl EntryHeader {
...
+
+fn build_default_header() -> BTreeMap<String, Value> {
+    let mut m = BTreeMap::new();
+
+    m.insert(String::from("imag"), {
+        let mut imag_map = BTreeMap::<String, Value>::new();
+
+        imag_map.insert(String::from("version"), Value::String(String::from("hardcoded")));
+        imag_map.insert(String::from("links"), Value::Array(vec![]));
+        imag_map.insert(String::from("content"), Value::Table(BTreeMap::new()));
```
TheNeikos: Don't put it in per default! It should only be present if the content is not in the file!

libimagstore/src/header.rs
```diff
@@ -30,3 +36,20 @@ impl EntryHeader {
...
 
 }
+
+fn build_default_header() -> BTreeMap<String, Value> {
+    let mut m = BTreeMap::new();
+
+    m.insert(String::from("imag"), {
+        let mut imag_map = BTreeMap::<String, Value>::new();
+
+        imag_map.insert(String::from("version"), Value::String(String::from("hardcoded")));
```
TheNeikos: hardcoded? Perhaps use version! or something similar?

matthiasbeyer:
@TheNeikos review pls.

TheNeikos:
ACK

matthiasbeyer:
Waiting on CI.

PR #135, opened 2016-01-23
rt: Make runtime module public

matthiasbeyer:
Make the `runtime` module of `libimagrt` public, so we can use the `Runtime` type.


neithernut:
ACK.

PR #136, opened 2016-01-23
bin/imag-store/init

matthiasbeyer:
Tracking branch for the user-interface implementation for the `libimagstore` library.

@TheNeikos comments on everything appreciated.


imag-store/src/create.rs
```diff
@@ -0,0 +1,72 @@
...
+        .value_of("from-raw")
+        .ok_or(StoreError::new(StoreErrorKind::NoCommandlineCall, None))
+        .map(|raw_src| {
+            let mut content = String::new();
+            if raw_src == "-" {
+                stdin().read_to_string(&mut content);
+            } else {
+                OpenOptions::new()
+                    .read(true)
+                    .write(false)
```
TheNeikos: It's false per default

libimagstore/src/store.rs
```diff
@@ -347,7 +347,7 @@ pub struct Entry {
 
 impl Entry {
 
-    fn new(loc: StoreId) -> Entry {
+    pub fn new(loc: StoreId) -> Entry {
```
TheNeikos: If you rebase this should disappear

matthiasbeyer:
@neithernut , @TheNeikos This is ready for review.

---

Some commits of this PR which introduce fixes in parts of the codebase are submitted as seperate PRs and will vanish as soon as this gets rebased.

CRUD works, though some issues still exist. I will report them as I find them, some of them have been reported already.

imag-store/src/create.rs
```diff
@@ -0,0 +1,115 @@
...
+                    .map(String::from)
+                    .unwrap_or(String::new())
+            })
+            .unwrap_or_else(|| {
+                debug!("Didn't find entry subcommand, getting raw content");
+                matches.value_of("from-raw")
+                    .map(|raw_src| content_from_raw(raw_src))
+                    .unwrap_or(String::new())
+            })
+    };
```
neithernut: This scope doesn't introduce any benefits, as it contains only one expression. Could you assign the expression directly?

imag-store/src/create.rs
```diff
@@ -0,0 +1,115 @@
...
+
+    rt.store()
+        .create(PathBuf::from(path))
```
neithernut: Having an explicit `StoreId` (or, in this case, a function which returns a `StoreId`) would be more expressive, and give us the freedom to replacing the actual implementation of `StoreId` in the backend without having to rewrite this.

imag-store/src/retrieve.rs
```diff
@@ -0,0 +1,81 @@
...
+
+                    }
+                })
+                .map_err(|e| {
+                    debug!("No entry.");
+                    debug!("{}", e);
+                })
+
+        });
+}
```
neithernut: This function is awfully long and nested. Care to split it up?

imag-store/src/create.rs
matthiasbeyer: So what do you propose? A PathBuf is absolutely valid, as it implements the `IntoStoreId` trait... don't know what you want...

neithernut: Essentially hide the fact that we are throwing paths around. One option would be to simply put the call to `build_entry_path()` directly on this site, instead of in `create()`.

matthiasbeyer:
I just pushed a few fixes and some more functionality cli-ui wise

(one can read either the _content_ from cli, stdin or an external file, read the whole entry from cli or file (for `cat /my/store/foo | imag store create --from-raw -`) and so on).

matthiasbeyer:
@TheNeikos Would like to hear what you think about the testing scripts in the latest two commits. The tests should fail, as the first test should fail, though I like to hear your opinion on the overall setup.

The `util.sh` script could be shared for all binaries for testing later on.

matthiasbeyer:
Ready for review here. @TheNeikos ?

matthiasbeyer:
Travis fails because the build output gets too long when calling `tree`, but also fails without the call to `tree` (I don't know why) ... any ideas how to do this?

TheNeikos:
Don't forget to include the version in the path or this will not work soon.

matthiasbeyer:
We actually don't want this in the store CLI interface. The store CLI interface is absolute low level access to the store and if a user has to use it, she has to specify absolutely everything. That's why I reverted the macro-call commit in dfee017.

TheNeikos:
Sure, the store *itself* does not want to export a module entry. However, if I want to create a file manually it needs to have a version associated with it. Which is what I wanted to point out.

matthiasbeyer:
Finally, this PR succeeds.

@TheNeikos and @neithernut review is appreciated, I guess this is ready now.

matthiasbeyer:
84c623d closes #178

imag-store/src/retrieve.rs
matthiasbeyer: This is done, though the diff annotation does not vanish.

neithernut:
Didn't see anything suspicious, yet.

TheNeikos:
Me neither

matthiasbeyer:
I will cleanup the history and merge after CI builds my stuff.

matthiasbeyer:
I just reworded the last three commits, as integrating them into history is a bit of a hassle. I will merge asa ci succeeds.

PR #140, opened 2016-01-23
Explain futureproofing store files

TheNeikos:


neithernut:
Er... could you put the new paragraphs below the example and not directly between the example and the text referencing it?

neithernut:
Now it's good. ACK.

PR #144, opened 2016-01-24
Fix: Add missing "pub" on Entry functions

matthiasbeyer:
Add missing `pub` keywords on a few `Entry` functions.


TheNeikos:
ack

Issue #103, opened 2016-01-17
Implement: Store::retrieve()


matthiasbeyer:

matthiasbeyer:
@TheNeikos

neithernut:
I just found an issue with the existing partial implementation of the `Store::retrieve()` method introduced in commit https://github.com/matthiasbeyer/imag/commit/d949cddc650fb2fd0f87d81307b11815e8230db0. The code currently
 * acquired a lock of the hashmap
 * tries to get an _existing_ store entry from the hashmap
 * gets an entry from the store entry
 * constructs a `FileLockEntry` from it

In other words: the current code for `store::retrieve()` presumes that the entry is already in the hash map, which is obviously wrong. Moreover, the code doesn't even check whether the existing store entry is borrowed or not.
The correct way to perform a retrieve would be to assert no entry with a matching id exists in the hashmap, instead of querying the entry.

----

It's kinda embarrassing that I didn't note this earlier.

neithernut:
IMO this also illustrates why having `get_entry()` on `StoreEntry` instead of having a private method on `Store` or an alternative constructor on `Entry` (taking a serialized entry) provides a fallacy people fall for.

neithernut:
I also just noticed that we have `StoreErrorKind::IdLocked` and `StoreErrorKind::EntryAlreadyBorrowed`, which mean the same thing.

neithernut:
Mh... this might even be used as a prime example why even the strictest type system, which saves your butt most of the time, can still be improved by statically proven contracts.

PR #145, opened 2016-01-24
Fix Store::retrieve()

neithernut:
This PR targets a primary issue laid out in #103.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -142,22 +142,16 @@ impl Store {
     /// Borrow a given Entry. When the `FileLockEntry` is either `update`d or
     /// dropped, the new Entry is written to disk
     pub fn retrieve<'a>(&'a self, id: StoreId) -> Result<FileLockEntry<'a>> {
-        // get hold of the hashmap
-        let hsmap = self.entries.write();
-        if hsmap.is_err() {
-            return Err(StoreError::new(StoreErrorKind::LockPoisoned, None))
-        }
-
-        // either get the existing entry ot create it
-        let mut entries = hsmap.unwrap();
-        let mut store_entry = entries.entry(id.clone()).or_insert_with(|| StoreEntry::new(id.clone()));
-
-        // update the store entry status
-        store_entry.status = StoreEntryStatus::Borrowed;
-
-        // make a file lock entry from the store entry
-        store_entry.get_entry().map(|entry| FileLockEntry::new(self, entry, id))
-    }
+        self.entries
+            .write()
+            .map_err(|_| StoreError::new(StoreErrorKind::LockPoisoned, None))
+            .and_then(|mut es| {
+                let mut se = es.entry(id.clone()).or_insert_with(|| StoreEntry::new(id.clone()));
+                se.status = StoreEntryStatus::Borrowed;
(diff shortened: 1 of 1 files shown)


neithernut:
Note: the way the store entry status is currently updated is inadequate. Maybe I'll come up with a better solution, maybe I'll move the contents of `get_entry()` to another place.

matthiasbeyer:
@neithernut Have a look at my branch https://github.com/matthiasbeyer/imag/tree/pr-145-cleanup-suggestions , maybe you want to ff-merge my suggestions? It was hacketyhackhack and it compiles, though I didn't review myself.

neithernut:
Apparently, it doesn't compile. I'll have a look at it later.

matthiasbeyer:
ah. Remove the 0443a0d commit, this breaks the build.

neithernut:
I thought so, but I was not sure how important forwarding the error was to you. Personally, I think that panicking is acceptable if the store itself (or a locking mechanism within the store) is broken.

matthiasbeyer:
Forwarding it would be nice, though I don't mind it that much, as the callee can rather clearly see what happened here. Panicing isn't an option as we are writing a library. I guess you mean returning an error.

neithernut:
No, I meant panicking. But in this case, the `delete()` method must also be adjusted.

TheNeikos:
No, you don't want to forward that error.

matthiasbeyer:
@TheNeikos ACK here?

libimagstore/src/store.rs
```diff
@@ -142,15 +142,16 @@ impl Store {
...
-            .and_then(|store_entry| store_entry.get_entry())
-            .and_then(|entry| Ok(FileLockEntry::new(self, entry, id)))
-    }
+        self.entries
+            .write()
+            .map_err(|_| StoreError::new(StoreErrorKind::LockPoisoned, None))
+            .and_then(|mut es| {
+                let mut se = es.entry(id.clone()).or_insert_with(|| StoreEntry::new(id.clone()));
+                se.status = StoreEntryStatus::Borrowed;
+                se.get_entry()
```
TheNeikos: This will always fail because it is already borrowed at this point.

TheNeikos:
Nah I don't ACK this, and I didn't have the time to write up an answer to the previous critique which is valid in the current context but is easily explained because the code itself is in an intermediary state.

libimagstore/src/store.rs
neithernut: Yep, he's right. Now I also remember why I had that many bindings.

neithernut:
So, you generally won't ACK this PR (and come up with a fix of your own)? Or can you point out concrete things you want to have changed?

TheNeikos:
After discussing with @matthiasbeyer what he would like to see in the code, which is radically different than what I envisioned, I can ACK this.

neithernut:
Done.

matthiasbeyer:
AFAIK, @TheNeikos doesn't consider this ready.

TheNeikos:
I'm fine with this.

matthiasbeyer:
So you want me to merge this?

TheNeikos:
Aye

PR #147, opened 2016-01-25
Add missing error handling

neithernut:
We don't want to panick if a lock inside the store is broken. We want to
notify the user, so she can start panicking.


matthiasbeyer:
Ack. Waiting on CI.

matthiasbeyer:
60 minutes later...

neithernut:
Looks like Travis is on strike.

Issue #105, opened 2016-01-17
Implement: Store::retrieve_copy()


matthiasbeyer:

neithernut:
I guess we should wait for a resolution of #115 before implementing this.

matthiasbeyer:
Definitively!

neithernut:
This issue looks outdated.

PR #148, opened 2016-01-25
Implement Store::retrieve_copy()

neithernut:
Targets #105.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -187,7 +187,19 @@ impl Store {
     /// Retrieve a copy of a given entry, this cannot be used to mutate
     /// the one on disk
     pub fn retrieve_copy(&self, id: StoreId) -> Result<Entry> {
-        unimplemented!();
+        let mut entries_lock = self.entries.write();
+        if entries_lock.is_err() {
+            return Err(StoreError::new(StoreErrorKind::LockPoisoned, None))
+        }
+
+        let mut entries = entries_lock.unwrap();
+
+        // if the entry is currently modified by the user, we cannot drop it
+        if entries.get(&id).map(|e| e.is_borrowed()).unwrap_or(false) {
+            return Err(StoreError::new(StoreErrorKind::IdLocked, None));
+        }
+
+        StoreEntry::new(id).get_entry()
     }
 
     /// Delete an entry


matthiasbeyer:
@TheNeikos what about this one?

TheNeikos:
:? It doesn't copy!

neithernut:
Could you elaborate?

matthiasbeyer:
@TheNeikos ping.

TheNeikos:
Nevermind, I misread I think

matthiasbeyer:
@TheNeikos so we can merge?

TheNeikos:
I think so, perhaps a test :3?

matthiasbeyer:
@TheNeikos I'm not sure about testing here... FS tests are a bit of a pain on travis, you remember?

neithernut:
So... merge?

TheNeikos:
There is one thing I would like to discuss beforehand, that is when it is locked, then I think it should return the 'old' copy.

neithernut:
Interesting idea. It may not be possible in all cases (e.g. if the file is locked by another process) and it would require us to cache the "old" version of the entry (which is doable and may even bring some other benefits with it).

TheNeikos:
I would just try to read the file and if that fails, fail as well.

neithernut:
Nah, reading a locked entry may give you inconsistent data. You don't want that. At least not silently.

neithernut:
Mh... after fiddeling around a little, I did not find any obvious solution which would not require either some redesign of `StoreEntry` and existing methods or impose some ugliness. I'd opt for postponing the extension @TheNeikos proposed. Maybe he could, instead, open an issue where we discuss whether we want this at all and what behaviour/impacts would be acceptable.

TheNeikos:
:+1:

neithernut:
@matthiasbeyer: the current state may still be merged as-is.

matthiasbeyer:
Still would like to see some tests :smile:

neithernut:
Well, that would require me coming up with a test-setup (remember: this function _does_ read from a file, in the end). For this reason, tests of other store methods are also still missing (at least I didn't find any). So... would you _really_ like me to implement testing of essentially store functionality in general in this PR?

matthiasbeyer:
Ah, right... we dicided to not having tests in this case as FS-IO tests suck a bit on travis. Sorry for the inconveniance!

PR #149, opened 2016-01-27
Add header field addressing by str

matthiasbeyer:
I guess something like this would be nice for the modules, so a module does not need to "crawl" through the header, but has a function which does it for the module.

What do you think?


matthiasbeyer:
Note: This will, of course, only work for Table types,... setting arrays won't work (or isn't planned. I _could_ implement this, but this would be almost a DSL then... if you want, I'd happily do this).

Maybe this could also be done as a macro, but I'm sure I am not good enough to implement such a thing as macro.

matthiasbeyer:
After thinking a bit: I guess it would be even better to implement this as macro, so we can fail at compile time if the `str` does not compile to a sane address string, right? @TheNeikos What do you think?

matthiasbeyer:
Maybe this (as macro) would make more sense if we put this into a dedicated crate or submit it to `toml-rs` if such a feature is desired there. @TheNeikos challenge?

matthiasbeyer:
#170 implements such a thing as well..

matthiasbeyer:
@TheNeikos can you help me here?

I need a function

```rs
fn walk_tokens(&mut self, tokens: Vec<Token>) -> Result<&mut Value>;
```

but I fail to implement it.

The idea is: I have a `Vec<Token>` and I need to recurse into the `self.toml` to get the value out of it.

matthiasbeyer:
@TheNeikos Review appreciated. I changed the interface of the `EntryHeader::{insert, set, read}` functions, and it compiles for me now... we have to test it, though.

matthiasbeyer:
I had to change the internals of the `EntryHeader` type, which resulted in removing two public functions. @TheNeikos I guess we want this. The change was 75068c6, which made a lot simpler for me.

libimagstore/src/store.rs
```diff
@@ -327,27 +333,16 @@ impl EntryHeader {
...
         EntryHeader {
-            toml: t
+            header: Value::Table(t)
         }
     }
 
-    /**
-     * Get the table which lives in the background
-     */
-    pub fn toml(&self) -> &Table {
```
matthiasbeyer: We might want to argue here.

I'd propose the following implementation for this function:

```rs
pub fn toml(&self) -> &Table  {
    match &self.header {
        &Value::Table(ref t) => &t,
        _ => panic!("something really weird just happened"),
    }
}
```

Same goes for the `toml_mut()` function.

@TheNeikos @neithernut objections? 

---

The whole point is:

Moving from `Table` to `Value` as internal type made implementation of the `insert()`, `set()` and `read()` functions _much_ simpler. Though, we still want to export the underlying `Table` object, not the `Value` itself, as this would cause some boilerplate code for all users. Thus: `panic!()`. Is it okay here or not?

matthiasbeyer: Should work for the non-mutable version, but seems to be not possible for the mutable one.

matthiasbeyer: @TheNeikos Maybe it would make sense to implement `Into<Value::Table>` and leave the destructuring and re-creating to the user of the library? I'm not sure... :confused:

TheNeikos: I see no problem with giving the user the table itself

matthiasbeyer:
@TheNeikos this is ready for merge, can you review?

TheNeikos:
This looks good to me :+1:

PR #153, opened 2016-01-28
Libimagstore/add debugging output

matthiasbeyer:
Add debugging output to store crate.


matthiasbeyer:
I would like to get some review here.

PR #159, opened 2016-01-29
Libimagutil/reimplement kv split

matthiasbeyer:
Simplifies the implementation of the string splitter and fixes bugs with the complex regex by using two distinct regexes.

We also do not support single quotes anymore, so you cannot do:

```
--header foo='bar'
```

but you have to

```
--header foo="bar"
```

@TheNeikos suggested this, as TOML uses double quotes as well, so I think it is okay to stick to this scheme.

### DIFF
--- a/libimagutil/src/key_value_split.rs
+++ b/libimagutil/src/key_value_split.rs
@@ -75,8 +75,7 @@ mod test {
 
     #[test]
     fn test_single_quoted() {
-        let s = String::from("foo='bar'").into_kv().unwrap();
-        assert_eq!(KeyValue::new(String::from("foo"), String::from("bar")), s);
+        assert!(String::from("foo='bar'").into_kv().is_none());
     }
 
     #[test]
--- a/libimagutil/src/key_value_split.rs
+++ b/libimagutil/src/key_value_split.rs
@@ -75,7 +75,8 @@ mod test {
 
     #[test]
     fn test_single_quoted() {
-        assert!(String::from("foo='bar'").into_kv().is_none());
+        let s = String::from("foo='bar'").into_kv().unwrap();
+        assert_eq!(KeyValue::new(String::from("foo"), String::from("\'bar\'")), s);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
AFAICS, we don't need the `log` crate here. I planned to reimplement the function with `debug!()` macros, but didn't do it. If you want me to remove the `log`-crate adding commit, tell me. Otherwise I could add some debugging output, if you want me to.

matthiasbeyer:
The latest two commits should fix the tests (one test was adapted, as we do not support single quotes anymore, the other commit fixes the code for a bug which was caught by the tests (yay!))

libimagutil/src/key_value_split.rs
```diff
@@ -42,16 +42,30 @@ pub trait IntoKeyValue<K, V> {
...
-                        .or(cap.name("DOUBLE_QVAL"))
-                        .or(cap.name("VAL"))
-                        .map(|value| KeyValue::new(String::from(name), String::from(value)))
-                }).unwrap_or(None)
+        let key = {
+            let r = "^(?P<KEY>([^=]*))=(.*)$";
+            let r = Regex::new(r).unwrap();
+            r.captures(&self[..])
+                .map(|caps| {
+                    caps.name("KEY").unwrap_or("")
```
TheNeikos: If you do this then on line 65 you will try to use an empty string as a key, which is not valid. Removing the unwrap or should work here.

libimagutil/src/key_value_split.rs
```diff
@@ -42,16 +42,30 @@ pub trait IntoKeyValue<K, V> {
...
+                })
+        };
+
+        let value = {
+            let r = "(.*)=(\"(?P<QVALUE>([^\'^\"]*))\"|(?P<VALUE>([^\'^\"]*)))$";
```
TheNeikos: This does not allow for constructs like these:
`foo="Neikos' Home"` or `foo=arr'veridisdasdsdadas`

matthiasbeyer:
Removed the `log`-crate introducing commit.

matthiasbeyer:
For documentation purposes: this was extracted from #136

TheNeikos:
LGTM

PR #161, opened 2016-01-29
Fix: Use env(HOME) instead of "~"

### LINKED ISSUE #160: Store creates "$(PWD)./~/.imag" runtimepath is "~/.imag"
This is an issue with the `libimagrt`, though it is also related to the store, as the error occurs in the `libimagstore` code, but originates in the `libimagrt` codebase.

matthiasbeyer:
Closes #160


neithernut:
Sooo much Ack. Should have seen this myself.

PR #164, opened 2016-01-29
Bin/imag

matthiasbeyer:
Implements a `imag` binary which is simply a shell script to call the actual real binary.

This should wait until #136 is merged, at least.


bin/imag
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
```
neithernut: Have you looked in how portable this is? May generating the shebang at compile/install time be an option?

neithernut: Also: maybe we can ask a certain student who likes scripting in shell and perl to have a look at making this work in `sh`-only environments?

neithernut: Mh... according to [wikipedia](https://en.wikipedia.org/wiki/Shebang_%28Unix%29#Portability) there are at least some systems on which the `/usr/bin/env` workaround doesn't work.

matthiasbeyer: For the first point: It should be really portable, yes. The `/usr/bin/env` binary should be available on all UNIXoids (it even is available on NixOS - the only binary in `/usr/bin`).

For the second point: @manuelmessner ping.

matthiasbeyer: I guess `/usr/bin/env` is good enough for us. It is rather portable and standard compliant. For everything else, the packager is responsible.

neithernut: Agreed.

matthiasbeyer:
@neithernut , @TheNeikos feel free to suggest additions for the helptext. I don't care whether it is 50 lines or 500 lines!

(And no, @neithernut, I don't like the idea of providing manpages at this point in time. Not before we get to something `1.0.0`ish for the whole suite! :smile:)

bin/imag
manuelmessner: `/bin/sh` should work on almost all systems. In fact, the only system I can remember of, which has a non standard path to `sh` is solaris.
`/bin/sh` itself is usually a link to either `dash` or `bash` => expect a `dash` since it has a smaller feature set.

Regarding the `bash` executable:
Most scripts start either with `/bin/bash` or `/usr/bin/bash`. 
Some projects are using `/usr/bin/env`.
I do not know which one is the most portable here, but counting the occurences `/bin/bash` and `/usr/bin/bash` win. I would prefer the first one.

matthiasbeyer: I'd not prefer the first one. `/bin/bash` is (for example) not available on my system, but `/bin/sh` is (which links to a bash executable).

manuelmessner: > I would prefer the first one
If you want to use bash. 

Otherways it's definitly `/bin/sh`, as I said in my first sentence above.
That should work almost everywhere.

matthiasbeyer: Relevant: https://www.reddit.com/r/linuxquestions/comments/43j624/how_portable_is_usrbinenv_as_shebang/

matthiasbeyer:
**Finally** this succeeds.

matthiasbeyer:
@TheNeikos Please do a careful review of my `.travis.yml` changes!

TheNeikos:
LGTM

PR #170, opened 2016-02-02
Libimagentryfilter/init

matthiasbeyer:
New library to build filters for entries.

Covers:

- [ ] Filter types
  - [x] Callable for the user with `entries.iter().filter(|e| filter.filter(e)).collect();`
  - [x] Constructable with `SomeFilter::new().and(SomeOtherFilter::new())`, so basic operators are included
    - [x] `::and()`
    - [x] `::or()`
    - [x] `::not()`
  - [ ] Builtin filters
    - [x] Header filters
      - [x] Header field equals X
      - [x] Header field grep
      - [x] Header field is type
      - [x] Header field is set (Bool -> bool, else true)
      - [x] Header field is empty (Null -> true, Array, Table)
    - [ ] Content filter
      - [x] Grep
      - [x] Length
      - [ ] Links in text
- [ ] CLI-Helper to build Filters from a commandline spec (a String, actually)
  - [ ] Language Specification
  - [ ] Tokenizer
  - [ ] Filter builder
  - [ ] Language extensions (let the user of the library implement custom syntax elements)

### DIFF
--- a/libimagentryfilter/src/builtin/content/grep.rs
+++ b/libimagentryfilter/src/builtin/content/grep.rs
@@ -1,0 +1,57 @@
+use std::convert::Into;
+
+use regex::Regex;
+use regex::Error as RError;
+
+use libimagstore::store::Entry;
+
+use builtin::header::field_path::FieldPath;
+use filter::Filter;
+
+pub trait IntoRegex {
+
+    fn into_regex(self) -> Result<Regex, RError>;
+
+}
+
+impl<'a> IntoRegex for &'a str {
+
+    fn into_regex(self) -> Result<Regex, RError> {
+        Regex::new(self)
+    }
+}
+
+impl<'a> IntoRegex for Regex {
+
+    fn into_regex(self) -> Result<Regex, RError> {
+        Ok(self)
+    }
+}
+
+pub struct ContentGrep {
+    regex: Regex,
+}
+
+impl ContentGrep {
+
+    pub fn new<IR>(regex: IR) -> Result<ContentGrep, RError>
+        where IR: IntoRegex
+    {
+        regex.into_regex()
+            .map(|reg| {
+                ContentGrep {
+                    regex: reg,
+                }
+            })
+    }
+
+}
+
+impl Filter for ContentGrep {
+
+    fn filter(&self, e: &Entry) -> bool {
+        self.regex.captures(&e.get_content()[..]).is_some()
+    }
+
+}
+
--- a/libimagentryfilter/src/builtin/header/field_exists.rs
+++ b/libimagentryfilter/src/builtin/header/field_exists.rs
@@ -1,0 +1,31 @@
+use libimagstore::store::Entry;
+
+use builtin::header::field_path::FieldPath;
+use filter::Filter;
+
+use toml::Value;
+
+pub struct FieldExists {
+    header_field_path: FieldPath,
+}
+
+impl FieldExists {
+
+    pub fn new(path: FieldPath) -> FieldExists {
+        FieldExists {
+            header_field_path: path,
+        }
+    }
+
+}
+
+impl Filter for FieldExists {
+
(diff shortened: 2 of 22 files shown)


matthiasbeyer:
I guess for the CLI implementation we should use https://marwes.github.io/combine/combine/index.html

matthiasbeyer:
@TheNeikos @neithernut review appreciated. Shall I continue implementing this?

TheNeikos:
I think that's a good idea!

matthiasbeyer:
Let me make the point `CLI-Helper to build Filters from a commandline spec (a String, actually)` optionally. This would be a really nice thing, though I doupt that I can implement such a thing in reasonable time.

matthiasbeyer:
I'd like to make the "links in text" sub point of the content-filter point optional as well, as this would pull in a markdown parser and I think it is a bit more of an effort to implement this.

@neithernut @TheNeikos I consider this ready for review now.

libimagentryfilter/src/builtin/content/grep.rs
```diff
@@ -0,0 +1,29 @@
...
+use builtin::header::field_path::FieldPath;
+use filter::Filter;
+
+pub struct ContentGrep {
+    regex: Regex,
+}
+
+impl ContentGrep {
+
+    pub fn new(regex: Regex) -> ContentGrep {
```
TheNeikos: How about an `IntoRegex` trait here? It would convert `Into<&str>` into a Regex or just pass through a regex so you could do this:
```rust
ContentGrep::new(".*")
```

libimagentryfilter/src/builtin/header/field_isset.rs
```diff
@@ -0,0 +1,44 @@
...
+
+    pub fn new(path: FieldPath) -> FieldIsSet {
+        FieldIsSet {
+            header_field_path: path,
+        }
+    }
+
+}
+
+impl Filter for FieldIsSet {
```
TheNeikos: How about `FieldExists` since a `Set` is something else?

libimagentryfilter/src/builtin/content/grep.rs
matthiasbeyer: ACK

libimagentryfilter/src/builtin/header/field_isset.rs
matthiasbeyer: ACK

matthiasbeyer: This changes the implementation of the type a bit. We return `true` as soon as walking the path down worked.

libimagentryfilter/src/builtin/header/field_exists.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl Filter for FieldExists {
+
+    fn filter(&self, e: &Entry) -> bool {
+        let header = e.get_header();
+        self.header_field_path
+            .walk(header)
+            .map(|_| true)
+            .unwrap_or(false)
```
matthiasbeyer: `.walt(header).is_some()` would do here.

libimagentryfilter/src/builtin/content/grep.rs
```diff
@@ -0,0 +1,50 @@
...
+
+}
+
+impl<'a> IntoRegex for &'a str {
+
+    fn into_regex(self) -> Result<Regex, RError> {
+        Regex::new(self)
+    }
+}
+
```
TheNeikos: Also implement it for Regex as well.

matthiasbeyer: Done. Diff annotation won't vanish, though.

matthiasbeyer:
I'm ready for a history cleanup & merge!

TheNeikos:
Looking good to me

PR #167, opened 2016-01-31
doc: Add chapter on linking

matthiasbeyer:
Chapter on how linking to {internal, external} content works.

- [x] Linking overall
- [x] Linking internal content
- [X] Linking external content

PR #171, opened 2016-02-03
Libimaglink/init

matthiasbeyer:
Tracking branch for implementing `libimaglink` which was proposed via #167 

This library implements _almost_ everything what would be needed to implement a "bookmarking" module, though it does focus on the _linking_ part and does not allow tagging and such.

The functionality of this library will be useful in other libraries and modules.

---

```
- [ ] `imag::/foo/bar~0.2.0` [requirement](https://github.com/matthiasbeyer/imag/blob/master/doc/src/02000-store.md#store-path-links-secthestorelinks)
```


matthiasbeyer:
@TheNeikos , @neithernut is such a functionality wanted? I think so, right?

libimaglink/src/internal.rs
```diff
@@ -0,0 +1,121 @@
...
+            let links : Vec<Link> = links.into();
+            let links = links.into_iter().map(|link| Value::String(link.into())).collect();
+
+            let old_links = content_section.insert(String::from("links"), Value::Array(links));
+            if old_links.is_none() {
+                return Ok(Links::new(vec![]));
+            }
+            let old_links = old_links.unwrap();
+
+            if match old_links { Value::Array(_) => { false }, _ => { true } } {
```
matthiasbeyer: @TheNeikos as this scheme gets more frequent: Would it be possible to implement a `is_type!()` macro which does basically this?

TheNeikos: I would do it like this:

```rust
if let Value::Array(_) = old_links { false } else { true } 
```

matthiasbeyer: Ah, sure!

Anyways, how about an macro for this?

TheNeikos: That is not that easily possible really.

matthiasbeyer:
Note: Merging this one after #149 will break the build, as we need the `store::EntryHeader::toml()` function here, which is removed in #149 .

matthiasbeyer:
This is on hold until #149 is merged, as this PR would greatly simplify this PR.

matthiasbeyer:
waiting on #194

matthiasbeyer:
As nobody seems to care in #194, I implemented two-way linking for internal links.

I _hope_ that someone reviews this ASAP.

neithernut:
It looks good on the first shallow read and some of the functionality introduced _is_ needed.

libimaglink/src/error.rs
```diff
@@ -0,0 +1,80 @@
...
+            cause: cause,
+        }
+    }
+
+}
+
+impl Display for LinkError {
+
+    fn fmt(&self, fmt: &mut Formatter) -> Result<(), FmtError> {
+        try!(write!(fmt, "[{}]", link_error_type_as_str(&self.kind.clone())));
```
mario-kr: is the clone necessary? argument is passed as reference.

matthiasbeyer: Nope, thanks for spotting this one (the other one was fixed as well)

matthiasbeyer:
I just rebased and cleaned up the history. I would keep the commits which are there now, even if there was a rewrite in between.

libimaglink/src/external.rs
```diff
@@ -0,0 +1,139 @@
...
+
+}
+
+pub trait ExternalLinker {
+
+    /// get the external link from the implementor object
+    fn get_external_link(&self) -> Result<Option<Link>>;
+
+    /// set the external link for the implementor object
+    fn set_external_link(&mut self, l: Link) -> Result<Option<Link>>;
```
neithernut: The semantics are a bit unclear here. After reading the implementation one might guess that the returned link is the one that was replaced, but a bit of documentation would still be nice.

matthiasbeyer:
As far as I can tell, we don't need the `imag://` scheme, as we do not mix external and internal links... so there is no point in defining this. I will alter the appropriate `doc` section.

neithernut:
We can (and should) still define the `imag://` scheme for use in (user-)interfaces and external programs.

matthiasbeyer:
You mean it should be supported but not be used internally?

neithernut:
Yes.

matthiasbeyer:
I agree on that, though I'm not sure how (and where) to implement this at the current state.

neithernut:
Leaving a specification of the scheme is sufficient for now.

PR #173, opened 2016-02-04
Doc/libimagtag

matthiasbeyer:
Closes #172

PR #174, opened 2016-02-04
Libimagtag/init

matthiasbeyer:
Targets the specification from #173


matthiasbeyer:
@TheNeikos may I recieve your suggestions on the CI error?

matthiasbeyer:
Note: We must ensure that a tag only occurs one.

TheNeikos:
How about letting toml do that for us? Make it a table with the key being the tag and the value being the add_date or something.

matthiasbeyer:
So you would attach even metadata to a tag? Hm... I must say I don't like that idea. Just for the sake of simplicity of our codebase introducing some overhead... nah, sorry, I don't like that.

neithernut:
It would also go against the widespread concept of a "tag", IMO.

matthiasbeyer:
@TheNeikos can you investigate why this PR fails?

matthiasbeyer:
This is on hold until #149 is merged, as this PR would greatly simplify this PR.

neithernut:
I don't see anything fishy.

matthiasbeyer:
@neithernut is this "ack for merging"?

neithernut:
It's an "I think it is ok for merging. I didn't invest much time in inspecting it, though."

PR #177, opened 2016-02-04
libimagrt: Log Debug logging to stderr

matthiasbeyer:


neithernut:
Wait... so log messages will be printed to stdout normally? They should _always_ go to stderr.

matthiasbeyer:
Fixed that. All our logging belongs to stderr now. Will merge now, as the review process is much too slow at the moment.

PR #184, opened 2016-02-06
Check whether the StoreId is inside the store

### LINKED ISSUE #179: [security]: Check whether the path of the entry is actually in the current store
If we {create, delete, retrieve} we do not check whether the path is something like

```
create -p /../../../../foo
```

which would be outside of the store.

matthiasbeyer:
Closes #179 

@TheNeikos review pls.


TheNeikos:
ACK

Issue #108, opened 2016-01-17
[store]: VCS-Backend


matthiasbeyer:
We should decide how to model the VCS backend.

Of course, nothing should be seen from the outside of the lib, though we might want to give users the possibility to specify custom commit messages.

We have to think about how to model and implement this. But it is not that important atm.

matthiasbeyer:
WIP in #486

PR #187, opened 2016-02-06
Libimagstore/git layer

matthiasbeyer:
Targets #108 

@TheNeikos How can I now implement CRUD-git actions? I mean... I'm rather sure git actions should not go to the `impl Store{}` but to some internal sub types, am I right?

Also, as @neithernut suggested, we may not want to make the git operations default (though I'd vote for enabling git by default). For now I will implement this without an on/off switch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,7 @@ script:
 addons:
     apt:
         packages:
+            - cmake
             - libcurl4-openssl-dev
             - libdw-dev
             - libelf-dev
@@ -81,6 +82,8 @@ addons:
             - openssl
             - pkgconfig
             - tree
+        sources:
+            - kalakris-cmake
 
 after_success:
     - travis-cargo --only stable doc-upload
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,8 +74,11 @@ addons:
     apt:
         packages:
             - libcurl4-openssl-dev
-            - libelf-dev
             - libdw-dev
+            - libelf-dev
+            - libgit2
+            - libssh2
+            - openssl
             - tree
 
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,6 +79,7 @@ addons:
             - libgit2
             - libssh2
             - openssl
+            - pkgconfig
             - tree
 
 after_success:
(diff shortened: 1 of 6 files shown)


matthiasbeyer:
And of course travis has not all dependencies installed.

matthiasbeyer:
From discussion with @TheNeikos:

We want to be able to switch git on and off via the configuration file (not via the commandline, though).

If the `Store`-Object gets created and afterwards notices that there are uncommited changes we do the following:

* The `Store::new()` function fails with an error
* The user shall update the store with an utility we provide
  * This utility uses a function of the `Store`-object to commit all entry changes
  * This utility provides several switches to 
    * commit the entry-updates on an extra branch
    * (optional but nice-to-have) use a custom commit message

neithernut:
Mh... I don't think that pulling in git as a dependency all the time is a good idea. Can we make this feature (and the dependencies) optional?

matthiasbeyer:
No, I would make it an explicit dependency, actually. One of the basic ideas of the store is to keep historical data. 

Note that we do not pull git as a dependency, but `libgit2`, which should work without git installed, afaik.

TheNeikos:
You would still need it! Which does not come with Rust.

matthiasbeyer:
@TheNeikos are you sure about this? AFAIK implements `libgit2` the git functionality and does not use core-git: https://libgit2.github.com/

This is because `git` itself is GPLv2 and `libgit2` is GPLv2 with linking exception!

TheNeikos:
I did mean `libgit2`, for example it is installed on my Laptop, but it is an extra package.

matthiasbeyer:
Ah, yes. Sure. Anyways, I would make it a hard dependency, the store should be version controlled!

matthiasbeyer:
Discussion appreciated here, as this would be _deeply_ integrated in the store. If we can come up with a "hook"-layer over the store (which we may need anyways, for other things), I would keep this out of the libstore core!
@neithernut , @TheNeikos

matthiasbeyer:
As we are currently discussing (or rather *as I am currently waiting for review on*) a Hook-system in #200 , I will put this on-hold now.

matthiasbeyer:
I guess the hook system will be merged, so I will close this now.

PR #188, opened 2016-02-06
Add error tracing utility

matthiasbeyer:
This is an utility for printing a trace of errors.

Could be rather useful for binary implementations, so we can tell the user what went wrong.

### DIFF
--- a/libimagutil/src/trace.rs
+++ b/libimagutil/src/trace.rs
@@ -1,0 +1,46 @@
+use std::error::Error;
+use std::io::Write;
+use std::io::stderr;
+
+pub fn trace_error(e: &Error) {
+    print_trace_maxdepth(count_error_causes(e), e, ::std::u64::MAX);
+    write!(stderr(), "");
+}
+
+pub fn trace_error_maxdepth(e: &Error, max: u64) {
+    let n = count_error_causes(e);
+    write!(stderr(), "{}/{} Levels of errors will be printed", (if max > n { n } else { max }), n);
+    print_trace_maxdepth(n, e, max);
+    write!(stderr(), "");
+}
+
+pub fn trace_error_dbg(e: &Error) {
+    print_trace_dbg(0, e);
+}
+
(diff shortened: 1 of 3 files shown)


libimagutil/src/trace.rs
```diff
@@ -0,0 +1,41 @@
...
+    print!("Error {:>4} : {}", idx, e.description());
+    if e.cause().is_some() && idx < max {
+        println!(" -- caused by:");
+        print_trace(idx + 1, e.cause().unwrap());
+    }
+    println!("");
+}
+
+fn print_trace_dbg(idx: u64, e: &Error) {
+    print!("Error {:>4} : {}", idx, e.description());
```
matthiasbeyer: `debug!()`

libimagutil/src/trace.rs
```diff
@@ -0,0 +1,41 @@
...
+    if e.cause().is_some() {
+        debug!(" -- caused by:");
+        print_trace(idx + 1, e.cause().unwrap());
+    }
+    println!("");
```
matthiasbeyer: `debug!()`

matthiasbeyer:
@TheNeikos can I get some review here?

TheNeikos:
Can we have a test output in the doc strings?

matthiasbeyer:
What do you mean by that?

TheNeikos:
As in, an example of what this would spit out.

matthiasbeyer:
I will do that and add some other niceties.

TheNeikos:
I am not convinced of the utility of `trace_error_maxdepth` why would you need to know the 3 first errors if the root cause is buried deeper? I would agree that it is more useful if it showed the _first three_ errors.

matthiasbeyer:
Sounds reasonable. Will implement.

neithernut:
Please print to `stderr` rather than `stdout`.

neithernut:
Ack.

PR #190, opened 2016-02-11
Remove codebase: libimagmodule

matthiasbeyer:
Remove this, as we rather have some small libraries than one big library with functionality for modules.


neithernut:
Ack.

PR #191, opened 2016-02-13
Libimagentrymarkup/init

matthiasbeyer:
Basic idea for the `entry-content -> markup processor -> html page` module.

---

This is mainly an experiment, as there are very few markup libraries for rust (actually only the three I already put into the `Cargo.toml` file). I think about adapting as much as possible here, so as soon as there are bindings for {latex, textile, foobarmarkup} we can adapt it here.

This can also be used as playground for multithreading with Rust maybe.


matthiasbeyer:
I will put this on hold as it might be added in another library.

matthiasbeyer:
Closing because this is too old and stalled.

PR #193, opened 2016-02-13
Imag counter/init

matthiasbeyer:
Just some boilerplate by now, waiting for #149, #171, #174 and maybe #191.

- [X] waiting on #149
- [ ] waiting on #171 
- [x] waiting on #174 
- [ ] waiting on #191

### DIFF
--- a/libimagcounter/src/counter.rs
+++ b/libimagcounter/src/counter.rs
@@ -1,0 +1,144 @@
+use std::convert::From;
+use std::convert::Into;
+use std::ops::DerefMut;
+use std::ops::Deref;
+
+use toml::Value;
+
+use std::collections::BTreeMap;
+
+use libimagstore::store::Store;
+use libimagstore::store::FileLockEntry;
+use libimagstore::storeid::StoreId;
+use libimagstore::error::StoreError;
+use libimagstore::store::Entry;
+use libimagstore::storeid::IntoStoreId;
+
+use module_path::ModuleEntryPath;
+use result::Result;
+use error::CounterError as CE;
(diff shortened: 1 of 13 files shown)


libimagcounter/src/counter.rs
```diff
@@ -0,0 +1,65 @@
...
+        &self.name
+    }
+
+    pub fn value(&self) -> u64 {
+        self.value
+    }
+
+}
+
+impl From<Entry> for Counter {
```
matthiasbeyer: I wonder whether this is actually possible. AFAIK, I can only get an `&Entry` from the store, but not `Entry`, am I right, @TheNeikos?

TheNeikos: Yup

matthiasbeyer:
Wait on #207 for more features via `ArgGroup`s from `clap`.

matthiasbeyer:
I don't want to wait any longer for the two PRs to be merged, so can I please get some review on this? Linking support will be added later on for this binary.

neithernut:
Some documentation would be nice.

neithernut:
Eg. some usecase

matthiasbeyer:
@neithernut What would you like to be documented? The binary? The library?

neithernut:
First of all, an (informal) use-case would be nice. More important, however, would be "module documentation" for which you already laid out the structure in `/doc/src`

matthiasbeyer:
ACK, will add this!

And then I will file some issues on how to improve this module.

TheNeikos:
Looks good to me

Issue #199, opened 2016-02-15
[RFC] [store]: hook layer


matthiasbeyer:
I think we should add a Hook layer to the store, which then can be used to execute hooks before and after {CRUD} actions. These Hooks should be able to abort store actions. These Hook actions should even be able to provide dependency-hook actions, so we can seperate Hooks and re-use them in a niceish way.

In the store we would need:

* A way to `register()` hooks with the store
  * for create
  * for read
  * for update
  * for delete
* One error type for wrapping Hook errors

As hook interface:

* A trait which gives us
  * an `explanation()` method which explains the hook to the user
  * an `execute()` for the hook
  * a `dependency_hooks()` which returns a `Vec<Hook>` which must be succeed before `Self`
* Hook error types
* An `enum` for identifying when the hook should be executed(1)

---

(1) I'm not sure whether an enum with `PreRead`, `PostRead` etc would be better or custom traits for each hook

---

With a hook system we could implement:

* git (including switching it on and off via configuration) or other VCS
* Link-consistency checkers
* Markup checkers (whether the user entered correct markup, would reduce noise in the binaries)
* even more great things...

and because some traits should be deeply integrated I would also opt for a standard-set of hooks (which are enabled by default, but could be disabled) which are `registered()` with the store upon creation of the `Store` object.

---

I propose to intergrate this with the store itself, for simplicity of the codebase as well as bundling this feature with the store. This way, the hooks are not "enable if you want" but preconfigured hooks are simple to implement and all this stuff is even backwards compatible with the current store interface!

matthiasbeyer:
If we do this and store the registered hooks in the store itself, we must be able to alter the `Store` object, thus `mut`. The `Store` object is stored in the `Runtime` object, which is immutable.

So we would need a way to alter the `Store` objects internals without making the `Store` itself mutable.

neithernut:
Sounds great. I'd like to sleep over it though, before commenting on it in detail.

matthiasbeyer:
#200 is just a proposal to the upper idea. It is implemented without the `enum` but with traits for each kind of hook, which I find a bit better, but which also introduces a bit overhead in the `Store` object implementation.

neithernut:
Using traits and having the hooks/functions be methods required by the traits would have a number of benefits.

matthiasbeyer:
And a number of drawbacks, as the implementation is rather complicated (see #200).

neithernut:
I see it now.

[...]

PR #200, opened 2016-02-15
Libimagstore/hooks

matthiasbeyer:
Targets #199

- [x] Hook sorting
- [x] Hook parallel execution (only works when non-mutable data is passed)


matthiasbeyer:
I first tried to implement the `register...()` functions via one generic function, though I failed. It is not that complex code, so I just vim-macro-copy-pasted it.

matthiasbeyer:
@TheNeikos I have a delicate problem here. I need to be able to cast a `Hook` to a specific `*Hook`-Trait implementation, as the base trait `Hook` does not have an `execute()` function.

This is the problem in short: https://play.rust-lang.org/?gist=15db7dc75595bd4a2e3d&version=stable

(see below for the lazy)

```rs
pub trait Hook {
  fn description(&self) -> &str;
}

pub trait SpecificHook : Hook {
  fn execute(dep: i32) -> i32;
}

pub trait AnotherHook : Hook {
  fn execute(deba: i32, debb: i64) -> i64;
}
```

The base trait should enforce `execute()`, though the function may have different signatures per implementation ... which makes things considerably hard...

I asked in irc.mozilla.org/rust-beginners as well.

libimagstore/src/store.rs
```diff
@@ -90,6 +99,19 @@ impl StoreEntry {
...
+     */
+
+    pre_read_hooks      : Arc<Mutex<Vec<Box<PreReadHook>>>>,
+    post_read_hooks     : Arc<Mutex<Vec<Box<PostReadHook>>>>,
+    pre_create_hooks    : Arc<Mutex<Vec<Box<PreCreateHook>>>>,
+    post_create_hooks   : Arc<Mutex<Vec<Box<PostCreateHook>>>>,
+    pre_retrieve_hooks  : Arc<Mutex<Vec<Box<PreRetrieveHook>>>>,
+    post_retrieve_hooks : Arc<Mutex<Vec<Box<PostRetrieveHook>>>>,
+    pre_delete_hooks    : Arc<Mutex<Vec<Box<PreDeleteHook>>>>,
+    post_delete_hooks   : Arc<Mutex<Vec<Box<PostDeleteHook>>>>,
```
matthiasbeyer: This does not work, because the Hook traits "cannot be made into an object"... I guess things will get ugly here, right @TheNeikos ?

libimagstore/src/hook.rs
```diff
@@ -0,0 +1,162 @@
...
+    FileLockEntryDep,
+}
+
+pub trait HookDataDependencySet<T> {
+    fn set_dependency_data(&mut self, data: T);
+}
+
+pub trait Hook : Eq + PartialEq + Debug + Display {
+    fn dependencies<H: Hook>(&self) -> Vec<Box<H>>;
+    fn get_data_dependency_ids(&self) -> Vec<DataDependencyId>;
```
matthiasbeyer: The whole idea with `DataDependencyId` and `HookDataDependencySet<T>` is _an idea how this could be implemented_. If someone can come up with a better way to solve this problem,... go ahead!

TheNeikos:
I fail to see why would like to have this depend on traits rather than implementations? Have a trait for each hook and a vec for each that get executed.

matthiasbeyer:
Because one could implement custom hooks in their application and register them with the store for their own usecases!

If you have an implementation, you cannot register own implementations, can you?

matthiasbeyer:
Notes to myself: The Hook system may NEVER be able to access the `Store` object it belongs to. As the Hooks may need configuration from the config file, which content lives in the `Runtime` object (where also the `Store` object lives), we need to be able to pass a sub-set of the configuration to the Hook objects.

matthiasbeyer:
@TheNeikos and @neithernut I'm ready for review here.

What we can implement with this hook system (what pops into my mind right now):

* A `flock()` mechanism
* Version controlled store (git, ...)
* Encrypted store objects (via gnupg for example)
* Link-Integrity checks

I think all of these things should be in a `libimagstdhooks` library and build into the store on `Runtime:new()` by default. One should be able to disable all of them, of course.

One could even build a DebugHook which spits out dumps of data for each store hook which gets execute. And, of course, each module is able to implement more hooks for the store. Possibilities at your hands! :smile:

matthiasbeyer:
One thing I like to implement for this PR is: Running hooks in parallel.

matthiasbeyer:
As I cannot forward `LockResult<T>` Errors (they do not implement `Send`: Would it be a good idea to implement an Error-Dummy (`UnknownError`) which can be passed to the error-creating for "Hey, something weird happened something,... I don't know more". I dislike the "None" because the backtrace we print ends at something which is not the final error. If we would have a "_Something_ is wrong" error type, we can at least see that we don't know the exact error.

@TheNeikos what do you think?

neithernut:
I want to note that `flock()`ing really should go into the store itself, as it should be considered base functionality.

libimagstore/src/hook.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+pub mod update {
+    use store::FileLockEntry;
+    use super::HookResult;
+    use super::Hook;
+
+    pub trait PreUpdateHook : Hook {
+        fn pre_update(&self, &FileLockEntry) -> HookResult<()>;
```
neithernut: There are use-cases where you want to alter the entry on it's way to the disk. Encryption, for example.

neithernut:
Mh... I see lot's of copy-paste-code in `store` (all those functions calling the hooks). Maybe there is a better way. (e.g. using a template-like-technique)

It would also be nice to have a special `HookErrorKind` by which a pre-hook can transparently _deny_ an action. With such a type, you could present the user some useful information like "hook disallowed action" instead of "some error occurred".

matthiasbeyer:
> Mh... I see lot's of copy-paste-code in store (all those functions calling the hooks). Maybe there is a better way. (e.g. using a template-like-technique)

If you can come up with an generic implementation of this... feel free. By now, this is a prove of concept. Improvements could be done later on.

On the second part: Sounds good, I will consider implementing this.

libimagstore/src/hook.rs
matthiasbeyer: I made these Hook trait function arguments non-mutable because I'm not sure whether we want to make it mutable.

The problem is that we need to be able to sort the hooks if we want mutability here. Usecase: Encrypt and compress or compress and encrypt? Even more obvious: Encrypt and checking for consistency afterwards or the other way round? Of course encryption is _always_ the last thing you want to do. Or isn't it? Maybe VCS is the last thing? Maybe there's even something else which might happen _after_ VCS? We don't know!

So the hooks must implement `Ord` and `Eq`. This is feasible, though it might get _really_ complicated.

matthiasbeyer:
@TheNeikos Would like to hear your opinion on 13bf95e. If this is good in your opinion, we could implement that scheme for all error kinds...

TheNeikos:
How would it be used? Why not just the `Into` trait?

matthiasbeyer:
Because of https://github.com/matthiasbeyer/imag/commit/13bf95ea97e0da61ecb8d40cdf01d33e16807d8c#diff-ccfafd1d441cb7416092c52651e32f49R132 - if you need a cause...

TheNeikos:
What about tuples?

```
impl Into<YourError> for Error {

}

impl Into<(YourError, Box<OtherError>)> for Error {

}
```

matthiasbeyer:
@TheNeikos ACK, will implement.

TheNeikos:
Err, yeah of course.

matthiasbeyer:
After #225 (and therefor #232) are merged, this will be ready for review and merge, as far as I can see. 

Of course all the commits will be rebased onto master and squashes whereever possible.

@TheNeikos, @neithernut

matthiasbeyer:
@neithernut and @TheNeikos This is now rebased and squashed where easily possible.

One thing I just noted: `execute_hooks_for_file()` is never called, but always `execute_hooks_for_mut_file()`. This is because the file is actually everywhere mutable where hooks should be called. So the `execute_hooks_for_mut_file()` should take care of mut/non-mut access to the file and we need to alter the calling mechanism a bit for it.

The Aspect gets called `access_mut()` and this function should take care of mut/nonmut access by checking whether the access of the hooks is mutable or nonmutable, dividing into chunks and then execute the non-mutable chunks in parallel and the mutable ones sequencially.

For simplicity, I would change the calling mechanism to be sequentially for the first implementation, adaptions are trivial (as far as multithreading is trivial) and could be done in a seperate PR later on.

libimagstore/src/hook/aspect.rs
```diff
@@ -0,0 +1,129 @@
...
+                    elem.map(|_| a).map_err(|_| HE::new(HEK::HookExecutionError, None))
+                })
+            })
+    }
+}
+
+impl MutableHookDataAccessor for Aspect {
+    fn access_mut(&self, fle: &mut FileLockEntry) -> HookResult<()> {
+        let accessors : Vec<HDA> = self.hooks.iter().map(|h| h.accessor()).collect();
+        if !accessors.iter().all(|a| match a { &HDA::MutableAccess(_)  => true, _ => false }) {
```
matthiasbeyer: We must allow `NonMutableAccess` here as well.

libimagstore/src/hook/aspect.rs
```diff
@@ -0,0 +1,129 @@
...
+            return Err(HE::new(HEK::AccessTypeViolation, None));
+        }
+
+        for accessor in accessors {
+            match accessor {
+                HDA::MutableAccess(accessor) => try!(accessor.access_mut(fle)),
+                _ => unreachable!(),
+            }
+        }
```
matthiasbeyer: ... And execution can be kept sequencially for the first step.

libimagstore/src/hook/aspect.rs
```diff
@@ -0,0 +1,129 @@
...
+        Ok(())
+    }
+}
+
+impl NonMutableHookDataAccessor for Aspect {
+    fn access(&self, fle: &FileLockEntry) -> HookResult<()> {
```
matthiasbeyer: This is actually never called by the store, as we always provide mutable access. If one hook does not want mutable access, it would still be called by the `MutableHookDataAccessor::access_mut()` implementation and therefor this chunk of code is dead.

@TheNeikos , @neithernut how to proceed? We clearly need the distinction between mutable and nonmutable access, though the calling process would be using mutable data.

TheNeikos:
This has merge conflicts :?

matthiasbeyer:
@TheNeikos resolved.

TheNeikos:
Looking good :+1:

PR #203, opened 2016-02-19
Add ismatch!() macro

matthiasbeyer:
@TheNeikos proves you wrong.

This adds a `is_match!()` macro, to condense this:

```rust
if match x { FooBar(_) => { true }, _ => { false } } {
    // ...
}
// or even
if foo.all(|x| match x { Some(FooBar(_)) => { true }, _ => false }) {
   // ...
}
```

down to

```rust
if is_match!(x, FooBar(_))
    // ...
}
// or 
if foo.all(|x| is_match!(x, Some(FooBar(_)))) {
    // ...
}
```

### DIFF
--- a/libimagutil/src/ismatch.rs
+++ b/libimagutil/src/ismatch.rs
@@ -1,0 +1,23 @@
+#[macro_export]
+macro_rules! is_match {
+    ($expression: expr, $($pattern:tt)+) => {
+        is_match! {tt
+            match $expression {
+                $($pattern)+ => true,
+                _            => false
+            }
+        }
+    };
+    (tt $value:expr) => ($value);
+}
+
+#[test]
+fn test_matching() {
+    let foo = Some("-12");
+    assert!(is_match!(foo, Some(bar) if
+        is_match!(bar.as_bytes()[0], b'+' | b'-') &&
+        is_match!(bar.as_bytes()[1], b'0'...b'9')
+    ));
+    assert!(!is_match!(foo, None));
+}
+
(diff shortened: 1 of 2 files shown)


neithernut:
Looks good. You have a use-case in mind?

matthiasbeyer:
@neithernut Some of the PRs which are pending (and should really be reviewed, as they are all pending because nobody reviews them, *hint, hint*) could use this to make the code less verbose. It would simplify exactly what I showed above.

TheNeikos:
please don't export two macros but something like this instead:


```
#[macro_export]
macro_rules! is_match {
    ($expression: expr, $($pattern:tt)+) => {
        is_match! {tt
            match $expression {
                $($pattern)+ => true,
                _            => false
            }
        }
    };
    (tt $value:expr) => ($value);
}

fn main() {
    let foo = Some("-12");
    assert!(is_match!(foo, Some(bar) if
        is_match!(bar.as_bytes()[0], b'+' | b'-') &&
        is_match!(bar.as_bytes()[1], b'0'...b'9')
    ));
}
```

TheNeikos:
I do not see the advantage to just using `if let` to this, since that is exactly what it was made for :?

matthiasbeyer:
The advantage is much less noise in the code as shown above. Especially the `::all()` example is way more readable.

matthiasbeyer:
At least it is just an `util`, so you don't have to use it... :smile:

TheNeikos:
Alright :+1:

PR #205, opened 2016-02-20
rt: Remove short versions for "config" and "runtimepath" std arg

### LINKED ISSUE #204: Remove short arguments from standard-args
`libimagrt` implements some standard arguments on the arg parser object. Only the long versions should be provided for the following arguments:

* config
* runtimepath ("rtp" is fine as long argument")

matthiasbeyer:
Closes #204


TheNeikos:
:+1:

PR #206, opened 2016-02-20
Regex lazy static

### LINKED ISSUE #201: Regex: Ensure we do not compile regexes multiple times
https://github.com/rust-lang-nursery/regex#usage-avoid-compiling-the-same-regex-in-a-loop

matthiasbeyer:
Closes #201


TheNeikos:
Sounds good! Should give some speed up.

Issue #169, opened 2016-02-02
[Feature Request] Store::move()


matthiasbeyer:
We could use a `Store::move()` functionality, which moves an Entry from one place to another, for renaming entries.

matthiasbeyer:
Interface would look like this:

```rs
pub fn move(&self, id: StoreId, new_id: StoreId) -> Result<()> {
    unimplemented!()
}
```

@TheNeikos , @neithernut

neithernut:
Sounds reasonable.

TheNeikos:
Yup

matthiasbeyer:
@TheNeikos wanna implement such a thing?

PR #209, opened 2016-02-21
Add functions for moving and copying store entries

matthiasbeyer:
Targets #169 , though I'm not sure I can implement such a thing.

@TheNeikos should implement this.

### DIFF
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -33,6 +33,7 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     GlobError               => "glob() error",
     EncodingError           => "Encoding error",
     StorePathError          => "Store Path error",
+    EntryRenameError        => "Entry rename error",
 
     CreateCallError            => "Error when calling create()",
     RetrieveCallError          => "Error when calling retrieve()",
@@ -42,7 +43,6 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     UpdateCallError            => "Error when calling update()",
     RetrieveCopyCallError      => "Error when calling retrieve_copy()",
     DeleteCallError            => "Error when calling delete()"
-
 );
 
 generate_custom_error_types!(ParserError, ParserErrorKind, CustomErrorData,
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -42,7 +42,9 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     RetrieveForModuleCallError => "Error when calling retrieve_for_module()",
     UpdateCallError            => "Error when calling update()",
     RetrieveCopyCallError      => "Error when calling retrieve_copy()",
-    DeleteCallError            => "Error when calling delete()"
+    DeleteCallError            => "Error when calling delete()",
+    MoveCallError              => "Error when calling move()",
+    MoveByIdCallError          => "Error when calling move_by_id()"
 );
 
 generate_custom_error_types!(ParserError, ParserErrorKind, CustomErrorData,
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -186,6 +186,8 @@ pub struct Store {
     post_update_aspects   : Arc<Mutex<Vec<Aspect>>>,
     pre_delete_aspects    : Arc<Mutex<Vec<Aspect>>>,
     post_delete_aspects   : Arc<Mutex<Vec<Aspect>>>,
+    pre_move_aspects      : Arc<Mutex<Vec<Aspect>>>,
+    post_move_aspects     : Arc<Mutex<Vec<Aspect>>>,
 
     /**
      * Internal Path->File cache map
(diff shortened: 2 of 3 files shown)


TheNeikos:
Well, a 'true' move would only be the last operation. The other one is more akin to a 'save as'.

matthiasbeyer:
So what do you want me to change?

TheNeikos:
Ya

matthiasbeyer:
> > So what do you want me to change?

> Ya

@TheNeikos What does this mean?

TheNeikos:
Ha! I over read the 'what'. 

I meant to change the name of the first function to something like 'save_as' since you do not really move anything. (From the function signature)

matthiasbeyer:
I just did so, feel free to review again.

Or, if you want to, use this commit to implement the functionalit (must happen soonish, so I can close this).

libimagstore/src/store.rs
```diff
@@ -247,6 +247,40 @@ impl Store {
...
+        let new_id = self.storify_id(new_id);
+        let old_id = self.storify_id(old_id);
+        let hsmap = self.entries.write();
+        if hsmap.is_err() {
+            return Err(StoreError::new(StoreErrorKind::LockPoisoned, None))
+        }
+        let mut hsmap = hsmap.unwrap();
+        if hsmap.contains_key(&id) {
+            // We have to unload the entry
+            unimplemented!()
```
matthiasbeyer: @TheNeikos How to un-load the entry from the hashmap, so we can move it on the FS?

I would also load the entry after the merge, what do you think?

matthiasbeyer: Maybe we can make the reloading optional via a argument flag.

matthiasbeyer: @TheNeikos I would still like to hear your opinion on this one.

matthiasbeyer:
@TheNeikos could you review?

libimagstore/src/error.rs
```diff
@@ -57,6 +59,8 @@ fn store_error_type_as_str(e: &StoreErrorKind) -> &'static str {
             => "The internal Store Lock has been poisoned",
         &StoreErrorKind::EntryAlreadyBorrowed => "Entry is already borrowed",
         &StoreErrorKind::EntryAlreadyExists   => "Entry already exists",
+        &StoreErrorKind::EntryNotExistent => "Entry does not exists",
```
TheNeikos: Never used?

matthiasbeyer: Will see after rebase :smile:

matthiasbeyer: It doesn't warn me because the variant is right there used where you annotated, but yes, it is not used elsewhere.

matthiasbeyer:
@TheNeikos would love another review iteration! :smile:

matthiasbeyer:
@TheNeikos ping.

libimagstore/src/store.rs
```diff
@@ -528,6 +544,80 @@ impl Store {
...
+        self.save_to_other_location(&entry, new_id, true)
+    }
+
+    fn save_to_other_location(&self, entry: &FileLockEntry, new_id: StoreId, remove_old: bool)
+        -> Result<()>
+    {
+        use std::fs::copy;
+        use std::fs::remove_file;
+
+        let new_id = self.storify_id(new_id);
```
matthiasbeyer: Missing pre-move aspect execution here?

libimagstore/src/store.rs
```diff
@@ -528,6 +544,80 @@ impl Store {
...
+
+    /// Move an entry without loading
+    pub fn move_by_id(&self, old_id: StoreId, new_id: StoreId) -> Result<()> {
+        use std::fs::rename;
+
+        let new_id = self.storify_id(new_id);
+        let old_id = self.storify_id(old_id);
+
+        if let Err(e) = self.execute_hooks_for_id(self.pre_move_aspects.clone(), &old_id) {
+            return Err(e);
```
matthiasbeyer: What about aborting vs. non-aborting hooks here? We must cover that! So this must be done in #370 if this is merged beforehand or the other way round.

matthiasbeyer:
Looks good besides the two annotations.

libimagstore/src/store.rs
matthiasbeyer: I left it out on purpose I guess.

Maybe we should add another hook aspect "saveas" or something like that. @TheNeikos opinion welcome here.

libimagstore/src/error.rs
```diff
@@ -32,7 +32,8 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     StorePathLacksVersion   => "The supplied store path has no version part",
     GlobError               => "glob() error",
     EncodingError           => "Encoding error",
-    StorePathError          => "Store Path error"
+    StorePathError          => "Store Path error",
+    EntryRenameError        => "Entry rename error"
```
TheNeikos: Add a comma here to not have a two line diff next time

libimagstore/src/store.rs
```diff
@@ -528,6 +544,80 @@ impl Store {
         self.execute_hooks_for_id(self.post_delete_aspects.clone(), &id)
     }
 
+    /// Save a copy of the Entry in another place
+    /// Executes the post_move_aspects for the new id
+    pub fn save_to(&self, entry: &FileLockEntry, new_id: StoreId) -> Result<()> {
+        self.save_to_other_location(entry, new_id, true)
```
TheNeikos: Don't delete the original

libimagstore/src/store.rs
TheNeikos: What would you put in stuff like that? It sounds to me like a `created` hook rather than a move, after all a move can be seen as removal/creation

PR #211, opened 2016-02-21
doc: Remove file for "libimagmodule", as this was removed

matthiasbeyer:
Remove the chapter on "libimagmodule" from the documentation. "libimagmodule" was removed.


TheNeikos:
:+1:

PR #212, opened 2016-02-23
Libimagrt/color log

matthiasbeyer:
Colored log is awesome log.

Will extend, so the functionality can be switched of via a commandline flag.

### DIFF
--- a/libimagrt/src/logger.rs
+++ b/libimagrt/src/logger.rs
@@ -40,8 +40,22 @@ impl Log for ImagLogger {
 
     fn log(&self, record: &LogRecord) {
         if self.enabled(record.metadata()) {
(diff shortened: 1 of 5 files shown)


libimagrt/src/logger.rs
```diff
@@ -25,8 +40,22 @@ impl Log for ImagLogger {
 
     fn log(&self, record: &LogRecord) {
         if self.enabled(record.metadata()) {
-            // TODO: This is just simple logging. Maybe we can enhance this lateron
-            writeln!(stderr(), "[imag][{: <5}]: {}", record.level(), record.args());
+            let mut term = stderr().unwrap();
```
matthiasbeyer: Same problem as in #213 - This might blow up here. We're waiting until #213 is solved, which would solve this one here as well.

PR #213, opened 2016-02-23
BRIGHT_RED colored error trace output

matthiasbeyer:
Colored error tracing.

Ready for review.

### DIFF
--- a/libimagutil/src/trace.rs
+++ b/libimagutil/src/trace.rs
@@ -1,6 +1,9 @@
 use std::error::Error;
 use std::io::Write;
-use std::io::stderr;
+
+use term::color::*;
+use term::stderr;
+use term::StderrTerminal;
 
 /// Print an Error type and its cause recursively
 ///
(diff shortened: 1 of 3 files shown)


libimagutil/src/trace.rs
```diff
@@ -29,9 +33,11 @@ pub fn trace_error(e: &Error) {
 /// Output is the same as for `trace_error()`, though there are only `max` levels printed.
 pub fn trace_error_maxdepth(e: &Error, max: u64) {
     let n = count_error_causes(e);
-    write!(stderr(), "{}/{} Levels of errors will be printed", (if max > n { n } else { max }), n);
-    print_trace_maxdepth(n, e, max);
-    write!(stderr(), "");
+    let mut term = stderr().unwrap();
```
TheNeikos: When could this panic?

matthiasbeyer: After reading through https://stebalien.github.io/doc/term/src/term/lib.rs.html#106-108 and https://github.com/Stebalien/term/blob/fe676a3abe189de841c174ef53c7d35b8afdddb0/src/terminfo/mod.rs#L303-L305 it seems that this only panics if there cannot be a terminal opened. At least there is no `panic!()` in the crates root afaics. Shall I add a comment on this?

TheNeikos: So if I pipe the output into a file this might panic?

matthiasbeyer: I'm not sure. I will test as soon as I can.

matthiasbeyer: Jop, it panics if `$TERM` is not set. I will implement this in another way.

matthiasbeyer: Waiting on https://github.com/Stebalien/term/issues/57

PR #216, opened 2016-02-23
Debug logging to trace logging

matthiasbeyer:
Moves some `debug!()` log output to `trace!()` level.

My policy on this is: If you print variables in the log output, it might be a good idea to `trace!()` instead of `debug!()`. `debug!()` should be done for _what is done_ and `trace!()` for _what data is used_.

Feel free to argue on the single diffs.

### DIFF
--- a/imag-store/src/create.rs
+++ b/imag-store/src/create.rs
@@ -73,7 +73,7 @@ fn create_from_cli_spec(rt: &Runtime, matches: &ArgMatches, path: &PathBuf) -> R
                 .unwrap_or(String::new())
         });
 
-    debug!("Got content with len = {}", content.len());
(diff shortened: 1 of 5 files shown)


TheNeikos:
I disagree on the semantics. Trace is meant to *trace* what is going on, i.e. single steps in an algorithm.

matthiasbeyer:
@TheNeikos this is what these logging calls do... they print values between computations in algorithms.

TheNeikos:
No, trace is below debug, or at least should be. Debug is for general system information. Trace is for **every** *single* step!

matthiasbeyer:
So you think I should close this unmerged, right?

PR #130, opened 2016-01-21
[WIP] Doc mod

matthiasbeyer:
Add high-level description what each module does.

---

The following modules are planned:

- [ ] bookmarks
- [ ] borrow-tracker
- [ ] calendar (via external tools, E.G.: vdirsyncer)
- [ ] contacts (via external tools, E.G.: vdirsyncer)
- [ ] counter
- [ ] cue cards
- [ ] dairy
- [ ] images (via extenal tools, E.G.: git-annex)
- [ ] ledger (via external tools, E.G.: ledger-cli)
- [ ] mails (via extenal tools, E.G.: mutt, offlineimap, msmtp)
- [ ] movies (via extenal tools, E.G.: git-annex)
- [ ] music (via extenal tools, E.G.: git-annex)
- [ ] news (via external tools, E.G.: newsbeuter)
- [ ] notes
- [ ] password (via external tools, E.G.: password-store)
- [ ] personal library
- [ ] todo (via external tools, E.G.: taskwarrior)
- [ ] weather
- [ ] wiki

and also some utility modules:

- [ ] link helper
- [ ] tag helper
- [ ] git helper
- [ ] content-to-website (markup converter for viewing things in the browser)
- [ ] store-integrity check (find dead links, etc)

PR #219, opened 2016-03-01
Add module files for several modules

matthiasbeyer:
Rename modules, so we have alphabetical order for the modules, which is
way nicer than the current order.

The following modules are added:

    - borrow
    - counter
    - cuecards
    - dairy
    - ledger
    - password
    - weather

Plus, this commit adds an utility module file, where the utility modules
are described

--- 

Extracted from #130


matthiasbeyer:
This PR includes PR #228

TheNeikos:
:+1:

PR #223, opened 2016-03-01
Imag tag/init

matthiasbeyer:
Tagging utility.


imag-tag/src/main.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    if line_out {
+        for tag in &tags {
+            println!("{}", tag);
+        }
+    }
+
+    if sepp_out {
+        let sepp = scmd.value_of("sep").unwrap(); // we checked before
+        println!("{}", tags.join(sepp));
```
matthiasbeyer: Somehow this fails to include the newslines when beeing called with seperator `" and "`.

I will submit this as bug report once this is merged.

PR #224, opened 2016-03-01
Imag view/init

matthiasbeyer:
Viewing utility.

### DIFF
--- a/imag-view/src/main.rs
+++ b/imag-view/src/main.rs
@@ -1,0 +1,3 @@
+fn main() {
+    println!("Hello, world!");
+}
--- a/imag-view/src/main.rs
+++ b/imag-view/src/main.rs
@@ -1,3 +1,7 @@
+extern crate clap;
+
+mod ui;
+
 fn main() {
     println!("Hello, world!");
 }
--- a/imag-view/src/main.rs
+++ b/imag-view/src/main.rs
@@ -66,7 +66,26 @@ fn main() {
         }
         let scmd = scmd.unwrap();
 
-        let viewer = build_viewer(scmd);
+        let viewer = {
+            if scmd.is_present("view-in-stdout") {
+                Box::new(StdoutViewer::new())
+            } else if scmd.is_present("view-in-ui") {
+                warn!("Viewing in UI is currently not supported, switch to stdout");
+                Box::new(StdoutViewer::new())
+            } else if scmd.is_present("view-in-browser") {
+                warn!("Viewing in browser is currently not supported, switch to stdout");
+                Box::new(StdoutViewer::new())
+            } else if scmd.is_present("view-in-texteditor") {
+                warn!("Viewing in texteditor is currently not supported, switch to stdout");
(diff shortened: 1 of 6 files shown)


matthiasbeyer:
This is basically ready, though features are missing. I'd merge it and implement features later on.

imag-view/src/main.rs
```diff
@@ -0,0 +1,169 @@
...
+                  rt: &'a Runtime)
+    -> StoreResult<FileLockEntry<'a>>
+{
+    debug!("Checking path element for version");
+
+    let version = {
+        version.unwrap_or_else(|| {
+            id.split("~").last().unwrap_or_else(|| {
+                warn!("No version");
+                exit(1);
```
TheNeikos: I am not sure about `exit` it basically does not allow destructors to run, is that something we can afford?

TheNeikos:
Except what I annotated :+1:

imag-view/src/main.rs
matthiasbeyer: I will fix this. This will require me to add a `error` module, so we can wrap the things and trace appropriately.

matthiasbeyer:
Will rebase onto lastest master, which will shorten this PR a bit.

matthiasbeyer:
@TheNeikos please reiterate.

imag-view/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    let about = "View entries (readonly)";
+    let ui = build_ui(Runtime::get_default_cli_builder(name, version, about));
+    let rt = {
+        let rt = Runtime::new(ui);
+        if rt.is_ok() {
+            rt.unwrap()
+        } else {
+            println!("Could not set up Runtime");
+            println!("{:?}", rt.err().unwrap());
+            exit(1); // we can affort not-executing destructors here
```
TheNeikos: with a d

imag-view/src/main.rs
```diff
@@ -0,0 +1,185 @@
...
+    debug!("Lets get rollin' ...");
+
+    info!("No implementation yet");
+
+    let entry_id = rt.cli().value_of("id").unwrap(); // enforced by clap
+
+    if rt.cli().is_present("versions") {
+        if let Err(e) = view_versions_of(entry_id, &rt) {
+            trace_error(&e);
+            exit(1);
```
TheNeikos: Same reasoning as above I presume?

imag-view/src/main.rs
matthiasbeyer: I guess I wanted to write "effort" ...

imag-view/src/main.rs
matthiasbeyer: Yes, though here it is a clean exit with status 1. How would you do that? The program would exit after `trace_error()` anyways, but I want to exit with `EXIT_FAILURE` here...

TheNeikos: I ask because you did not add a comment :)

PR #225, opened 2016-03-01
Libimagstore/hooks sorting

matthiasbeyer:
**This is a PR for #200**

This is a proposal for hook sorting. The current state is absolutely not nice and this PR will be rewritten.


matthiasbeyer:
@TheNeikos, @neithernut This is a proposal on how to sort hooks.

@neithernut Said that we might need partial dependencies. I'm not sure what he means, though I'd love to discuss!

neithernut:
Sry for the late reply, I had more to do in the last few days than expected.

----

Well, not all of the hooks can/must be ordered in some sensible way. Consider, for example, some checkers, which merely disallow writing malformed entries to disk. Ordering them explicitly does not bring any advantages.

neithernut:
An idea which popped a day or two ago: maybe we should not order the hooks themselves but "aspects". An aspect would be a simple string and maybe some constraints (e.g. "check", "do not mutate the entry" or "encrypt", "only one hook allowed"). A hook would be flagged as belonging to some aspect by simply supplying the string when registering the hook.
Aspects would be global and conceptual. We can enforce them to be strictly ordered. We could, for example, list the aspects in the order they are processed in the configuration in a simple list of strings. We would need to reverse the order for some operations though. (e.g. configuration defines order for writing, reverse when reading)

matthiasbeyer:
I basically agree and I also like the idea of user-defined "aspects".

I'm not sure, though, on the following point: Shall _all_ the aspects be configured via the config file? If yes, the hooks would need to be configured with aspects as well... something like:

```toml
hook-aspects = [ "misc", "encryption", "version-control"]

[hooks]
[[gnupg]]
aspect = "encryption"
key = "0x123456789"

[[git]]
aspect = "version-control"
```

So the user would sort the hooks?

I really like that idea!

neithernut:
Yep, I think we should require all the aspects to be configured via the configuration file and all the hooks to be assigned an aspect if they should ever be executed.

matthiasbeyer:
I guess we have to discuss how this should work now.

I'd like to do it this way:

1. Each hook is also a `HookDataAccessorProvider` which has a `accessor() -> Box<Accessor>` function.
1. The `Accessor` is an `enum` over `StoreIdAccess`, `FileAccess { MutableFileAccess, NonMutableFileAccess }`. Each of these has the `access()` method with appropriate signature
1. An `Aspect` takes hooks. The `Aspect` type contains logic to execute the hooks either in parallel or not .

I'm not sure how to model the `Store->Aspect` data passing. Maybe also via `HookDataAccessorProvider` which then passes the data to the underlying hooks?

matthiasbeyer:
@TheNeikos please review this branch, whether it goes into the right direction (and make suggestions). I personally think, that this stuff should be completely reimplemented and the current state is ... aweful :cry: !

TheNeikos:
I've been thinking about it, and without making some hard choices you cannot make this 100% modular and self-working.

matthiasbeyer:
@TheNeikos please elaborate.

TheNeikos:
Nevermind, misread the previous posts. You already made the choice! A config file

libimagstore/src/hook/aspect.rs
```diff
@@ -0,0 +1,65 @@
+use store::FileLockEntry;
+use storeid::StoreId;
+use hook::{Hook, HookResult};
+use hook::{StoreIdDataAccessor,
+    MutableHookDataAccessor,
+    NonMutableHookDataAccessor,
+    HookDataAccessor};
+
+struct Aspect<H: Hook> {
```
TheNeikos: How do you plan on using this?

matthiasbeyer: The idea is that an `Aspect` contains the hooks and the aspect also contains the logic to execute the hooks. I'm working on a re-write of this PR, as I think this PR has become a major mess, but the implementation will look the same.

matthiasbeyer:
@TheNeikos the new commits are pushed now.

The idea is:

1. `Hook`s get registered with their `Aspect`s, an `Aspect` can only contain one `Accessor` type of `Hook` (so only ID, mutable File or nonmutable File) - This is not representable via the typesystem afaik.
1. `Store` calls each `Aspect` from `pre_read_aspects: Vec<Aspect>` (for example)
1. ... and gets an `Accessor` for each `Aspect` which he calls with the appropriate 
1. ... and calls that `Accessor`
1. That `Accessor` (`Accessor` is a trait which is implemented on the `Aspect`) calls each `Hook` inside the `Aspect` - Either in parallel or sequentially

About to be written:

* Aspect registration
* Registering of hooks with aspects

Both of these points need configuration-processing code. I'm a bit lazy at this point, though.

libimagstore/src/hook/aspect.rs
TheNeikos: I mean the actual, code use

TheNeikos: I ask, because if you put in Hook, you can't use anything else, even if it of another type, and you can't upgrade.

http://is.gd/HVNMTE

matthiasbeyer: What do you propose?

matthiasbeyer: Ah, I guess this is not relevant anymore, as all `Hook`s are `HookDataAccessorProvider`s, so an `Aspect` has a `Vec<Box<HookDataAccessorProvider>>` and isn't generic anymore! :smile:

matthiasbeyer:
@TheNeikos review please? This is implemented without parallelization by now, will add parallel exec now.

matthiasbeyer:
With #232, this will be ready for review and merge into #200, as far as I can see.

@TheNeikos, @neithernut

TheNeikos:
I saw nothing wrong with it, I can see the appeal and hope it is one of the very *few* major parts of such a program.

Good job however :dancer: 

:+1:

PR #227, opened 2016-03-02
Libimagstore/clean warnings

matthiasbeyer:
Remove some warnings in the store library (almost all).

Remaining:

- [ ] Non-used methods
- [ ] unused variable in `retrieve_for_module()` (which will be removed)
- [ ] unused result for
  - [ ] `file.write_all()` call
  - [ ] `file.seek()' call

I guess these can remain for now.


neithernut:
Good.

PR #228, opened 2016-03-02
doc: alphabetical ordering for modules

matthiasbeyer:
Moves to a simpler setup for the module documentation, by ordering them alphabetically. Makes more sense IMHO.


neithernut:
90s init systems called. They want their initialization ordering scheme back.

matthiasbeyer:
What do you argue about? It is documentation and it is perfectly fine that the modules are documented in alphabetical order... there's no other sorting scheme which would be "more sane" at this point.

neithernut:
I mistook this with something different. ACK.

PR #229, opened 2016-03-04
doc: Add hook aspect definition

matthiasbeyer:
**This is a PR for #200 ***

---

This is just the documentation proposal. Implementation will be done in #225, to keep discussion + definition and implementation seperated.

@neithernut, @TheNeikos what do you think?


neithernut:
Ack.

PR #231, opened 2016-03-05
Switch configuration language to TOML

matthiasbeyer:
Required for #225, as with TOML everything gets simpler everywhere and especially in the store library.

This also solves some dependency issues with the `config` crate (which depends on `nom ~1.0.0` behind the scenes and this leads to incompatible versions in some places).

@TheNeikos and @neithernut please review ASAP!


neithernut:
Ack.

matthiasbeyer:
Rewrote latest commit msg. Will merge as soon as CI succeeded.

PR #233, opened 2016-03-06
EntryHeader: Function to get toml object

matthiasbeyer:
Add function to get the inner toml object from an `EntryHeader`.

Ready for review here.


neithernut:
Ack.

PR #242, opened 2016-03-10
Revert "Check whether the StoreId is inside the store, before doing a…

### LINKED ISSUE #241: Store::id_in_store() fails to do its job
We pass store ids like this:

```
/test/foo
```

Which should translate to `$STORE_PATH/test/foo`, but `Store::id_in_store()` causes this to fail because it expects the store-id to be the full path of the store entry.

matthiasbeyer:
…nything on the FS"

This reverts commit 373502217e62ad07f76af4579680ad7156bbe390
as this commit was introducing a bug.

The StoreId type says `/test/example` for a store id path, which is
completely valid, as the root (`/`) is the store itself. The
id_in_store() function assumed that the store-id is the full
(file-system) path to the store entry, which is false.

This commit does not introduce a fix but revert the check.

---

Quickfix for and
Closes #241

### DIFF
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -19,7 +19,6 @@ pub enum StoreErrorKind {
     IoError,
     StorePathExists,
     StorePathCreate,
-    StorePathOutsideStore,
     LockPoisoned,
     EntryAlreadyBorrowed,
     EntryAlreadyExists,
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@TheNeikos Carefully review this, as I'm changing your macro here! :+1:

TheNeikos:
:+1: Don't see anything wrong

PR #244, opened 2016-03-10
Fix: implicitely create directories on file saving

### LINKED ISSUE #243: Cannot catch errors with Store::create()
When I create an entry with `Store::create()`, set some data and then let the `FileLockEntry` `Drop`, errors are ignored and I cannot catch errors when writing the `FileLockEntry` to disk.

@TheNeikos Your department.

matthiasbeyer:
Closes #243


TheNeikos:
:+1:

PR #246, opened 2016-03-12
Libimagentryfilter/more filters

matthiasbeyer:
This PR adds some filter functionality by implementing a by-predicate-filter. This filter type was then used to reimplement some other filters (so complexity was moved from N filter implementations to _one_ filter implementation)..

With help from @asuivelentine

### DIFF
--- a/libimagentryfilter/src/builtin/header/field_isempty.rs
+++ b/libimagentryfilter/src/builtin/header/field_isempty.rs
@@ -30,7 +30,7 @@ impl Filter for FieldIsEmpty {
                     Some(Value::Boolean(_)) => false,
                     Some(Value::Float(_))   => false,
                     Some(Value::Integer(_)) => false,
-                    Some(Value::String(_))  => false,
+                    Some(Value::String(s))  => s.is_empty(),
                     Some(Value::Table(t))   => t.is_empty(),
                     _                       => true,
                 }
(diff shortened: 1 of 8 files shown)


libimagentryfilter/src/builtin/header/field_isempty.rs
```diff
@@ -27,10 +27,10 @@ impl Filter for FieldIsEmpty {
...
                 match v {
                     Some(Value::Array(a))   => a.is_empty(),
-                    Some(Value::Boolean(_)) => false,
-                    Some(Value::Float(_))   => false,
-                    Some(Value::Integer(_)) => false,
-                    Some(Value::String(_))  => false,
+                    Some(Value::Boolean(b)) => b,
+                    Some(Value::Float(f))   => f == 0.0,
+                    Some(Value::Integer(i)) => i == 0,
+                    Some(Value::String(s))  => s.is_empty(),
```
neithernut: A field being "empty" and a field being "NULL" are two completely different things.

neithernut:
Still not fit enough to do a proper review. I found something striking, though.

matthiasbeyer:
Yes, how we define emptyness is not yet covered here, though this is a detail I'm ready to talk about. The major changes are introduced, so let's discuss details!

neithernut:
Consider, for example, a field "rating". If you look for unrated material, you clearly don't want to catch everything that's rated "0".
The original code was correct as fields were only considered empty if the data was either an empty array or the field was not set. If a field exists and it's set to a concrete value, the field _is not empty_.
For strings, one might indeed make an exception, just as for arrays, as an field which has an empty string for its value might be considered empty itself.

matthiasbeyer:
I agree with you in all points and will provide the appropriate patch asap.

PR #248, opened 2016-03-14
Add implementation for Store::retrieve_for_module()

### LINKED ISSUE #168: Redefine how a module gets all its entries.
This is from a face-to-face discussion with @TheNeikos .

---

So the basic problem is, that a module may want to get _all_ entries which are available in the store, which are written by the module itself. We have `libimagstore::store::Store::retrieve_for_module()` for this, though this does not fit our needs anymore. We need to redefine this in a sane way.

---

Usecase: `imag bookmark list --all`

matthiasbeyer:
Closes #168

### DIFF
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -33,6 +33,7 @@ pub enum StoreErrorKind {
     PreHookExecuteError,
     PostHookExecuteError,
     StorePathLacksVersion,
+    GlobError,
         // maybe more
 }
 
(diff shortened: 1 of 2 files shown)


neithernut:
I'm not sure whether returning a void iterator in an error case would be also acceptable.

matthiasbeyer:
Nope. Exactly the same as with "Any value vs. empty value". An Empty iterator suggests that there are no entries. An error is something else.

neithernut:
Yep, hence the "I'm not sure"-part. ACK.

neithernut:
It's not exactly the same situation. I totally agree that errors should be reported to the caller, _if_ they are really errors. However, returning a void iterator instead of an error should be considered valid behaviour in some situations (most notably a non-existing directory).
Consider, for example, a new store which simply does not yet contain entries for a specific module. The module-directory inside the store does not yet exist at this point. I'll assume that the globbing will fail with an error at this point, without actually checking. This is, however, counter-intuitive.

matthiasbeyer:
@neithernut I agree. So _some_ errors should be turned into empty iterators. Will do.

matthiasbeyer:
@neithernut I checked the docs for `glob`. The error is only returned in case of, well, `PatternError`. There is no other error case and the `::cause()` does return `None` (as it is the default implementation).

I added another error for type conversion failures, though.

I consider this ready now.

PR #256, opened 2016-03-18
Libimagentryfilter/version filter

matthiasbeyer:
Filter for filtering entries by version.


matthiasbeyer:
Reimplement with `Ord`

matthiasbeyer:
@TheNeikos Actually it is not possible to implement this with `Ord`, as the `Filter` trait wants `filter() -> bool`.

PR #272, opened 2016-03-25
Enhance error printing on config file parsing errors

matthiasbeyer:


matthiasbeyer:
These `println!()` should be `write!(stderr(), ...`s

PR #273, opened 2016-03-25
Enhance store config errors by printing (println)

matthiasbeyer:


neithernut:
You also did some reformatting here.

libimagstore/src/configuration.rs
```diff
@@ -49,15 +49,23 @@ pub fn config_is_valid(config: &Option<Value>) -> bool {
 
     fn has_key_with_string_ary(v: &BTreeMap<String, Value>, key: &str) -> bool {
         v.get(key)
-            .map(|t| match t {
-                &Value::Array(ref a) => a.iter().all(|elem| {
-                    match elem {
-                        &Value::String(_) => true,
-                        _ => false,
+            .map(|t| {
```
matthiasbeyer: @neithernut I assume you mean this new scope here with "reformatting"?

neithernut: No, you added a bunch of empty lines in another place.

matthiasbeyer:
Removed the newlines you mentioned.

libimagstore/src/configuration.rs
neithernut: Now that you said it: why did you add the extra scope here?

matthiasbeyer: I did some `println!()` foo to debug. Will remove.

matthiasbeyer:
All of these `println!()` should be `write!(stderr(), ...`s.

PR #275, opened 2016-03-25
Fix Config-file searching filter bug

matthiasbeyer:


neithernut:
Wouldn't you want to report an error when encountering a directory in the place you'd expect a configuration file?

matthiasbeyer:
Nope, not really. I found this bug because I have a directory `$HOME/config` which is completely okay and also should be okay for imag. Reporting a "config file not found" is acceptable in this case, IMHO.

neithernut:
Wait... `imag` really tries to read from `$HOME/config` (instead of files _within_ that directory)? That's insane!

matthiasbeyer:
That's what this bug fixes.
The valid pathes `~/.imag/config`, `$RUNTIMEPATH/config` are covered.

One could argue to remove `"config"` and `"config.toml"` from https://github.com/matthiasbeyer/imag/blob/master/libimagrt/src/configuration.rs#L213, but I'd do this in another PR.

neithernut:
Well, I'll ACK this PR but the lookup algorithm does need some tweaking. Maybe I'll come up with a PR (soonish).

PR #278, opened 2016-03-25
Make tags unique before writing

### LINKED ISSUE #252: libimagtag does not unique() tags
Adding the same tag twice is possible with libimagtag.

matthiasbeyer:
Closes #252


neithernut:
Ack.

PR #279, opened 2016-03-25
Truncate file to zero size before writing new content, so we do not p…

### LINKED ISSUE #251: Store does not rewrite complete entries
How to reproduce:

1. Create a store entry with some content
1. Update the store entry with _less_ content as before

-> The store entry file is not cleared before re-writing it.

@TheNeikos your department!

matthiasbeyer:
…artially write the file

Closes #251 

@TheNeikos Your department, so please review.


neithernut:
Note: the file is also truncated if the file was just created.

matthiasbeyer:
Yes, but that's okay, I guess.

PR #292, opened 2016-03-25
Move logger initialization to earliest possible point

### LINKED ISSUE #280: Move logger creation to earliest possible point
We should move the initialization of the logger object to the earliest possible point.

matthiasbeyer:
Closes #280 

The code which was run _before_ the logger init was getting too complex. We need logging there. Therefor I'm changing this to happen implicitely on runtime setup.

### DIFF
--- a/imag-counter/src/main.rs
+++ b/imag-counter/src/main.rs
@@ -49,8 +49,6 @@ fn main() {
         }
     };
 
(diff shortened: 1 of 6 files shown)


matthiasbeyer:
I did not test-build this one locally. Lets just see what the CI tells us...

Issue #20, opened 2015-12-05
[module]: Diary


matthiasbeyer:
Combination of Calendar and Notes/Journal module

matthiasbeyer:
I'm already working on this, though I'm not basing it on a calendar module. I'm using the notes module, though.

What this will support for the first iteration:

* multiple diaries (that's the hardest part, actually)
* hourly, daily (default), weekly and monthly entries
* `grep`ing though diaries

PR #309, opened 2016-03-30
Libimagdiary/init

matthiasbeyer:
Library for the diary module.

Work in progress.

### DIFF
--- a/libimagdiary/Cargo.toml
+++ b/libimagdiary/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "libimagdiary"
+version = "0.1.0"
+authors = ["Matthias Beyer <mail@beyermatthias.de>"]
+
+[dependencies]
+log = "0.3.5"
+semver = "0.2"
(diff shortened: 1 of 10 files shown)


matthiasbeyer:
@TheNeikos May I get some comments on the iterator foo I'm doing here?

matthiasbeyer:
It seems that one iterator is _way_ better.

matthiasbeyer:
I guess it was a bad idea to base the diary implementation on the notes implementation.

I will rewrite this completely, I think.

matthiasbeyer:
This PR is now a rewrite.

matthiasbeyer:
Closes #20 btw

matthiasbeyer:
Still missing: The diary Id should always be two digits for month and day, so we do not create `.../2016/1/1/0:0~0.1.0` but `.../2016/01/01/0:0~0.1.0`.

matthiasbeyer:
Not perfect or bug-free, but IMHO in a mergable state.

PR #310, opened 2016-03-30
Libimagentrylist/cli

matthiasbeyer:
CLI interface helpers for `libimagentrylist`


matthiasbeyer:
I don't like how we get the lister instances here. We really should have something like a `EntryListerBuilder` or something like this.

matthiasbeyer:
I'm not completely happy with this PR, though it is a step into the right direction IMHO. Will merge and open issues for the remaining fixes.

Issue #300, opened 2016-03-28
Rewrite libimaglink - external linking


matthiasbeyer:
We should rewrite the `libimaglink` external linking module to

- [ ] Generate Entries in the store under `/external-link` where each entry links to one URL
- [ ] Allow linking operations on `Entry` objects:
  - [ ] linking an external link by putting the actual link into `/external-link/<something>` and linking this new entry with an internal link
  - [ ] listing external links for one entry
  - [ ] listing Entries which are linked from one external link (as internal linking is two-way-linking)

This would contain API changes of the `external` module in `libimaglink` and of course `imag-link` should be adapted.

PR #325, opened 2016-04-08
Libimaglink/internal linking use storeid

matthiasbeyer:
Use `StoreId` internally for linking library, not String.

Makes implementation more complex, but use _way_ simpler.

This is a prequel for the PR for #300


matthiasbeyer:
I guess `imag-link` will not build anymore now.

matthiasbeyer:
Actually I did a good job when writing `imag-link`, so it builds fine with these patches applied. Wonderful!

PR #339, opened 2016-04-14
Remove typedef for StoreId type

matthiasbeyer:
but make it a new type instead.

Also derive and implement as many traits as possible for it, so we keep
backwards compatibility as much as possible.

Anyways, we need to rewrite some code for this.

Suggested-by: Marcel Müller <neikos@neikos.email>
Signed-off-by: Matthias Beyer <mail@beyermatthias.de>

---

@TheNeikos can you come up with code which keeps backwards compatibility to 100%? I don't think so, feel free to surprise me!

---

For the curious: This refactoring would enable us to implement traits on the type elsewhere which would allow us to greatly improve the overall codebase.

It also gives us the possibility to implement functions on the type which then can be used to check certain invariants, for example that the `StoreId` should always be an absolute path but starting from the store root directory rather than from the OS root directory. This would greatly improve and simplify some code I'm currently working on in some other PRs.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -23,7 +23,7 @@ use glob::glob;
 
 use error::{ParserErrorKind, ParserError};
 use error::{StoreError, StoreErrorKind};
-use storeid::{StoreId, StoreIdIterator};
+use storeid::{IntoStoreId, StoreId, StoreIdIterator};
 use lazyfile::LazyFile;
 
 use hook::aspect::Aspect;
@@ -245,8 +245,8 @@ impl Store {
     }
 
     /// Creates the Entry at the given location (inside the entry)
-    pub fn create<'a>(&'a self, id: StoreId) -> Result<FileLockEntry<'a>> {
-        let id = self.storify_id(id);
+    pub fn create<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
+        let id = self.storify_id(id.into_storeid());
         if let Err(e) = self.execute_hooks_for_id(self.pre_create_aspects.clone(), &id) {
             return Err(e);
         }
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@TheNeikos ping.

libimagstore/src/store.rs
```diff
@@ -245,8 +245,8 @@ impl Store {
     }
 
     /// Creates the Entry at the given location (inside the entry)
-    pub fn create<'a>(&'a self, id: StoreId) -> Result<FileLockEntry<'a>> {
-        let id = self.storify_id(id);
+    pub fn create<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
```
matthiasbeyer: @TheNeikos any objections against this change?

libimagstore/src/store.rs
```diff
@@ -273,8 +273,8 @@ impl Store {
 
     /// Borrow a given Entry. When the `FileLockEntry` is either `update`d or
     /// dropped, the new Entry is written to disk
-    pub fn retrieve<'a>(&'a self, id: StoreId) -> Result<FileLockEntry<'a>> {
-        let id = self.storify_id(id);
+    pub fn retrieve<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
```
matthiasbeyer: @TheNeikos any objections against this change?

libimagstore/src/store.rs
TheNeikos: Sounds good

matthiasbeyer:
@TheNeikos :open_mouth: it builds. I only fixed a few places outside of `libimagstore` ... :open_mouth:

matthiasbeyer:
@TheNeikos It should also build with the latest two reverts :open_mouth: I guess I rock a bit here... only one patch outside of `libimagstore` to get this working... looks awesome, doesn't it?

And exactly because of this I hope you do a _really_ careful review! :smile:

matthiasbeyer:
@TheNeikos please review. Also: Can I leave the `Suggested-by:` Tag in the commit message of the first commit?

TheNeikos:
@matthiasbeyer It looks good to me

PR #349, opened 2016-04-16
Libimagstore/walk

matthiasbeyer:
@TheNeikos could you please review this?

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -20,6 +20,8 @@ use std::fmt::Error as FMTError;
 use toml::{Table, Value};
 use regex::Regex;
 use glob::glob;
+use walkdir::WalkDir;
+use walkdir::Iter as WalkDirIter;
 
 use error::{ParserErrorKind, ParserError};
 use error::{StoreError, StoreErrorKind};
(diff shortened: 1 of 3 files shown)


libimagstore/src/store.rs
```diff
@@ -52,6 +54,60 @@ struct StoreEntry {
...
+        self.dirwalker
+            .next()
+            .and_then(|res| {
+                let res = res.map(|next| {
+                    if next.file_type().is_dir() {
+                        Some(StoreObject::Collection(next.path().to_path_buf()))
+                    } else if next.file_type().is_file() {
+                        Some(StoreObject::Id(next.path().to_path_buf()))
+                    } else {
+                        None // TODO: FIXME: Shall we really stop iterating in this case?
```
TheNeikos: Put it in a  loop :)

matthiasbeyer: What do you mean exactly? Do you mean we get the `next()` item in this case and ignore this one?

TheNeikos: Yup? Or am I misunderstanding something?

matthiasbeyer: Nope, could be a solution though I'm not sure I'm happy with it.

matthiasbeyer:
@TheNeikos Could you reiterate for another review? I fixed what we had in the diff discussion and changed the code to loop as we discussed.

PR #366, opened 2016-04-17
Update dependency/regex

### LINKED ISSUE #341: Update crate: Regex
All crates need to have the "regex" crate dependency updated.

We should go for a wildcard here: `regex = "0.1.*"`.

matthiasbeyer:
Closes #341


TheNeikos:
Looking good :+1:

PR #367, opened 2016-04-17
Unwrap err

### LINKED ISSUE #333: Replace .err().unwrap() with unwrap_err() calls
This is a minor issue we have with the codebase: Sometimes we call `foo.err().unwrap()` - it would be better to call `foo.unwrap_err()` instead.

Applies to all crates.

Really simple beginner task.

matthiasbeyer:
Closes #333

### DIFF
--- a/imag-counter/src/list.rs
+++ b/imag-counter/src/list.rs
@@ -15,11 +15,11 @@ pub fn list(rt: &Runtime) {
                         let value   = c.value();
 
                         if name.is_err() {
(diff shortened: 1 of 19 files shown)


TheNeikos:
Doesn't build :C

matthiasbeyer:
It didn't build because `unwrap_err()` couldn't be called because some trait bounds were not fulfilled. Calling `err().unwrap()` works, though.

PR #370, opened 2016-04-18
Libimagstore/hook errors not aborting

### LINKED ISSUE #368: Hook system: Hooks should be able to return errors without aborting the store operation
For example if my Hook A fails, the fail isn't critical but should nevertheless be reported to the user, the current way of doing this is by calling `warn!()` or `info!()` within the hook.

We should refactor this, so a Hook can return an `Err()` without causing the store operation to abort.

I guess we should refactor the error type of the hook operation to include an `abort` field which can be set to a boolean value and causes the operation to be aborted if it is set to true, otherwise the error is simply printed to `stderr`.

This is a complex change which requires a big refactoring in the ´libimagstore` and `libimagstdhook`. Maybe we can get this working without rewriting parts of `libimagstdhook`, but I'm not sure about this.

matthiasbeyer:
Closes #368

### DIFF
--- a/libimagstore/src/hook/error.rs
+++ b/libimagstore/src/hook/error.rs
@@ -1,10 +1,23 @@
-generate_error_module!(
-    generate_error_types!(HookError, HookErrorKind,
-        HookExecutionError  => "Hook exec error",
-        AccessTypeViolation => "Hook access type violation"
-    );
+use std::convert::Into;
+generate_error_imports!();
+
+generate_custom_error_types!(HookError, HookErrorKind, CustomData,
+    HookExecutionError  => "Hook exec error",
+    AccessTypeViolation => "Hook access type violation"
 );
 
-pub use self::error::HookError;
-pub use self::error::HookErrorKind;
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Copy)]
+pub struct CustomData {
+    aborting: bool,
+}
 
+impl HookError {
+
+    pub fn is_aborting(&self) -> bool {
+        match self.custom_data {
+            Some(b) => b.aborting,
+            None => true
+        }
+    }
+
+}
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -28,10 +28,13 @@ use storeid::{IntoStoreId, StoreId, StoreIdIterator};
 use lazyfile::LazyFile;
 
 use hook::aspect::Aspect;
+use hook::error::HookErrorKind;
+use hook::result::HookResult;
 use hook::accessor::{ MutableHookDataAccessor,
             StoreIdAccessor};
 use hook::position::HookPosition;
 use hook::Hook;
+use libimagerror::trace::trace_error;
 
 use libimagerror::into::IntoError;
 
@@ -301,7 +304,13 @@ impl Store {
     pub fn create<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
         let id = self.storify_id(id.into_storeid());
         if let Err(e) = self.execute_hooks_for_id(self.pre_create_aspects.clone(), &id) {
-            return Err(e);
+            if e.is_aborting() {
+                return Err(e)
+                    .map_err(Box::new)
+                    .map_err(|e| SEK::PreHookExecuteError.into_error_with_cause(e))
+            } else {
+                trace_error(&e);
+            }
         }
 
         let mut hsmap = match self.entries.write() {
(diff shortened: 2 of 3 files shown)


matthiasbeyer:
Backwards compatible. Yay! :smile:

matthiasbeyer:
@TheNeikos would be nice if you could have a look whether I've overlooked something.

libimagstore/src/hook/error.rs
```diff
@@ -80,6 +82,14 @@ impl HookError {
         self.err_type.clone()
     }
 
+    pub fn aborting(&mut self, b: bool) {
```
TheNeikos: Uhm, why should I be able to call `aborting(false)` ?

matthiasbeyer: To set the error to be non-aborting. The default is that a `HookError` aborts the store action (create, update, whatever). If you set the error to non-aborting, the error gets `trace()`ed, but doesn't about the store action.

matthiasbeyer:
This is _not_ ready as the `Aspect` code has to be adapted to go on calling hooks even if there was an error (as long as the error is non-aborting).

matthiasbeyer:
Waiting on #414

matthiasbeyer:
Okay, we are ready here. @TheNeikos review pls!

libimagstore/src/store.rs
```diff
@@ -299,7 +302,13 @@ impl Store {
     pub fn create<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
         let id = self.storify_id(id.into_storeid());
         if let Err(e) = self.execute_hooks_for_id(self.pre_create_aspects.clone(), &id) {
-            return Err(e);
+            if e.is_aborting() {
+                return Err(e).map_err(|e| {
+                    SE::new(SEK::PreHookExecuteError, Some(Box::new(e)))
```
matthiasbeyer: We can shorten this with the new error setup: 

```
.map_err(Box::new)
.map_err(|e| SEK::ErrorKind.into_error_with_cause(e))
```

libimagstore/src/store.rs
```diff
@@ -330,7 +339,13 @@ impl Store {
     pub fn retrieve<'a, S: IntoStoreId>(&'a self, id: S) -> Result<FileLockEntry<'a>> {
         let id = self.storify_id(id.into_storeid());
         if let Err(e) = self.execute_hooks_for_id(self.pre_retrieve_aspects.clone(), &id) {
-            return Err(e);
+            if e.is_aborting() {
+                return Err(e).map_err(|e| {
+                    SE::new(SEK::PreHookExecuteError, Some(Box::new(e)))
```
matthiasbeyer: Shorten this, see above.

libimagstore/src/store.rs
```diff
@@ -357,7 +372,13 @@ impl Store {
     {
         let id = self.storify_id(id.into_storeid());
         if let Err(e) = self.execute_hooks_for_id(self.pre_retrieve_aspects.clone(), &id) {
-            return Err(e);
+            if e.is_aborting() {
+                return Err(e).map_err(|e| {
+                    SE::new(SEK::PreHookExecuteError, Some(Box::new(e)))
```
matthiasbeyer: Shorten this, see above.

libimagstore/src/store.rs
```diff
@@ -450,14 +471,21 @@ impl Store {
     /// Return the `FileLockEntry` and write to disk
     pub fn update<'a>(&'a self, mut entry: FileLockEntry<'a>) -> Result<()> {
         if let Err(e) = self.execute_hooks_for_mut_file(self.pre_update_aspects.clone(), &mut entry) {
-            return Err(e);
+            if e.is_aborting() {
+                return Err(e).map_err(|e| {
+                    SE::new(SEK::PreHookExecuteError, Some(Box::new(e)))
```
matthiasbeyer: Shorten this, see above.

libimagstore/src/store.rs
```diff
@@ -590,29 +625,29 @@ impl Store {
...
     {
         match aspects.lock() {
-            Err(_) => return Err(SE::new(SEK::PreHookExecuteError, None)),
+            Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
             Ok(g) => g
         }.iter().fold(Ok(()), |acc, aspect| {
             debug!("[Aspect][exec]: {:?}", aspect);
             acc.and_then(|_| (aspect as &StoreIdAccessor).access(id))
-        }).map_err(|e| SE::new(SEK::PreHookExecuteError, Some(Box::new(e))))
+        }).map_err(|e| HookError::new(HookErrorKind::HookExecutionError, Some(Box::new(e))))
```
matthiasbeyer: Shorten this, see above.

libimagstore/src/store.rs
```diff
@@ -590,29 +625,29 @@ impl Store {
...
     {
         match aspects.lock() {
-            Err(_) => return Err(SE::new(SEK::PreHookExecuteError, None)),
+            Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
             Ok(g) => g
         }.iter().fold(Ok(()), |acc, aspect| {
             debug!("[Aspect][exec]: {:?}", aspect);
             acc.and_then(|_| aspect.access_mut(fle))
-        }).map_err(|e| SE::new(SEK::PreHookExecuteError, Some(Box::new(e))))
+        }).map_err(|e| HookError::new(HookErrorKind::HookExecutionError, Some(Box::new(e))))
```
matthiasbeyer: Shorten this, see above.

TheNeikos:
What about a macro that works similarly to `try!`? It could return on abort, and otherwise return a `Result<T>`

PR #371, opened 2016-04-18
Libimagentryfilter/lang

### LINKED ISSUE #245: Filter DSL
I want to introduce a Filter "DSL" which would be implemented within `libimagentryfilter` to provide a `--filter <filter_string>` commandline argument which can be integrated into the commandline.

This DSL must provide:
- Access to all builtin filters
- Access to all non-builtin filters which are added from a library via the `Filter` trait

The interface from `libimagentryfilter` must provide:
- A builder for the commandline argument (trivial)
- A compiler function which translates `&str -> Result<Box<Filter>, FilterCompilerError>` (non-trivial)

---
# DSL

The DSL itself would be rather simple:
- Each filter has a name, this name can be called with `name()`
- inside the parentheses of a filter call, there can be arguments to modify the filter
  - For example: `header_field_eq(header.path.field, String(foo))` would return a `FieldEq` filter which would filter for the header field at `header.path.field` to be equal to the string `"foo"`.

Syntax elements such as `and` (or `&&`), `or` (or `||`), `not` (or `!`), `xor` (or `^`) should be provided.

So a more complex filter source would look like this:

```
imag module list --filter "header_field_eq(foo.bar, String(1)) && !header_field_eq(foo.baz, String(2))"
```
## Dreams...

I think the upper DSL would be nice, though here goes what I'd love to have:

```
imag module list --filter "$H.foo.bar == String(1) && $C.grep('Something')"
```

where `$H` is a variable for the header and `$C` is for the content. `$H.foo.bar` translates therefor to "The header field at 'foo.bar'" and `$C.grep()` translates to "The content should be grep'd".

If someone wants to play with parsing and so on, go ahead! `nom` is a (afaik) good crate for parsing things.
# Importance of this

This is not that crititcal by now, but will become rather important as soon as we start to implement bigger modules. The filter infrastructure which is already there should become a consistent interface all modules can use and this is the very heart of it. With this, we can provide _one_ filtering interface over _all_ modules, which I consider rather important to our main goals.

matthiasbeyer:
Turned out that this isn't as complex as I thought.

@TheNeikos do you see anything I might have forgotten?

Closes #245


TheNeikos:
Can you give an example call on how to use this?

matthiasbeyer:
My intend is to

```bash
imag-something list --filter-with 'entry.get_header()["imag"]["links"].length() == 3'
```

or something like this. Of course, this PR is WIP. The `entry` variable is not in scope, functions need to be defined for all the things...

And of course I should write tests.

TheNeikos:
Very well. I can see that it is useful. It gets a :+1: from me

mario-kr:
What's the status here? Is this the filter-language feature?

matthiasbeyer:
Yes, this is some experiments. It is not working and stalled.

PR #379, opened 2016-04-21
Libimagrt/output module path

matthiasbeyer:
Some more informations in the debugging output.


matthiasbeyer:
After checking the output I think the file path is enough. Will close this unmerged.

Issue #294, opened 2016-03-26
Store should automatically fallback to latest version of entry


matthiasbeyer:
If we load an entry `/foo/bar`, we do not pass the version of the entry which should be loaded.

The store should fallback to the last version instead of failing.

This is mainly useful for the user interface, of course.

neithernut:
This issue's headline does not match the description: do you mean the _latest_ version or the _oldest_?

matthiasbeyer:
Wah, yes. The latest.

matthiasbeyer:
This becomes really annoying, we should takle this soon. I added the `importance/high` label therefor.

mario-kr:
Question: With #667 , wouldnt the version part be removed, rendering this issue obsolete?

matthiasbeyer:
@mario-kr yes.

neithernut:
Well, than you should probably clode this issue once #667 is merged.

matthiasbeyer:
Obsolete now.

PR #391, opened 2016-05-12
Libimagstore/get

matthiasbeyer:
Getter for entries in the store.

This targets #294 as we should use implicit fallbacks to older versions, but we must provide an interface to find older versions of entries. This seems to be a good approach for me.

@TheNeikos what do you think?

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -324,6 +324,9 @@ impl Store {
 
     /// Borrow a given Entry. When the `FileLockEntry` is either `update`d or
     /// dropped, the new Entry is written to disk
+    ///
+    /// Implicitely creates a entry in the store if there is no entry with the id `id`. For a
+    /// non-implicitely-create look at `Store::get`.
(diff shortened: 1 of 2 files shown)


libimagstore/src/store.rs
```diff
@@ -352,7 +355,83 @@ impl Store {
                 }
 
             })
-   }
+    }
+
+    /// Get an entry from the store if it exists.
+    ///
+    /// This executes the {pre,post}_retrieve_aspects hooks.
+    pub fn get<'a, S: IntoStoreId>(&'a self, id: S) -> Result<Option<FileLockEntry<'a>>>
```
matthiasbeyer: We should rewrite some binaries, as they implicitely create entries sometimes.

For example tagging an entry "foo" with `imag-tag` works, even if "foo" does not exist yet, because `imag-tag` uses `Store::retrieve()` which implicitely creates the entry, which shouldn't happen in this case.

PR #392, opened 2016-05-12
Libimagentrylink/external link objects

matthiasbeyer:
Another external linking code rewrite.

This actually changes the interface for the external linking traits, as we need a functionality to get the objects which are generated behind the scenes.

I will start implementing the bookmarking module based on this PR and we'll see whether this patch works as intended. For more information, see the commit messages.


matthiasbeyer:
After some rethinking of my work: I guess this is the wrong approach to do this. We shouldn't alter the interface, as it makes perfect sense here.

And what I want can be modeled by using the `get_internal_links` functionality. So I will close this unmerged.

---

Clearification:

If you have an object which has `N` external links, you can get the external link objects (the actual Store entries which were generated by the `add_external_link()` code) by calling `get_internal_links()` and filtering for the external link store path.

What we should do: Add functionality to get all external link `StoreId` objects from a vector of `StoreId`s.

PR #400, opened 2016-05-14
Libimagutil/match helper

matthiasbeyer:
As I'm not comfortable with the macro foo... @TheNeikos could you review this changeset?


matthiasbeyer:
@TheNeikos This gets fishy now... could you tell me what I've done wrong here?

PR #410, opened 2016-05-15
Libimagerror/init

### LINKED ISSUE #408: Write Macro to generate default Error type structure
Almost all error types look the same. We have a `FooError` type, a `FooErrorKind` enum and some functions on these.

We could write a macro to generate this code. Would be way better and less code-duplication, IMHO.

matthiasbeyer:
Macro for generating error types.

Could replace a lot of boilerplate in the whole imag codebase.

I'd also vote for moving the `trace_error()` functions from `libimagutil` to this `libimagerror` crate, @TheNeikos what do you think?

---

Closes #408


matthiasbeyer:
Okay, this finally builds. Now we can also move the `trace_error` code, can't we?

matthiasbeyer:
Clearly, this will conflict with #370, as we have "custom" code in the error definition there.

@TheNeikos can you think of a way how to merge this PR with features to implement custom fields and code within the error types and functions? Like we introduce in #370? I guess your macro-foo is good enough for this, isn't it?

matthiasbeyer:
I just removed some commits from this PR. I will re-file them in a seperate PR, so we have more little PRs instead of one big one.

This should now build, so we can merge and rebase the other branches from my pipeline and open new PRs for them.

PR #414, opened 2016-05-17
Libimagerror/custom membersntypes

matthiasbeyer:
WIP to add custom member types to error generator macro.


matthiasbeyer:
@TheNeikos can you help here?

matthiasbeyer:
We could use the approach from #422 and split all things into more macros, but I guess there is a better way. Challenge @TheNeikos ?

matthiasbeyer:
@TheNeikos Does it look like you would've done it?

PR #416, opened 2016-05-17
Libimagerror/into error

matthiasbeyer:
Add `IntoError` Traits and implementation.


matthiasbeyer:
The solution from the last commits is really bad. I would prefer to generate a full `mod` with the macro, but then I have the problem that some users of the macro (libimagstore or libimagrt) want to implement some traits (Mostly `Into<T>`) on the error type.

@TheNeikos do you have an idea how to do this with the macro?

matthiasbeyer:
Damn it. How can I fix this? :cry:

matthiasbeyer:
#422 solves the issues we have here, I guess.

PR #418, opened 2016-05-18
Libimagentrytag/more helpers

matthiasbeyer:
Extract some functionality which was already in `libimagentrytag` and export it in new functions.

Does not change the functionality of anything, but adds more functions to the public interface of the crate.

### DIFF
--- a/libimagentrytag/src/ui.rs
+++ b/libimagentrytag/src/ui.rs
@@ -75,13 +75,9 @@ pub fn get_add_tags(matches: &ArgMatches) -> Option<Vec<Tag>> {
 ///
 /// Returns none if the argument was not specified
 pub fn get_remove_tags(matches: &ArgMatches) -> Option<Vec<Tag>> {
-    if let Some(v) = extract_tags(matches, tag_subcommand_remove_arg_name(), '-') {
-        return Some(v);
-    } else {
-        matches
-            .values_of(tag_subcommand_remove_arg_name())
-            .map(|values| values.map(String::from).collect())
-    }
+    let rem = tag_subcommand_remove_arg_name();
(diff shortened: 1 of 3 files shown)


libimagentrytag/src/ui.rs
```diff
@@ -60,14 +66,26 @@ pub fn tag_argument_name() -> &'static str {
 ///
 /// Returns none if the argument was not specified
 pub fn get_add_tags(matches: &ArgMatches) -> Option<Vec<Tag>> {
-    extract_tags(matches, "add-tags", '+')
+    if let Some(v) = extract_tags(matches, tag_subcommand_add_arg_name(), '+') {
+        return Some(v);
+    } else {
+        matches
+            .values_of(tag_subcommand_add_arg_name())
+            .map(|values| values.map(String::from).collect())
```
matthiasbeyer: Better would be `extract().or(matches...)`

libimagentrytag/src/ui.rs
```diff
@@ -60,14 +66,26 @@ pub fn tag_argument_name() -> &'static str {
...
 ///
 /// Returns none if the argument was not specified
 pub fn get_remove_tags(matches: &ArgMatches) -> Option<Vec<Tag>> {
-    extract_tags(matches, "remove-tags", '-')
+    if let Some(v) = extract_tags(matches, tag_subcommand_remove_arg_name(), '-') {
+        return Some(v);
+    } else {
+        matches
+            .values_of(tag_subcommand_remove_arg_name())
+            .map(|values| values.map(String::from).collect())
```
matthiasbeyer: Better would be `extract().or(matches...)`

PR #426, opened 2016-05-23
Libimagentrymarkdown/init

matthiasbeyer:
As #191 gets way to complex with all the markup languages implemented inside one crate, I opted for adding layers around single crates and implementing helper functionality upon them.

Later, #191 can be rewritten to use all the other crates to implement a unique interface for all of them.

---

This is my approach on adding some functionality in a wrapper around `hoedown` which might be useful for imag.

### DIFF
--- a/libimagentrymarkdown/src/link.rs
+++ b/libimagentrymarkdown/src/link.rs
@@ -1,13 +1,34 @@
+use error::MarkdownErrorKind as MEK;
 use result::Result;
 
 use hoedown::renderer::Render;
 use hoedown::Buffer;
 use hoedown::Markdown;
+use url::Url;
+
+use libimagerror::into::IntoError;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct Link {
     pub title: String,
     pub link: String,
+}
+
+impl Link {
+
+    /// Translate a `Link` into a `UrlLink`
+    fn into_urllink(self) -> Result<UrlLink> {
+        Url::parse(&self.link[..])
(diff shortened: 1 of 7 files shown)


matthiasbeyer:
Would be nice to add some tests and I guess this is rather simple to test, actually.

libimagentrymarkdown/src/link.rs
```diff
@@ -0,0 +1,85 @@
...
+impl Render for LinkExtractor {
+
+    fn link(&mut self,
+            _: &mut Buffer,
+            _: Option<&Buffer>,
+            link: Option<&Buffer>,
+            title: Option<&Buffer>)
+        -> bool
+    {
+        let link  = link.and_then(|l| l.to_str().ok()).map(String::from);
```
TheNeikos: Please don't use a lowercase `l`, rather something like `x` or `c` or w/e

IlLi|

libimagentrymarkdown/src/link.rs
```diff
@@ -0,0 +1,85 @@
...
+        let title = title.and_then(|l| l.to_str().ok()).map(String::from);
+
+        match (link, title) {
+            (Some(link), Some(title)) => {
+                self.links.push(Link { link: link, title: title });
+                true
```
TheNeikos: Going by the docs of hoedown you can return false here as you do not really render anything.

http://www.blaenkdenum.com/hoedown/src/hoedown/src/renderer/mod.rs.html#89-441

matthiasbeyer:
I guess I will merge this now. Enhancements can be done in new branches then!

PR #430, opened 2016-05-23
Reimplement Store::retrieve_for_module() with WalkDir

matthiasbeyer:
What the title says, basically.

This removes the dependency on `glob`.


matthiasbeyer:
Still WIP as this yields absolute pathes to root, not to store-root.

matthiasbeyer:
I can't get this to work. It is absolutely stupid that the `walkdir` crate returns always absolute paths. Who the hell decided for that BS?

PR #439, opened 2016-05-26
Remove duplicated StoreId member in FileLockEntry, use member from Entry

### LINKED ISSUE #405: FileLockEntry and Entry have StoreId
This seems to be duplicated data, as they both have a member of type `StoreId`. I guess we should investigate and remove the duplication if possible.

matthiasbeyer:
Closes #405 

@TheNeikos Please review this and whether I'm doing serious bullshit here or whether this is a good idea actually.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -327,7 +327,7 @@ impl Store {
             se
         });
 
-        let mut fle = FileLockEntry::new(self, Entry::new(id.clone()), id);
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
I _guess_ the `FileLockEntry::new()` function is only visible inside the `libimagstore::store` module, so these changes should work as-is. I might have overlooked something, though.

Lets see what travis tells us.

PR #440, opened 2016-05-26
Libimagstore/load unload hooks

### LINKED ISSUE #406: Hook system: Store load / Store unload hook operations?
Maybe we should also offer a way to run some hooks when the store get opened and before it gets closed.

This way git hooks would be _way_ simpler, as we can stage during the store modification and commit before the store gets `drop()`ed.

If we do this, it would also be nice to allow the hooks to pass a `State` object around. Mutable, of course. This would introduce another big overhaul over the hook system, but I guess it is for the better good, isn't it?

@neithernut @TheNeikos what do you guys think?

matthiasbeyer:
Closes #406 

Unfortunately, we cannot have store-load hooks, as the store is the object where the hooks have to be registered. And we cannot register hooks in the store object before the `::new()` call of course, so we have a chicken-egg problem here.

So this PR only implements the `store-unload` hooks.

### DIFF
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -3,6 +3,10 @@
 
 [store]
 
+# Hooks which get executed right before the Store is closed.
+# They get the store path as StoreId passed, so they can alter the complete
+# store, so these hooks should be chosen carefully.
+store-unload-hook-aspects  = [ "debug" ]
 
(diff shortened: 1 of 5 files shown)


matthiasbeyer:
@TheNeikos would appreciate a casual review here.

PR #446, opened 2016-05-27
Fix Store::get()

### LINKED ISSUE #445: Store::get() only gets from store internal cache, not FS
The store function `get` only asks the store cache for the entries, not the filesystem.

matthiasbeyer:
Closes #445 

@TheNeikos review? I know the code is ugly, but this small piece took ~2 hrs to craft, so I just committed it and awaiting review now.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -353,38 +353,12 @@ impl Store {
     /// Get an entry from the store if it exists.
     ///
     /// This executes the {pre,post}_retrieve_aspects hooks.
-    pub fn get<'a, S: IntoStoreId>(&'a self, id: S) -> Result<Option<FileLockEntry<'a>>>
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Wah, this implicitely creates entries in the store.

Still working on it... :cry:

matthiasbeyer:
@TheNeikos The problem that store entries are getting created are _way_ deep down in `LazyFile`.

I'm not sure we can change this, can we? @TheNeikos you are expert on this level... would be nice to hear from you!

matthiasbeyer:
Sometimes the simple solutions are the best.

TheNeikos:
Love your solution haha

matthiasbeyer:
Yeah ... I mean it is a hack, but a valid one, isn't it?

PR #316, opened 2016-03-31
Imag diary/init

matthiasbeyer:
Diary module based on #309 (libimagdiary).

Currently based on the PR for #309, so must be rebased as soon as #309 is merged.

PR #449, opened 2016-05-28
Libimagtimeui/init

matthiasbeyer:
Crate for building nice interfaces for specifying time.

Will be useful for fixing #316 (edit functionality) but also others maybe.


matthiasbeyer:
We don't need the error module, should remove this before merging.

PR #455, opened 2016-06-04
Adds a Makefile for building modules

kbknapp:
One can now type:

```
$> make
```

Which builds all the modules and places them in a `out/` directory of the project root for easy
finding.

Alternatively one can type:

```
$> make $module
```

Where `$module` is one of the `imag-*` names, such as `counter`, `imag-link`, etc.

To clean up the `out/` directory it's `$> make clean`

-----

The `make clean` could also run `cargo clean` if needbe, but I wasn't sure if that would be desirable or not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,33 +1,13 @@
-toml = imag-$@/Cargo.toml
-bin = imag-$@/target/debug/imag-$@
+toml = $@/Cargo.toml
+bin = $@/target/debug/$@
 
 default: all
 
 .PHONY: clean
 
-all: counter link notes store tag view
+all: imag-counter imag-link imag-notes imag-store imag-tag imag-view
 
-counter: prep
-	cargo build --manifest-path $(toml)
-	cp $(bin) out/
-
-link: prep
-	cargo build --manifest-path $(toml)
-	cp $(bin) out/
-
-notes: prep
-	cargo build --manifest-path $(toml)
-	cp $(bin) out/
-
-store: prep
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,38 @@
+toml = imag-$@/Cargo.toml
+bin = imag-$@/target/debug/imag-$@
+
+default: all
+
+.PHONY: clean
+
+all: counter link notes store tag view
+
+counter: prep
```
matthiasbeyer: Hm, this is interesting. Can one make this generic like this:

```Makefile
imag-%: prep
    cargo build --manifest-path $@/Cargo.toml
    cp $@/$(bin) out/
```

or something like this?

matthiasbeyer:
Very cool.

I'm not sure whether we want to change the `.travis.yml` for this... cargo will get a feature were one can specify _one_ ./target directory for _several_ crates, so this might be what we want for the `.travis.yml` later on, as we do not have to recompile all the way down every time compiling one of the crates.

Opinions?

This is clearly a nice enhancement!

kbknapp:
Great idea on the deduplication! I would say, if the current Travis build script is working, leave it as is until this new cargo feature lands. This make file really just makes it nicer for users to test things, or while developing to run a quick and easy to command to build and test.

matthiasbeyer:
Awesome! Thanks for the fixup. I'll merge as soon as travis succeeded (actually, I cannot merge before travis did not succeed,... :smile:)

PR #468, opened 2016-06-09
libimagref/init

matthiasbeyer:
A library for referencing out-of-store files.

I think this will be really helpful for implementing modules where one only references to out-of-store files, later on.


matthiasbeyer:
I think this is good and I will merge this if CI succeeds to get some progress here.

PR #469, opened 2016-06-10
Add Developers Certificate of Origin

matthiasbeyer:
I'm not sure here: We do not have a license-header in the files itself, but a license file which lives in the repository.

We might have to fix this.

@neithernut , @TheNeikos and maybe @llogiq ... what do you think on the subject of having a license statement in all files - is it necessary?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ can review your PR. This means that once _I am_ okay with your patchset, I will
 submit it as PR in the github repository, so more people can review it and CI
 can test it. I might come back to you if something broke in CI or someone has a
 suggestion how to improve your PR.
-I will keep you as author auf the commits.
+I will keep you as author of the commits.
 
 The following sections describe the way how to contribute with github.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -161,6 +161,47 @@ from the others by `"/`". See below:
 | test/missing              | Test missing                      | [search][search-test/missing]              |
 | test/new                  | New test                          | [search][search-test/new]                  |
 
+## Developer Certificate of Origin
+
+```
+Developer Certificate of Origin
+Version 1.1
+
+Copyright (C) 2004, 2006 The Linux Foundation and its contributors.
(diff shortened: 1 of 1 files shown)


llogiq:
First the obligatory disclaimer: I'm not a lawyer, so I'm not qualified to speak on license issues. For what it's worth, I usually don't bother, and I follow project licenses when I learn of them. However, having a license note in each source file reduces the risk that one source file gets copied without the license header "by mistake", so some projects opt to put one in each source file. I don't mind either way.

matthiasbeyer:
I guess I will add license headers in the files and ensure via CI that the header is present. Better safe than sorry!

matthiasbeyer:
@TheNeikos do you see what's going on here? If you `git ls-files` in this branch, there is no "imag-diary", but travis reports it... I don't see why...

matthiasbeyer:
Seems like the problem is that travis merges. Latest fixup tries to fix this by checking out the `$TRAVIS_COMMIT` after a successful `cargo build` step.

.travis.yml
```diff
@@ -77,6 +76,9 @@ addons:
 after_success:
     - travis-cargo --only stable doc-upload
 
+after_script:
+    - source .lgpl_check.sh
```
matthiasbeyer: @TheNeikos Will this actually break my build if the script calls `exit(1)`? I mean... does `after_script:` still cause a build failure if it fails on travis? I don't know...

TheNeikos: Afaik it does not

matthiasbeyer: @TheNeikos it seems like it actually does. Look at the latest evaluation. Or do you see something which failed the build besides the check?

matthiasbeyer:
Okay, now we have the confirmation that it actually is not what we wanted. :cry: 

I guess I have to rewrite the testing script :cry:

matthiasbeyer:
I guess this PR needs to be splitted. One for the Developers Certificate of Origin and the other for the license header.

matthiasbeyer:
I moved the LGPL-Check stuff to #479, this PR is for the Developers Certificate of Origin only.

PR #479, opened 2016-06-24
Add lgpl checker

matthiasbeyer:
Extracted from #469


matthiasbeyer:
I'll close this unmerged. I don't think it is necessary.

PR #480, opened 2016-06-25
Libimagerror/result helper

matthiasbeyer:
@TheNeikos what do you think?

I really need an Opinion here...


TheNeikos:
That's actually a really neat idea! :+1:

matthiasbeyer:
So I will merge this after travis succeeds. I will then open PRs for cleaning up a bit of code in the crates.

PR #481, opened 2016-06-26
Imag ref/init

matthiasbeyer:
Currently based on #468 (therefor contains commits of it), but will be rebased as soon as #468 is working as expected.


matthiasbeyer:
This could use https://github.com/kbknapp/clap-rs/blob/master/examples/15_custom_validator.rs#L17 (clap validators) to ensure that a referenced file exists when creating the reference and such...

PR #489, opened 2016-07-02
Libimagstore/storify

matthiasbeyer:
This is a PR where I try to introduce a concept to differentiate between internal and external store ids.

Internal store ids should contain the full path to an entry inside the store. They should be absolute to the filesystem root.

External store ids should be absolute to the store root they live in.

This makes things rather complicated, actually, as the internal setup of the store makes differentiation rather complicated.


matthiasbeyer:
This does not seem to break stuff... so I can merge it, I guess.

PR #492, opened 2016-07-04
fix command finding in imag/bin/src/main.rs

### LINKED ISSUE #491: imag binary --help doesnt show commands
Hello,
after calling cargo install for every binary and imag/bin, something like
```sh
imag view --help
```
works as expected, however
```sh
imag --help
```
doesnt actually show any comments.
This might as well be a local problem, so this Issue can be closed anytime.
greetings

mario-kr:
this should fix #491

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -68,17 +68,18 @@ fn get_commands() -> Vec<String> {
                         .into_iter()
                         .filter(|path| {
                             match path {
-                                &Ok(ref path) => path.path().starts_with(format!("{}/imag-", elem)),
+                                &Ok(ref p) => p.file_name()
+                                    .to_str()
+                                    .map_or(false, |filename| filename.starts_with("imag-")),
                                 &Err(_)   => false,
                             }
                         })
                         .filter_map(|x| x.ok())
-                        .map(|path| {
-                            path.path()
-                                .to_str()
-                                .map(String::from)
+                        .filter_map(|path| {
+                           path.file_name()
+                               .to_str()
+                               .map(String::from)
                         })
-                        .filter_map(|x| x)
                         .collect()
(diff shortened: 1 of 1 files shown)


bin/src/main.rs
```diff
@@ -68,15 +68,19 @@ fn get_commands() -> Vec<String> {
                         .into_iter()
                         .filter(|path| {
                             match path {
-                                &Ok(ref path) => path.path().starts_with(format!("{}/imag-", elem)),
+                                &Ok(ref path) => match path.file_name().to_str() {
+                                    Some(filename) => filename.starts_with("imag-"),
+                                    None => false,
+                                },
```
matthiasbeyer: Would be cleaner here to 

```rust
path.file_name()
    .to_str()
    .map(|f| f.starts_with("imag-"))
    .unwrap_or(false)
```
 IMHO.

bin/src/main.rs
```diff
@@ -68,15 +68,19 @@ fn get_commands() -> Vec<String> {
...
                         })
                         .filter_map(|x| x.ok())
                         .map(|path| {
-                            path.path()
-                                .to_str()
-                                .map(String::from)
+                            match path.file_name().to_str() {
+                                Some(filename) => Some(String::from(filename)),
+                                None => None,
+                            }
```
matthiasbeyer: Why not simply changing the `path()` call to `file_name()`? You don't need to replace everything by a match if you can simply `map` over things here!

matthiasbeyer:
Wow, another PR from you! Awesome!

matthiasbeyer:
Okay, this works... but please fix up the coding style as I mentioned above.

matthiasbeyer:
Thanks a lot!

PR #502, opened 2016-07-05
libimagstore: std::fs overlay and Store testing

### LINKED ISSUE #501: Test libimagstore
We can test `libimagstore` by wrapping all `std::fs` functions we need in an own submodule with functions that either just call the `std::fs` equivalent functions or, if compiled for testing, provide a pseudo-fs which might fail - so we are able to test.

I guess this is best possible if we use global state in the tests to set FS operations failing or such things... not sure on the details, though.

matthiasbeyer:
Closes #501 

@TheNeikos feedback highly appreciated here!

This must also implement a test infrastructure for this.. I'm not yet sure how to do this properly. I assume it would be an idea to pass a special  context via a global variable (in the test-case) which is the backend the tests talk to and which implements failures and such things.

### DIFF
--- a/libimagstore/src/fs.rs
+++ b/libimagstore/src/fs.rs
@@ -5,6 +5,111 @@
 //! should be inlined by the compiler then, so we have zero costs for this thin private layer over
 //! std::fs.
 
+#[cfg(test)]
+pub mod filesystem {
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
On hold in favor of #514

PR #487, opened 2016-06-30
Impl Display for StoreId

matthiasbeyer:
@TheNeikos 

Does this make any sense here?

PR #503, opened 2016-07-05
Fix Display impl for StoreId by using lossy string rep if PathBuf::to…

matthiasbeyer:
…_str() fails

As discussed in https://github.com/matthiasbeyer/imag/pull/487#discussion_r69544153


TheNeikos:
Yup!

PR #504, opened 2016-07-05
Makefile additional targets

mario-kr:
Added targets for `install` and `doc`
To generate a single doc for all of imag, the crate imagdoc was created, and a script to updates its Cargo.toml. New libraries need to be added in the list of that script. If there are no more `libimag*` directories without a Cargo.toml (after adding some features/cleaning no more used directories), the script may be altered to add the crates a bit more dynamical.
The `install` target iterates over the `imag-*` plus `bin` directories and calls `cargo install` in all of them.
The install fails if the binaries are already installed. The install-directory seems system-dependent, and i couldnt find a way to query cargo for it, so manual deletion is required to reinstall (update) the binaries.


matthiasbeyer:
Hey, awesome! I will annotate your PR with a few improvements now...

Makefile
```diff
@@ -1,16 +1,29 @@
...
 
+imag: prep
+	cargo build --manifest-path bin/Cargo.toml
+	cp bin/target/debug/imag out/
+
+install:
+	$(foreach b,$(bin_crate_dirs),$(shell pushd $(b) 1>/dev/null; cargo install 1>&2; popd 1>/dev/null;))
+
+doc:
+	./imagdoc/update_Cargo.toml.sh ./
```
matthiasbeyer: Could you put the location of this script in a variable rather than calling it in the target? Its cleaner this way...

imagdoc/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "imag-link"
```
matthiasbeyer: This is clearly a C/P failure.

imagdoc/update_Cargo.toml.sh
```diff
@@ -0,0 +1,31 @@
...
+
+CARGOTOML=
+case $1 in
+    (*/) CARGOTOML="${1}imagdoc/Cargo.toml";;
+    (*) CARGOTOML="$1/imagdoc/Cargo.toml";;
+esac
+
+# package section
+echo "[package]
+name = \"imag-link\"
```
matthiasbeyer: And another C/P failure here.

matthiasbeyer:
I will provide one or two commits for this PR if you don't mind...

imagdoc/update_Cargo.toml.sh
mario-kr: that's the only c/p failure. fix incoming

matthiasbeyer:
Feel free to cherry-pick 

* https://github.com/matthiasbeyer/imag/commit/ffa1d3f14c6640970b22013158f8fab3d7efe011
* https://github.com/matthiasbeyer/imag/commit/41ff89fb8c784d3218d5b954b1a2fcc02711901a

from the [pr-504 branch I just pushed](https://github.com/matthiasbeyer/imag/tree/pr-504)

imagdoc/update_Cargo.toml.sh
```diff
@@ -0,0 +1,31 @@
...
+version = \"0.1.0\"
+authors = [\"Mario Krehl <mario-krehl@gmx.de>\"]
+
+[dependencies]" > $CARGOTOML
+
+# using the command below will cause an error in cargo doc if one of the
+# directories doesnt contain a Cargo.toml. Those are either not implemented
+# features or features which were removed/moved elsewhere. Therefore, a
+# manually maintained list is used
+#for lib in $(find $1 -maxdepth 1 -name "libimag*" -type d); do
```
matthiasbeyer: You could, rather than maintaining a list manually, create a function which does the checks...

```bash
function list_dirs_with_cargotoml() {
    for d in $(find $1 -maxdepth 1 -name "libimag*" -type d); do
        [[ -e ${d}/Cargo.toml]] && echo "${d}"
    done
}
```

or something like this, and then call it in `for lib in $(list_dirs_with_cargotoml)`

imagdoc/update_Cargo.toml.sh
```diff
@@ -23,9 +20,14 @@ authors = [\"Mario Krehl <mario-krehl@gmx.de>\"]
 # directories doesnt contain a Cargo.toml. Those are either not implemented
 # features or features which were removed/moved elsewhere. Therefore, a
 # manually maintained list is used
-#for lib in $(find $1 -maxdepth 1 -name "libimag*" -type d); do
-for lib in libimagcounter libimagdiary libimagentryfilter libimagentrylink libimagentrylist libimagentrymarkdown libimagentryselect libimagentrytag libimagentryview libimagerror libimaginteraction libimagnotes libimagrt libimagstore libimagstorestdhook libimagtimeui libimagutil; do
-echo "$lib = { path = \"../$lib\" }"
+echo $1
+for lib in $(find $1 -maxdepth 1 -name "libimag*" -type d | sed -r "s|$1/?||"); do
```
matthiasbeyer: You could propably add a `| sort` here, so we have it sorted in the output.

imagdoc/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+libimagentryfilter = { path = "../libimagentryfilter" }
+libimagentrytag = { path = "../libimagentrytag" }
+libimagstore = { path = "../libimagstore" }
+libimagentryview = { path = "../libimagentryview" }
+libimaginteraction = { path = "../libimaginteraction" }
+libimagcounter = { path = "../libimagcounter" }
+libimagerror = { path = "../libimagerror" }
+libimagdiary = { path = "../libimagdiary" }
+libimagrt = { path = "../libimagrt" }
+
```
matthiasbeyer: I'm not sure we would want to commit this file to the repository. I would actually put it in the gitignore file.

As we generate it from a script, there is no point in committing it to the repository, IMO.

mario-kr: btw, the out/ directory should also be placed in the .gitignore, shoudnt it? or is it already?

matthiasbeyer: Nope, it is not... but great suggestion!

matthiasbeyer:
@mario-kr the documentation stuff was already done in https://github.com/matthiasbeyer/imag/pull/589 - can you remove the commits for the documentation stuff from this branch and only focus on makefile stuff?

Maybe #618 gets covered by this as well? Would be nice!

mario-kr:
superceeded by #646 , closing this PR.
I'm keeping the branch on my fork until this is merged, so you have a source of inspiration ;)

PR #513, opened 2016-07-06
Libimagstore/hooks/optionally deny mutable hooks

### LINKED ISSUE #302: Store Hook Aspect configuration: Setting to deny mutable hooks
Would be nice to have a setting in the configuration file which allows to configure whether a certain aspect is allowed to execute mutable hooks.

If the setting is `false`, the aspect simply does not execute a Hook (and prints a `warn!()`ing, of course).

---

This _could_ be an easy task, though I'm not sure here and I'd label it `medium` therefor.

---

This is kinda secruity-related, as if you don't want to execute hooks which alter data, you can disable it with such a setting.

matthiasbeyer:
Closes #302

### DIFF
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -24,6 +24,7 @@ post-delete-hook-aspects   = [ "debug" ]
 
 [[aspects.debug]]
 parallel = false
+mutable_hooks = false
 
 [store.hooks]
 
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
Lets see whether this actually works, as I can remember something about this that we are only able to have mutable access at the moment because of some design flaws in the hook system.

PR #514, opened 2016-07-06
Add Abstraction layer for FS in Store for better testing

TheNeikos:


matthiasbeyer:
Hiho! Any progress on this?

libimagstore/src/store.rs
```diff
@@ -140,7 +139,6 @@ impl StoreEntry {
                 // TODO:
                 let mut file = file.unwrap();
                 let entry = Entry::from_file(self.id.clone(), &mut file);
-                file.seek(SeekFrom::Start(0)).ok();
```
matthiasbeyer: Why do you remove this line?

matthiasbeyer: Read: Why does it make sense to remove this line and doesn't introduce a bug? :smiley:

matthiasbeyer:
Closed as #569 superceeds.

PR #525, opened 2016-07-08
Update doc

matthiasbeyer:
This aims for a rewrite of the documentation. This rewrite will move the documentation from a technical document to a user documentation, which is in my opinion a good move.

imag is a moving target. Keeping a specification or technical documentation is just not possible. Having a user-documentation, though, makes more sense as we can show users how to use imag. The commandline interface are (thanks to clap here) another part of the documentation for imag, but this document is more for a "how can I do things with imag" rather than another user interface documentation.


matthiasbeyer:
I guess we can merge this as well...

PR #529, opened 2016-07-12
libimagstore: Entry does only see store-absolute id

### LINKED ISSUE #523: libimaglink links with absolute (to FS root) pathes
This is a bug.

We have to link with absolute pathes to store root, not to FS.

matthiasbeyer:
Closes #523 

Here we can clearly see that we need type abstraction over this, right @TheNeikos?


matthiasbeyer:
Would be superceded by #530 .,,

matthiasbeyer:
Superceeded by #667

PR #530, opened 2016-07-12
libimagstore/internal storeid type

matthiasbeyer:
@TheNeikos please look at this.

I'm trying to solve this huge source of bugs right now... but I still think it is a bit ugly how I solve it.

Maybe we should rework the _whole_ codebase (I mean with ignoring backwards compatibility) to fix this issue.

This is another approach to make the internal use of store ids absolute _to the store root_ rather than to the filesystem root.

I'm not sure I did it right and I guess there are still flaws. For instance, we might preserve the actual `StoreId` instance when creating a `EntryStoreId` instance from it and offer the possibility to fully restore it in `Into<Storeid>`.

### DIFF
--- a/libimagstore/src/entrystoreid.rs
+++ b/libimagstore/src/entrystoreid.rs
@@ -27,6 +27,14 @@ impl Into<StoreId> for EntryStoreId {
 
 }
 
+impl Deref for EntryStoreId {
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
Superceeded by #667

Issue #499, opened 2016-07-05
Optimize: `fold(Ok(()), ...`


matthiasbeyer:
@TheNeikos I ping you because we talked yesterday.

One thing that could be optimized is the use of `Iterator::fold` to fold an `Ok(())` and to break iterating on error.

We have 8 occurrences in the code where I do

```rust
foo.iter().fold(Ok(()), |acc, elem| acc.and_then(|_| elem.do_something()))
```

We could optimize this by providing a `Iterator::fold_ok` which hides the passing ok `Ok(())` and the use of `and_then()` and simply applies the closure.

This could be done in a new helper crate `libimagiter` or in `libimagutil` (I guess the latter is sufficient).

bjorn3:
Could you use `.map(_).collect::<Result<Vec<T>, E>>()`?

matthiasbeyer:
No, as `fold()` calulates _one_ result rather than a list of results (as in value, not as in `Result`).

Have a look at [the documentation for fold](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold), it is not an advanced `map()`! The result is of type `Result<T, E>` rather than `Result<Vec<T>, E>` as you described it.

bjorn3:
Oops

jsirois:
Can this be closed or is there more work to do?

matthiasbeyer:
Oh, it wasn't closed by the merge ... I'm sorry. Yes it can be closed! Thanks for mentioning!

Issue #138, opened 2016-01-23
[RFC] [Store]: Entry-getter to iterate over entries of a module


matthiasbeyer:
Problem:

I'm a module which allowes the user to tag entries. I want to offer an option "list all entries with tag 'X'" to the user. I have no possibility to load all entries which belong to me.

The proposed solution:

Provide a `Store::retrieve_for_module()` which allows to iterate over all entries for given module. Version foo should be considered, of course.

Such an interface would look something like this:

```rust
pub fn retrieve_for_module(&self, mod_name: &str, mod_version: &str) -> FileLockEntryIterator {
    unimplemented!()
}
```

neithernut:
There are a few considerations to make:
 * mod_version being a raw string is probably not as ideal as it could be. Maybe we could use a special data type for that.
 * What are the exact semantics of `mod_version`? Is it the minimum version, the exact version or a range (possibly implicitly provided by omitting e.g. the minor level)? Can this be an optional parameter?
 * Of what kind should the iterator be? @matthiasbeyer proposes an iterator over `FileLockEntry`s, but this is not ideal for all cases, especially for cases where items are just read (he himself described such a use-case). Iterating over `StoreId`s would give you more flexibility, but it would also make things more complicated for the caller.

I'd like to propose having an alternative approach:
 * Have a method `pub fn retrieve_for_module(&self, mod_name: &str) -> StoreIdIterator`.
 * Provide facilities for easy filtering by available versions (which can be easily used with `Iterator::filter()`)
 * Users can easily invoke `Iterator::map()` with a closure calling `Store::retrieve()` or `retrieve::copy()`, just as they please.

----

Thinking about this: maybe the signatures of `Store::retrieve()` and `Store::retrieve_copy()` have to be adjusted to accommodate for the fact that we know version entries? I'll open an issue for this.

matthiasbeyer:
I agree with you on the `StoreIdIterator` part. Filtering would then be done by the callee, which is completely fine.

PR #532, opened 2016-07-13
Add a `fold_ok` utility.

jsirois:
Add a utility that folds an iterator into a result and uses it to reduce
boilerplate in the codebase.

### DIFF
--- a/libimagstore/src/hook/aspect.rs
+++ b/libimagstore/src/hook/aspect.rs
@@ -1,5 +1,5 @@
 use libimagerror::trace::trace_error;
-use libimagutil::iter::fold_ok;
+use libimagutil::iter::FoldResult;
 
 use store::FileLockEntry;
 use storeid::StoreId;
@@ -46,7 +46,7 @@ impl StoreIdAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_defresult(|accessor| {
             let res = match accessor {
                 &HDA::StoreIdAccess(accessor) => accessor.access(id),
                 _ => unreachable!(),
@@ -88,7 +88,7 @@ impl MutableHookDataAccessor for Aspect {
         // More sophisticated version would check whether there are _chunks_ of
         // NonMutableAccess accessors and execute these chunks in parallel. We do not have
         // performance concerns yet, so this is okay.
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_defresult(|accessor| {
             let res = match accessor {
                 &HDA::MutableAccess(ref accessor)    => accessor.access_mut(fle),
                 &HDA::NonMutableAccess(ref accessor) => accessor.access(fle),
@@ -118,7 +118,7 @@ impl NonMutableHookDataAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_defresult(|accessor| {
             let res = match accessor {
                 &HDA::NonMutableAccess(accessor) => accessor.access(fle),
                 _ => unreachable!(),
--- a/libimagstore/src/hook/aspect.rs
+++ b/libimagstore/src/hook/aspect.rs
@@ -1,4 +1,5 @@
 use libimagerror::trace::trace_error;
+use libimagutil::iter::fold_ok;
 
 use store::FileLockEntry;
 use storeid::StoreId;
@@ -45,29 +46,25 @@ impl StoreIdAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        accessors
-            .iter()
-            .fold(Ok(()), |acc, accessor| {
-                acc.and_then(|_| {
-                    let res = match accessor {
-                        &HDA::StoreIdAccess(accessor) => accessor.access(id),
-                        _ => unreachable!(),
-                    };
+        fold_ok(accessors.iter(), |accessor| {
+            let res = match accessor {
+                &HDA::StoreIdAccess(accessor) => accessor.access(id),
+                _ => unreachable!(),
+            };
 
-                    match res {
-                        Ok(res) => Ok(res),
-                        Err(e) => {
-                            if !e.is_aborting() {
-                                trace_error(&e);
-                                // ignore error if it is not aborting, as we printed it already
-                                Ok(())
-                            } else {
-                                Err(e)
-                            }
-                        }
+            match res {
+                Ok(res) => Ok(res),
+                Err(e) => {
+                    if !e.is_aborting() {
+                        trace_error(&e);
+                        // ignore error if it is not aborting, as we printed it already
+                        Ok(())
+                    } else {
+                        Err(e)
                     }
-                })
-            })
+                }
+            }
+        })
     }
 }
 
@@ -91,27 +88,25 @@ impl MutableHookDataAccessor for Aspect {
         // More sophisticated version would check whether there are _chunks_ of
         // NonMutableAccess accessors and execute these chunks in parallel. We do not have
         // performance concerns yet, so this is okay.
-        accessors.iter().fold(Ok(()), |acc, accessor| {
-            acc.and_then(|_| {
-                let res = match accessor {
-                    &HDA::MutableAccess(ref accessor)    => accessor.access_mut(fle),
-                    &HDA::NonMutableAccess(ref accessor) => accessor.access(fle),
-                    _ => unreachable!(),
-                };
+        fold_ok(accessors.iter(), |accessor| {
+            let res = match accessor {
+                &HDA::MutableAccess(ref accessor)    => accessor.access_mut(fle),
+                &HDA::NonMutableAccess(ref accessor) => accessor.access(fle),
+                _ => unreachable!(),
+            };
 
-                match res {
-                    Ok(res) => Ok(res),
-                    Err(e) => {
-                        if !e.is_aborting() {
-                            trace_error(&e);
-                            // ignore error if it is not aborting, as we printed it already
-                            Ok(())
-                        } else {
-                            Err(e)
-                        }
+            match res {
+                Ok(res) => Ok(res),
+                Err(e) => {
+                    if !e.is_aborting() {
+                        trace_error(&e);
+                        // ignore error if it is not aborting, as we printed it already
+                        Ok(())
+                    } else {
+                        Err(e)
                     }
                 }
-            })
+            }
         })
     }
 }
@@ -123,29 +118,25 @@ impl NonMutableHookDataAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        accessors
-            .iter()
-            .fold(Ok(()), |acc, accessor| {
-                acc.and_then(|_| {
-                    let res = match accessor {
-                        &HDA::NonMutableAccess(accessor) => accessor.access(fle),
-                        _ => unreachable!(),
-                    };
+        fold_ok(accessors.iter(), |accessor| {
+            let res = match accessor {
+                &HDA::NonMutableAccess(accessor) => accessor.access(fle),
+                _ => unreachable!(),
+            };
 
-                    match res {
-                        Ok(res) => Ok(res),
-                        Err(e) => {
-                            if !e.is_aborting() {
-                                trace_error(&e);
-                                // ignore error if it is not aborting, as we printed it already
-                                Ok(())
-                            } else {
-                                Err(e)
-                            }
-                        }
+            match res {
+                Ok(res) => Ok(res),
+                Err(e) => {
+                    if !e.is_aborting() {
+                        trace_error(&e);
+                        // ignore error if it is not aborting, as we printed it already
+                        Ok(())
+                    } else {
+                        Err(e)
                     }
-                })
-            })
+                }
+            }
+        })
     }
 }
 
--- a/libimagentrylist/Cargo.toml
+++ b/libimagentrylist/Cargo.toml
@@ -14,3 +14,6 @@ path = "../libimagstore"
 [dependencies.libimagerror]
 path = "../libimagerror"
 
+[dependencies.libimagutil]
+path = "../libimagutil"
+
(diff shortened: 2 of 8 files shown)


jsirois:
This is intended to address #499.
I was directed here by "This Week In Rust #138" and thought I'd finally try my hand at some rust.

jsirois:
If this looks reasonably OK I'll add a unit test for `fold_ok`.

libimagstore/src/hook/aspect.rs
```diff
@@ -45,29 +46,25 @@ impl StoreIdAccessor for Aspect {
...
-                                Err(e)
-                            }
-                        }
+        fold_ok(accessors.iter(), |accessor| {
+            let res = match accessor {
+                &HDA::StoreIdAccess(accessor) => accessor.access(id),
+                _ => unreachable!(),
+            };
+
+            match res {
```
jsirois: Note to self, follow up in a new PR and convert this `match res {` and the next 2 to `res.map_err...`

matthiasbeyer:
Wow, awesome that you want to do this. This is so awesome, thanks a lot.

I have a few points which need to be improved with this PR, but I also got some inspiration for some more beat things we could do in this PR if you want to. I'm rather tired right now, so I will come back tomorrow and provide some in-depth review for this, if this is okay with you.

Thanks a lot for your contribution, it is awesome that more people start to contribute to this codebase!

jsirois:
I can do something like this, but the ~implicit conversion like behavior seems a bit odd, in particular, the imports of `IteratorEx` appear unused:
```diff
$ git diff
diff --git a/libimagentrylist/src/listers/line.rs b/libimagentrylist/src/listers/line.rs
index 241e39d..dd048fc 100644
--- a/libimagentrylist/src/listers/line.rs
+++ b/libimagentrylist/src/listers/line.rs
@@ -5,7 +5,7 @@ use lister::Lister;
 use result::Result;
 
 use libimagstore::store::FileLockEntry;
-use libimagutil::iter::fold_ok;
+use libimagutil::iter::IteratorEx;
 
 pub struct LineLister<'a> {
     unknown_output: &'a str,
@@ -23,11 +23,11 @@ impl<'a> LineLister<'a> {
 
 impl<'a> Lister for LineLister<'a> {
 
-    fn list<'b, I: Iterator<Item = FileLockEntry<'b>>>(&self, entries: I) -> Result<()> {
+    fn list<'b, I: IteratorEx<Item = FileLockEntry<'b>>>(&self, entries: I) -> Result<()> {
         use error::ListError as LE;
         use error::ListErrorKind as LEK;
 
-        fold_ok(entries, |entry| {
+        entries.fold_result(|entry| {
             write!(stdout(), "{:?}\n", entry.get_location().to_str().unwrap_or(self.unknown_output))
                 .map_err(|e| LE::new(LEK::FormatError, Some(Box::new(e))))
         })
diff --git a/libimagentrylist/src/listers/path.rs b/libimagentrylist/src/listers/path.rs
index d08e818..7069812 100644
--- a/libimagentrylist/src/listers/path.rs
+++ b/libimagentrylist/src/listers/path.rs
@@ -5,7 +5,7 @@ use lister::Lister;
 use result::Result;
 
 use libimagstore::store::FileLockEntry;
-use libimagutil::iter::fold_ok;
+use libimagutil::iter::IteratorEx;
 
 pub struct PathLister {
     absolute: bool,
@@ -23,11 +23,11 @@ impl PathLister {
 
 impl Lister for PathLister {
 
-    fn list<'a, I: Iterator<Item = FileLockEntry<'a>>>(&self, entries: I) -> Result<()> {
+    fn list<'a, I: IteratorEx<Item = FileLockEntry<'a>>>(&self, entries: I) -> Result<()> {
         use error::ListError as LE;
         use error::ListErrorKind as LEK;
 
-        fold_ok(entries, |entry| {
+        entries.fold_result(|entry| {
             Ok(entry.get_location().clone())
                 .and_then(|pb| {
                     if self.absolute {
diff --git a/libimagstore/src/hook/aspect.rs b/libimagstore/src/hook/aspect.rs
index c58f34f..9f12c5a 100644
--- a/libimagstore/src/hook/aspect.rs
+++ b/libimagstore/src/hook/aspect.rs
@@ -1,5 +1,5 @@
 use libimagerror::trace::trace_error;
-use libimagutil::iter::fold_ok;
+use libimagutil::iter::IteratorEx;
 
 use store::FileLockEntry;
 use storeid::StoreId;
@@ -46,7 +46,7 @@ impl StoreIdAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_result(|accessor| {
             let res = match accessor {
                 &HDA::StoreIdAccess(accessor) => accessor.access(id),
                 _ => unreachable!(),
@@ -88,7 +88,7 @@ impl MutableHookDataAccessor for Aspect {
         // More sophisticated version would check whether there are _chunks_ of
         // NonMutableAccess accessors and execute these chunks in parallel. We do not have
         // performance concerns yet, so this is okay.
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_result(|accessor| {
             let res = match accessor {
                 &HDA::MutableAccess(ref accessor)    => accessor.access_mut(fle),
                 &HDA::NonMutableAccess(ref accessor) => accessor.access(fle),
@@ -118,7 +118,7 @@ impl NonMutableHookDataAccessor for Aspect {
             return Err(HE::new(HEK::AccessTypeViolation, None));
         }
 
-        fold_ok(accessors.iter(), |accessor| {
+        accessors.iter().fold_result(|accessor| {
             let res = match accessor {
                 &HDA::NonMutableAccess(accessor) => accessor.access(fle),
                 _ => unreachable!(),
diff --git a/libimagstore/src/store.rs b/libimagstore/src/store.rs
index 7ca8cd3..710dd71 100644
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -37,7 +37,7 @@ use hook::position::HookPosition;
 use hook::Hook;
 
 use libimagerror::into::IntoError;
-use libimagutil::iter::fold_ok;
+use libimagutil::iter::IteratorEx;
 
 use self::glob_store_iter::*;
 
@@ -693,16 +693,14 @@ impl Store {
                             id: &StoreId)
         -> HookResult<()>
     {
-        fold_ok(
-            match aspects.lock() {
-                Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
-                Ok(g) => g
-            }.iter(),
-            |aspect| {
-                debug!("[Aspect][exec]: {:?}", aspect);
-                (aspect as &StoreIdAccessor).access(id)
-            }).map_err(Box::new)
-                .map_err(|e| HookErrorKind::HookExecutionError.into_error_with_cause(e))
+        match aspects.lock() {
+            Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
+            Ok(g) => g
+        }.iter().fold_result(|aspect| {
+            debug!("[Aspect][exec]: {:?}", aspect);
+            (aspect as &StoreIdAccessor).access(id)
+        }).map_err(Box::new)
+            .map_err(|e| HookErrorKind::HookExecutionError.into_error_with_cause(e))
     }
 
     fn execute_hooks_for_mut_file(&self,
@@ -710,16 +708,14 @@ impl Store {
                                   fle: &mut FileLockEntry)
         -> HookResult<()>
     {
-        fold_ok(
-            match aspects.lock() {
-                Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
-                Ok(g) => g
-            }.iter(),
-            |aspect| {
-                debug!("[Aspect][exec]: {:?}", aspect);
-                aspect.access_mut(fle)
-            }).map_err(Box::new)
-                .map_err(|e| HookErrorKind::HookExecutionError.into_error_with_cause(e))
+        match aspects.lock() {
+            Err(_) => return Err(HookErrorKind::HookExecutionError.into()),
+            Ok(g) => g
+        }.iter().fold_result(|aspect| {
+            debug!("[Aspect][exec]: {:?}", aspect);
+            aspect.access_mut(fle)
+        }).map_err(Box::new)
+            .map_err(|e| HookErrorKind::HookExecutionError.into_error_with_cause(e))
     }
 
 }
diff --git a/libimagutil/src/iter.rs b/libimagutil/src/iter.rs
index 0f682bf..577910c 100644
--- a/libimagutil/src/iter.rs
+++ b/libimagutil/src/iter.rs
@@ -1,8 +1,23 @@
-/// Processes `iter` returning the last successful result or the first error.
-pub fn fold_ok<X, I, R, E, F>(iter: I, mut func: F) -> Result<R, E>
-    where I: Iterator<Item = X>,
-          R: Default,
-          F: FnMut(X) -> Result<R, E>
-{
-    iter.fold(Ok(R::default()), |acc, item| acc.and_then(|_| func(item)))
+/// Enhanced `Iterator` methods.
+pub trait IteratorEx {
+    type Item;
+
+    /// Processes all contained items returning the last successful result or the first error.
+    fn fold_result<R, E, F>(self, mut func: F) -> Result<R, E>
+        where R: Default,
+              F: FnMut(Self::Item)
+        -> Result<R, E>;
 }
+
+impl<X, I: Iterator<Item = X>> IteratorEx for I {
+    type Item = X;
+
+    fn fold_result<R, E, F>(self, mut func: F) -> Result<R, E>
+        where R: Default,
+              F: FnMut(Self::Item)
+        -> Result<R, E>
+    {
+        self.fold(Ok(R::default()), |acc, item| acc.and_then(|_| func(item)))
+    }
+}
+
````

**Comment by matthiasbeyer** [comment]
Wow, :+1: :+1: :+1: what you just suggested in your last comment what I wanted to suggtest _almost exactly_.

I would not use the `Default` trait, as this makes the implementation for this dependent on it, which is less flexible (as not all types implement `Default`). I would pass the default argument to the `fold_result()` function as parameter, actually. I guess that should be possible for empty tuples (`()`) as well. 

One could implement a version which uses the `Default` trait _upon_ the version where you pass the default object which gets wrapped in an `Ok()` then... this way we would have a generic `fold_result()` and another function, possibly `fold_defresult` to fold with a `Default`.

I would name the trait `FoldResult`, actually... but you know... naming things. :smile: 

---

Besides that: With the patch you posted as comment, you don't need to change the type argument in the `Lister` implementation(s), as you implement the `IteratorEx` (or `FoldResult` or whatever) for all `Iterator`s ... so everything is fine with the current version(s)!

---

On your note:

> the ~implicit conversion like behavior seems a bit odd

Maybe it looks like a conversion for you (I don't know what your background is, but I guess it is C++ maybe? Tell me if I'm wrong here :smile:) but it isn't. There's absolutely no type conversions involved in the code. I don't know how it works on a technical level, but I guess you just add another function to the vtable of the type when implementing a trait on it... and that's basically it... so no conversion and hence no runtime cost! :+1: awesome! :smile: 

On the other note:

> the imports of `IteratorEx` appear unused

We can look at that as soon your code is in the PR. :dancers:

**Comment by jsirois** [comment]
OK - See what you think. What I find unsettling is in an expression like this: `x.iter().fold_defresult(...` there is no hint what type provides `fold_defresult`.  You have to hunt through the use statements to find what typeclass is providing the method. This may be why you suggested naming the trait after the method though, since a `use ... ::FoldResult` is a very likely candidate to provide a method of the same name!

I wanted to provide a bit more convenience and save the `.iter()` calls but failed, not sure if you can provide guidance:
```diff
$ git diff
diff --git a/libimagutil/src/iter.rs b/libimagutil/src/iter.rs
index 89db6f0..c289c36 100644
--- a/libimagutil/src/iter.rs
+++ b/libimagutil/src/iter.rs
@@ -28,3 +28,13 @@ impl<X, I: Iterator<Item = X>> FoldResult for I {
     }
 }
 
+impl<X, I: Iterator<Item = X>, II: IntoIterator<Item = X, IntoIter = I>> FoldResult for II {
+    type Item = X;
+
+    fn fold_result<R, E, F>(self, default: R, mut func: F) -> Result<R, E>
+        where F: FnMut(Self::Item) -> Result<R, E>
+    {
+        self.into_iter().fold_result(default, func)
+    }
+}
+
```

Leads to this:
```
$ cargo build --manifest-path libimagutil/Cargo.toml 
   Compiling libimagutil v0.2.0 (file:///home/jsirois/dev/matthiasbeyer/jsirois-imag/libimagutil)
libimagutil/src/iter.rs:31:1: 39:2 error: conflicting implementations of trait `iter::FoldResult`: [E0119]
libimagutil/src/iter.rs:31 impl<X, I: Iterator<Item = X>, II: IntoIterator<Item = X, IntoIter = I>> FoldResult for II {
                           ^
libimagutil/src/iter.rs:31:1: 39:2 help: run `rustc --explain E0119` to see a detailed explanation
libimagutil/src/iter.rs:21:1: 29:2 note: conflicting implementation is here:
libimagutil/src/iter.rs:21 impl<X, I: Iterator<Item = X>> FoldResult for I {
                           ^
error: aborting due to previous error
error: Could not compile `libimagutil`.

To learn more, run the command again with --verbose.
```

matthiasbeyer:
Wow, your latest push looks good.

> I wanted to provide a bit more convenience and save the .iter() calls but failed, not sure if you can provide guidance

I'm not sure this is possible, because here we actually have a type conversion. For example: if you call the [`iter()` function](https://doc.rust-lang.org/std/primitive.slice.html#method.iter) on a slice, you convert that slice into a `Iter` object, which then implements `Iterator` (the trait). Because of this type conversion, which is explicit via the `iter()` call, what is good, you cannot do what you want to do. (At least as far as I know).

Besides, you would need to offer two variants of the implementation, one using `iter()` and another using `into_iter()` - so the benefit would be _very_ minimal.

---

Do you want me to review your code now or do you want to make some more adjustments? I would also include coding-style comments if this is okay with you... :smile:

jsirois:
Go for it on the review.

libimagstore/src/hook/aspect.rs
matthiasbeyer: I'm not sure whether this is possible, as a `map_err()` expects the closure to return an Error... but we return either a `Ok(())` or an `Err(e)`, depending on whether the `Error` was aborting.

But of course, feel free to play with the code! :+1:

matthiasbeyer:
Okay, I had a close look and I actually cannot find things that I wouldn't merge as-is. So I'll wait for the CI and if you're fine with it I would merge this after CI succeeds! :+1:

matthiasbeyer:
I guess we can merge this... any objections?

jsirois:
Nope!

matthiasbeyer:
Thank you very much for your contribution! I hope there will be some more PRs from you! :smile:

libimagstore/src/hook/aspect.rs
jsirois: Ah yes.

PR #534, opened 2016-07-14
Temporary fix: Check whether an StoreId is an external link by str::c…

matthiasbeyer:
…ontains()

This is a temporary fix as long as we do not have a clear model how we
handle internal/external StoreId types.

@TheNeikos, @neithernut here we can see what the problems are if we do not have a clear invariant for this type. :cry:


matthiasbeyer:
The second commit contains a more robust version, actually.

PR #535, opened 2016-07-14
Libimagcounter/units

### LINKED ISSUE #516: [counter] Add units support
Would be nice to have a way to specify a unit when creating a counter.
That unit should also be printed when the counters are listed.

This is a rather easy task. It requires changes in `imag-counter` and `libimagcounter`.

asuivelentine:
Closes #516


libimagcounter/src/counter.rs
```diff
@@ -51,6 +52,11 @@ impl<'a> Counter<'a> {
                 if setres.is_err() {
                     return Err(CE::new(CEK::StoreWriteError, Some(Box::new(setres.unwrap_err()))));
                 }
+
+                let setres = header.set("counter.unit", Value::String(unit));
+                if setres.is_err() {
```
matthiasbeyer: You can join this line and the one above.

libimagcounter/src/counter.rs
```diff
@@ -121,6 +127,17 @@ impl<'a> Counter<'a> {
             })
     }
 
+    pub fn unit(&self) -> Result<CounterUnit> {
+        self.fle.get_header().read("counter.unit")
+            .map_err(|e| CE::new(CEK::StoreWriteError, Some(Box::new(e))))
+            .and_then(|v| {
+                match v {
+                    Some(Value::String(s)) => Ok(s),
+                    _ => Err(CE::new(CEK::HeaderTypeError, None)),
```
matthiasbeyer: I'm not sure whether all required `use`es are done, but you can write it this way: `Err(CEK::HeaderTypeError.into_error())` if the right stuff is included.

libimagcounter/src/counter.rs
```diff
@@ -16,14 +16,15 @@ use error::CounterError as CE;
 use error::CounterErrorKind as CEK;
 
 pub type CounterName = String;
+pub type CounterUnit = String;
```
matthiasbeyer: I'm not sure on this one, actually. I'm not sure whether we want a type alias here or a new type.

I would rather make a new type: `pub struct CounterUnit(String);` and derive `Clone`, `Eq`, `PartialEq`, `Ord`, `PartialOrd` on it and implement `Display` and `Debug` on it myself.

Are you willing to do that?

imag-counter/src/create.rs
```diff
@@ -16,14 +16,17 @@ pub fn create(rt: &Runtime) {
...
                 .unwrap_or(0);
+            let unit = scmd
+                .value_of("unit")
+                .unwrap_or("unit");
 
-            match Counter::new(rt.store(), String::from(name), init) {
+            match Counter::new(rt.store(), String::from(name), init, String::from(unit)) {
                 Err(e) => {
-                    warn!("Could not create Counter '{}' with initial value '{}'", name, init);
+                    warn!("Could not create Counter '{}' with initial value '{}{}'", name, init, unit);
```
matthiasbeyer: Please add a space between the init and the unit.

imag-counter/src/create.rs
```diff
@@ -16,14 +16,17 @@ pub fn create(rt: &Runtime) {
...
                     trace_error(&e);
                     exit(1);
                 },
-                Ok(_) => info!("Created Counter '{}' with initial value '{}'", name, init),
+                Ok(_) => info!("Created Counter '{}' with initial value '{}{}'", name, init, unit),
```
matthiasbeyer: Also here: please add a space between the init and the unit.

matthiasbeyer:
It is awesome to watch you learn, actually! Like... six weeks ago you would've asked me a ton of questions to implement this,... now you just do it. awesome! :tada:

libimagcounter/src/counter.rs
asuivelentine: What do you want to tell me?

matthiasbeyer: Try to do `Err(CEK::HeaderTypeError.into_error())`.

libimagcounter/src/counter.rs
```diff
@@ -3,20 +3,37 @@ use std::ops::DerefMut;
...
+#[derive(Clone, Eq, PartialEq, Ord, PartialOrd)]
+pub struct CounterUnit(String);
+
+impl Display for CounterUnit {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "({})", self.0)
+    }
+}
+
+impl Debug for CounterUnit {
```
matthiasbeyer: Would be enough to `derive(Debug)` this...

libimagcounter/src/counter.rs
```diff
@@ -130,10 +143,10 @@ impl<'a> Counter<'a> {
     pub fn unit(&self) -> Result<CounterUnit> {
         self.fle.get_header().read("counter.unit")
             .map_err(|e| CE::new(CEK::StoreWriteError, Some(Box::new(e))))
-            .and_then(|v| {
-                match v {
-                    Some(Value::String(s)) => Ok(s),
-                    _ => Err(CE::new(CEK::HeaderTypeError, None)),
+            .and_then(|u| {
+                match u {
```
matthiasbeyer: Neat! Now you can also omit the scope of the closure: `/*..*/.and_then(|u| match { /* ... */ })`

:smile:

libimagcounter/src/counter.rs
```diff
@@ -142,13 +136,13 @@ impl<'a> Counter<'a> {
...
-            .map_err(|e| CE::new(CEK::StoreWriteError, Some(Box::new(e))))
-            .and_then(|u| {
+            .map_err(|e| CEK::StoreWriteError.into_error_with_cause(Box::new(e)))
+            .and_then(|u| 
                 match u {
                     Some(Value::String(s)) => Ok(CounterUnit(s)),
                     _ => Err(CEK::HeaderTypeError.into_error())
                 }
-            })
+            )
```
matthiasbeyer: You can remove two more lines by putting the `}` and the `)` on the same line and if you put the `match` right after the `|u|` (and remove also some indentation).

libimagcounter/src/counter.rs
```diff
@@ -21,7 +21,7 @@ use error::CounterErrorKind as CEK;
 pub type CounterName = String;
 
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
-pub struct CounterUnit(String);
+pub struct CounterUnit(pub String);
```
matthiasbeyer: Nope, don't make that `pub`.

Implement a `new()` function for the type.

imag-counter/src/create.rs
```diff
@@ -20,7 +21,7 @@ pub fn create(rt: &Runtime) {
                 .value_of("unit")
                 .unwrap_or("unit");
 
-            match Counter::new(rt.store(), String::from(name), init, String::from(unit)) {
+            match Counter::new(rt.store(), String::from(name), init, CounterUnit(String::from(unit))) {
```
matthiasbeyer: This line also gets shorter with a `new()` function for the type.

imag-counter/src/create.rs
```diff
@@ -16,14 +17,18 @@ pub fn create(rt: &Runtime) {
                 .value_of("initval")
                 .and_then(|i| FromStr::from_str(i).ok())
                 .unwrap_or(0);
+            let unit = scmd
+                .value_of("unit")
+                .unwrap_or("unit");
```
matthiasbeyer: You use the unit `"unit"` if there is no unit? How about passing an `Option<Unit>` to the `Counter::new()` call instead?

matthiasbeyer: Or even better: Initialize the counter without a unit (`None`) and add a function in the `Counter` which adds one: `Counter::with_unit(self, Option<Unit>) -> Counter`... so we can do `Counter::new().with_unit(unit)` in the calling code.

asuivelentine: In my opinion the value cant live without a unit? If there's no unit, the defaultunit - "unit(s)" should be used.

Imagine a counter with value 1... I would think: "1 - what? Apples?"

matthiasbeyer: Yes, exactly. And that's what an `Option<Unit>` would model.

asuivelentine: So you want that CounterUnit gets a member of counter? actually we dont need any type for it? cauese it's just represented by the string in the headerfile. Why shouldn't we handle it in this way?

matthiasbeyer: Well, we need a `Option<CounterUnit>` member in the `Counter` type. We do the `pub struct CounterUnit(String);` thing to be able to convert the type later to a full-blown struct, which may have several members and so on.

It's all about beeing forward-compatible, really!

libimagcounter/src/counter.rs
```diff
@@ -75,7 +74,20 @@ impl<'a> Counter<'a> {
...
+    }
+
+    pub fn with_unit(mut self, unit: Option<CounterUnit>) -> Result<Counter<'a>> {
+        self.unit = unit;
+
+        if let Some(u) = self.unit.clone() {
+            let mut header = self.fle.deref_mut().get_header_mut();
+            let setres = header.set("counter.unit", Value::String(u.0));
+            if setres.is_err() {
+                return Err(CE::new(CEK::StoreWriteError, Some(Box::new(setres.unwrap_err()))));
```
matthiasbeyer: Make sure you `self.unit = None` before returning the error!

libimagcounter/src/counter.rs
```diff
@@ -140,21 +152,19 @@ impl<'a> Counter<'a> {
...
+    pub fn unit(&self) -> Option<CounterUnit> {
         self.fle.get_header().read("counter.unit")
-            .map_err(|e| CEK::StoreWriteError.into_error_with_cause(Box::new(e)))
-            .and_then(|u| match u {
-                Some(Value::String(s)) => Ok(CounterUnit(s)),
-                _ => Err(CEK::HeaderTypeError.into_error())
+            .ok()
+            .and_then(|s| match s {
+                Some(Value::String(s)) => Some(CounterUnit::new(s)),
+                _ => None,
```
matthiasbeyer: I'm not sure whether we want to return `self.unit` here or re-read it from the store.

matthiasbeyer: Well, yes... this implementation makes sense as you're using it in `from_storeid()`, actually.

So I guess I'm fine with that, though I'm not fine with hiding the error. Returning an `Option::None` over an `Result:Err` is not fine, please revert the code to return an `Result` in case of error and adapt the rest of the codebase for it.

matthiasbeyer: Maybe we can offer two functions here:

* `unit(&self) -> Option<&CounterUnit>` - returning the unit member, if any
* `read_unit(/*..*/) -> Result<CounterUnit>` - trying to read the store and loading the unit from there.

That would make much more sense and we would have a clear distinction!

matthiasbeyer: You could then even provide a helper function to get a `Result<Option<CounterUnit>>` from a `FileLockEntry` and use that in `read_unit()` _and_ in `from_storeid()`!

asuivelentine: Even with a wrapper I need both (unit, read_unit) functions..
Do you mean to create
- unit(&self) -> Option<&CounterUnit> => returning the member
- read_unit(/*..*/) -> Result<Option<CounterUnit>> => returning the store entry?

matthiasbeyer: > unit(&self) -> Option<&CounterUnit> => returning the member

Yes. A reference to it, actually.

> read_unit(/*..*/) -> Result<Option<CounterUnit>> => returning the store entry?

No. In this case we already have the `FileLockEntry` at hand, and we just need to read the header to get the counter unit entry. If that fails we return `Err(e)`, if there is none we return `Ok(None)` and if we can read it and everything is fine we return `Ok(Some(CounterUnit))`.

libimagcounter/src/counter.rs
```diff
@@ -152,12 +154,16 @@ impl<'a> Counter<'a> {
...
+
+    pub fn read_unit(&self) -> Result<Option<CounterUnit>> {
         self.fle.get_header().read("counter.unit")
-            .ok()
+            .map_err(|e| CEK::StoreReadError.into_error_with_cause(Box::new(e)))
             .and_then(|s| match s {
-                Some(Value::String(s)) => Some(CounterUnit::new(s)),
-                _ => None,
+                Some(Value::String(s)) => Ok(Some(CounterUnit::new(s))),
+                _ => Ok(None),
```
matthiasbeyer: Split it up here:

* In case of `Some(_)` we can return an error indicating that the header is malformed
* In case of `None` we can return `Ok(None)`

libimagcounter/src/counter.rs
```diff
@@ -152,12 +154,16 @@ impl<'a> Counter<'a> {
...
 
-    pub fn unit(&self) -> Option<CounterUnit> {
+    pub fn unit(&self) -> Option<&CounterUnit> {
+        self.unit.as_ref()
+    }
+
+    pub fn read_unit(&self) -> Result<Option<CounterUnit>> {
         self.fle.get_header().read("counter.unit")
-            .ok()
+            .map_err(|e| CEK::StoreReadError.into_error_with_cause(Box::new(e)))
```
matthiasbeyer: You can use `.map_err_into()` to make this less noisy.

libimagcounter/src/counter.rs
```diff
@@ -193,8 +199,12 @@ impl<'a> FromStoreId for Counter<'a> {
             Err(e) => Err(CE::new(CEK::StoreReadError, Some(Box::new(e)))),
             Ok(c)  => {
                 let mut counter = Counter { fle: c, unit: None };
-                counter.unit = counter.unit();
-                Ok(counter)
+                counter.read_unit()
+                    .map_err(|e| CE::new(CEK::StoreReadError, Some(Box::new(e))))
```
matthiasbeyer: Here you can also use `.map_err_into()`

imag-counter/src/create.rs
```diff
@@ -17,13 +18,17 @@ pub fn create(rt: &Runtime) {
                 .and_then(|i| FromStr::from_str(i).ok())
                 .unwrap_or(0);
 
-            match Counter::new(rt.store(), String::from(name), init) {
-                Err(e) => {
+            let unit = scmd
+                .value_of("unit")
+                .and_then(|u| Some(CounterUnit::new(u)))
```
matthiasbeyer: You can simply `.map(CounterUnit::new)` here.

asuivelentine:
I've played around with it. Real SW-tests do not exist.

imag-counter/src/create.rs
```diff
@@ -20,8 +20,7 @@ pub fn create(rt: &Runtime) {
 
             let unit = scmd
                 .value_of("unit")
-                .and_then(|u| Some(CounterUnit::new(u)))
-                .or(None);
+                .map(|unit| CounterUnit::new(unit));
```
matthiasbeyer: You can literally `map(CounterUnit::new)` here, not using a closure!

matthiasbeyer:
Playing around is good enough for now, I guess.

imag-counter/src/create.rs
asuivelentine: Woot, that's really awesome.

matthiasbeyer: :smile:

matthiasbeyer:
Now this looks good. What you should do now is: Cleaning up the history. There are a lot of commits that could be squashed.

After that (or before...) you should rebase on latest master, as this has merge conflicts.

matthiasbeyer:
Waiting on CI now.

PR #538, opened 2016-07-15
Cleanup `HookError` tracing logic.

jsirois:
Use `Result::or_else` instead of matching and no-oping the `Ok` case
manually and extract a helper to centralize repeated trace logic.


matthiasbeyer:
Yeah, I like that change!

PR #486, opened 2016-06-30
libimagstorestdhook/git

matthiasbeyer:
The #187 follow up.

Implementation of a git hook which automatically commits and so on. Basic functionality will be implemented here, more will follow in followup-prs.

PR #552, opened 2016-07-16
Remove "Sync" trait bound - we don't need to sync between threads here

matthiasbeyer:
(and we do not have the possibility to, actually).

Extracted from #486 

@TheNeikos can you review this ... I'm not that familiar with Send/Sync, so it would be better if you would have a close look...


TheNeikos:
lgtm

PR #557, opened 2016-07-16
Libimagstore/no implicit create

matthiasbeyer:
A neat feature to be able to set this via the config, IMHO.

### DIFF
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -3,6 +3,10 @@
 
 [store]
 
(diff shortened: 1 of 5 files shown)


matthiasbeyer:
Nice, travis fails because we do not have a `imagrc` file specified in our tests, so implicit store creation is now denied by these patches.

matthiasbeyer:
Waiting on #571 now.

PR #564, opened 2016-07-19
Fix flags for root command not working; Added in debug forwarding

Geemili:
Previously, flags like `--version` were ignored, and would always display the help. This fixes that and also adds in the ability to write `imag --debug foo`.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -158,15 +158,15 @@ fn main() {
         },
 
         s => {
-            let subcommand_args = &find_args(s)[..];
+            let mut subcommand_args = find_args(s);
             if is_debug {
-                subcommand_args.push("--debug");
+                subcommand_args.push(String::from("--debug"));
             }
             match Command::new(format!("imag-{}", s))
                 .stdin(Stdio::inherit())
                 .stdout(Stdio::inherit())
                 .stderr(Stdio::inherit())
-                .args(subcommand_args)
+                .args(&subcommand_args[..])
                 .spawn()
                 .and_then(|mut handle| handle.wait())
             {
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -166,7 +166,7 @@ fn main() {
         s => {
             let mut subcommand_args = find_args(s);
             if is_debug && subcommand_args.iter().find(is_debug_flag).is_none() {
-                subcommand_args.push(String::from(DBG_FLAG));
+                subcommand_args.insert(0, String::from(DBG_FLAG));
             }
             match Command::new(format!("imag-{}", s))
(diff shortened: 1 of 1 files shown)


bin/src/main.rs
```diff
@@ -150,11 +158,15 @@ fn main() {
         },
 
         s => {
+            let mut subcommand_args = find_args(s);
+            if is_debug && subcommand_args.iter().find(|x| *x == "--debug").is_none() {
+                subcommand_args.push(String::from("--debug"));
```
cyplo: Would it be helpful to have `--debug` string in one place, as a constant maybe ?

bin/src/main.rs
```diff
@@ -113,11 +117,15 @@ fn main() {
...
-            exit(0);
+        None    => match find_flag() {
+            Some(s) => s,
+            None => {
+                help(commands);
+                exit(0);
+            },
         },
     };
+    let is_debug = env::args().skip(1).filter(|x| x == "--debug").next().is_some();
```
cyplo: Seems to be using same lambda as line 162 - may it be worth extracting ?

bin/src/main.rs
matthiasbeyer: Yup, I like that. Maybe a global `DBG_FLAG` constant...

bin/src/main.rs
matthiasbeyer: If you look closely, it is not. Here you have `x ==`, on line 162 you have `*x ==`. I'm not sure we can extract them, but feel free to try!

matthiasbeyer:
Thanks a lot for your PR, it looks good so far (though I didn't test it yet).

bin/src/main.rs
matthiasbeyer: Oh, @Geemili wrote this PR... so maybe @Geemili can try to extract the lambda ...

@cyplo thanks for having a look at this! :smile: :+1:

matthiasbeyer:
Hmm... `imag --debug counter --rtp /tmp/ list` yields `error: Found argument '--debug' which wasn't expected, or isn't valid in this context` for me.

matthiasbeyer:
I guess it would be easier if we switch over to `clap` in the `imag` binary as well, and use primitives from `clap` to cover this whole thing. I really don't know why I didn't use clap in the first place... :smirk: :question: 

If you want to do this, I'd love that! But I guess it would need another PR and should be seperated from this one. I will open an issue, at least.

Geemili:
@matthiasbeyer You're welcome! And that error that you were getting should be fixed.

matthiasbeyer:
Thanks a lot for working on this!

PR #567, opened 2016-07-21
Remove dash from command listing

matthiasbeyer:
With this patch, we list "imag counter" instead of "imag-counter" in the
help output.


matthiasbeyer:
I will put this on hold until #564 is merged, as it might produce an conflict with #564 and I'd rather like to provide a good PR experience for @geemili rather than getting this merged quickly.

PR #570, opened 2016-07-24
Libimagrt/config override settings

### LINKED ISSUE #565: Ability to override config file settings via commandline
This would go into `libimagrt` and override the config object directly.

We should offer a commandline flag `--override-config`, which can be used to override the configuration parsed from the config file. This could be used several times, IMHO.

Something like `--override-config store.implicit-create=true` or something like that. I'm not sure whether we already have the infrastructure for that, but I think so.

Details have to be worked out, but I guess it may fail (hard) if there is no configuration there. So if you `--override-config foo.bar`, `foo.bar` has to be in the configuration file, you are not allowed to set configuration from the commandline.

matthiasbeyer:
PR on #557 to be able to override configuration via commandline.

Closes #565


matthiasbeyer:
Setting array values is not yet supported (as of setting whole arrays. toml-rs supports accessing indexes inside errors, though the docs lack an example - but it should be possible).

matthiasbeyer:
Now this PR could also go against master and then rebase #557 against master so we have these stuff in there.

Anyways, now it is done this way, filing a new PR (why doesn't github have a feature where you can change the branch the PR is filed against?) would be too much noice.

matthiasbeyer:
Arg, f***. I will resubmit rebased and stuff..

PR #593, opened 2016-07-28
Libimagnotification/init

### LINKED ISSUE #590: use `notify-rust` to write a `libimagnotification`
For the first step it would be nice to have a `libimagnotification` functionality to notify errors.

I think of an interface like this:

``` rust
struct ErrorNotification { /* ... */ }

ErrorNotification::new() -> ErrorNotification;
ErrorNotification::with_trace_depth(self, usize) -> ErrorNotification;
ErrorNotification::with_timeout(self, msec: usize) -> ErrorNotification;
ErrorNotification::notify(self, e: &Error) -> Result<(), NotificationError>;
```

And maybe some extensions to `Result` so we can

``` rust
let f : Result<_, Error> = foo();
f.notify(|| ErrorNotification::new().with_trace_depth(5).with_timeout(6000)).map(|_| ()).map_err(|e| ...)
```

So basically a function which is implemented on `Result` which takes the `Err()` (if there is one), calls the passed closure if there is an error and traces the error with it.
It _must_ yield the error itself, so we can build it into our error handling.

---

Maybe we can even include primitives to get stuff from the configuration - things like timeout length and trace depth for example.

---

I'm tagging this with high complexity as it requires some knowledge about our error handling infrastructure as well as a good taste for API design, IMHO.

---

This is clearly functionality which should only be used by the binaries.

matthiasbeyer:
Closes #590

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,12 @@ matrix:
 cache:
   cargo: true
 
+addons:
+  apt:
+    packages:
+    - libdbus-1-dev
+    - pkg-config
+
 script:
   - cargo build --all --all-features --verbose -j 1
   - cargo test  --all --all-features --verbose -j 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,6 +43,7 @@ members = [
     "lib/entry/libimagentrytag",
     "lib/entry/libimagentryview",
     "lib/etc/libimaginteraction",
+    "lib/etc/libimagnotification",
     "lib/etc/libimagtimeui",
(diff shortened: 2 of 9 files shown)


matthiasbeyer:
Note to self:

We could also include an extension to `Option<T>`... something like `notify_none(what, exp)` which does `debug!("{what} is None, but was expected to be Some({exp})")`.

or something like this...

matthiasbeyer:
The lib itself is ready, though not used by the codebase just yet.

Issue #505, opened 2016-07-05
UI Wrapper Framework around clap?


matthiasbeyer:
We have a lot of inconsistency in the UI of the binary crates. Maybe we could create a wrapper framework around clap which helps with creating the UI? At least for the CRUD actions most CLI interfaces offer?

Would name it libimagcli. Maybe this could also go into libimagutil, would make sense IMHO.

matthiasbeyer:
A fix for this would close #493 as well.

asuivelentine:
I'll try this

matthiasbeyer:
Considered abandoned.

PR #596, opened 2016-07-30
Libimagcliutil/init

asuivelentine:
regarding #505 
Could you have a short look?


matthiasbeyer:
This looks indeed like I wanted it to look! Awesome work here! :+1: 

I will now start adding some annotations! Just minor stuff, though!

libimagcliutil/Cargo.toml
```diff
@@ -0,0 +1,7 @@
+[package]
+name = "libimagcliutil"
+version = "0.1.0"
```
matthiasbeyer: Use version `0.2.0` here. We use the same version for all crates in the distribution until we have a major release... it is more sane IMHO :-)

libimagcliutil/Cargo.toml
```diff
@@ -0,0 +1,7 @@
...
+authors = ["Kai Sickeler <k.sickeler@gmail.com>"]
+
+[dependencies]
+clap = "2.1.1"
```
matthiasbeyer: Please add a newline at the end of this file.

libimagcliutil/src/argbuilder.rs
```diff
@@ -0,0 +1,117 @@
+use std::option::Option;
```
matthiasbeyer: I guess this is in the prelude, so you do not need this.

libimagcliutil/src/create.rs
```diff
@@ -0,0 +1,15 @@
+use argbuilder::ArgBuilder;
+
+pub struct CreateArgBuilder;
```
matthiasbeyer: You want to wrap an `ArgBuilder` here: `pub struct CreateArgBuilder(ArgBuilder);` because of the following.

libimagcliutil/src/create.rs
```diff
@@ -0,0 +1,15 @@
...
+
+impl<'a> CreateArgBuilder {
+    pub fn new() -> ArgBuilder<'a> {
```
matthiasbeyer: You want to return a `CreateArgBuilder` here.

libimagcliutil/src/create.rs
```diff
@@ -0,0 +1,15 @@
...
+        ArgBuilder::new("create")
+            .with_short("c")
+            .with_long("create")
+            .with_helptext("creates a new element with name")
+            .with_takes_value(true)
+            .with_required(false)
+            .with_value_name("CREATE")
+    } 
+}
```
matthiasbeyer: Please add a newline at the end of the file.

Also Implement `Deref` and `DerefMut` (from `std::ops`) on the `CreateArgBuilder`.

matthiasbeyer: The same goes for all other arg builders, of course.

libimagcliutil/src/lib.rs
```diff
@@ -0,0 +1,28 @@
...
+pub mod create;
+pub mod delete;
+pub mod edit;
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn it_works() {
+    }
+}
```
matthiasbeyer: You can remove the `tests` module. Make sure you keep a blank line at the end of the file.

matthiasbeyer:
As said, this looks awesome.

I will think about which specific implementations are needed and which are not. :+1:

matthiasbeyer:
So from what I see we need:

* `--name`
* `--add`
* `--remove`
* `--edit` and `--no-edit`

and maybe more obscure ones:

* `--yes`
* `--select`
* `--range` or `--date-range` or `--time-range` (or all three) could be helpful as well

---

What I think we could also need is even more abstraction around this: A `SubCommandBuilder`!

When you look at
https://github.com/matthiasbeyer/imag/blob/master/imag-notes/src/ui.rs#L7,
https://github.com/matthiasbeyer/imag/blob/master/imag-notes/src/ui.rs#L25,
https://github.com/matthiasbeyer/imag/blob/master/imag-notes/src/ui.rs#L36,
https://github.com/matthiasbeyer/imag/blob/master/imag-counter/src/ui.rs#L35,
https://github.com/matthiasbeyer/imag/blob/master/imag-counter/src/ui.rs#L60,
https://github.com/matthiasbeyer/imag/blob/master/imag-counter/src/ui.rs#L71,
https://github.com/matthiasbeyer/imag/blob/master/imag-diary/src/ui.rs#L12,
https://github.com/matthiasbeyer/imag/blob/master/imag-diary/src/ui.rs#L51,
https://github.com/matthiasbeyer/imag/blob/master/imag-diary/src/ui.rs#L68,
https://github.com/matthiasbeyer/imag/blob/master/imag-diary/src/ui.rs#L95,
https://github.com/matthiasbeyer/imag/blob/master/imag-bookmark/src/ui.rs#L7,
https://github.com/matthiasbeyer/imag/blob/master/imag-bookmark/src/ui.rs#L29 and
https://github.com/matthiasbeyer/imag/blob/master/imag-bookmark/src/ui.rs#L63,
you can clearly see that we need an abstraction over `clap::SubCommand` even more than over the `Arg` type.

So it would be awesome if you could build the very same thing for `SubCommand`!

Make sure you only implement `create`, `delete`, `edit` and `remove` (CRED if you want :smile: ) in this PR. The `list` and `view` implementations must go to `libimagentrylist` and `libimagentryview` respectively and should be added in follow-up PRs, IMHO.

---

Thanks for your awesome work on this! I hope you want to continue implementing what I've proposed above. If not, feel free to ping me so I can do it in a seperate PR.

You are awesome! :heart:

matthiasbeyer:
Note to self: Propose code duplication removal through macro usage. Might need to teach @asuivelentine for it! :smile:

asuivelentine:
yeah, a Marco would be great for this

asuivelentine:
Not sure if the should create a trait for build. Then we can use it in subCommandBuilder.
```
fn with_arg<T: Argbuild>(arg: T) -> SubCommandBuilder{ /*...*/}
```

matthiasbeyer:
So I have some more annotations, but I will sum them up here:

* We might talk about help texts, some of them are not that appropriate, IMHO
* We don't need value names all the time, sometimes they make no sense... (for example if there is no value...)
* We might not use `::new()` as constructor but implement `std::default::Default`. If you want to do this, you can remove the `::new()` function IMO.
* "short" arguments should _always_ be a single character. I spotted at least one builder where you had two characters

---

More things: We need functions to exctract these arguments from [the Matches](https://kbknapp.github.io/clap-rs/clap/struct.ArgMatches.html). I, personally, would implement them on the `FooArgBuilder` types, maybe even with a trait:

```rust
pub trait ValueFetcher<T> {
    fetch_value(&Self) -> Option<T>;
    arg_present(&Self) -> bool;
}
```

Let me explain the type signature: We impl this trait on a type, so we can `Type::fetch_value(arg_matches)`. The implementation decides what return type is chosen, depending on the `Type` of the arg builder.

For example for an `TimeRangeArgBuilder` we return an `(DateTime, DateTime)` or something like this, for a `SelectArgBuilder` we return a `String`... Note that not all arg builders implement this trait (If there is no value (for example for the `YesArgBuilder`) we cannot fetch something, of course. For the `YesArgBuilder` we have at least the `arg_present()` function.

I'm still not sure whether such a trait would be a good idea (because of the generic type parameter...). Maybe we could implement this functionality on the types directly, without a trait. Anyhow, if you like to experiment with this, make sure to organize your commits in a way so they can be easily reverted! :smile: 

---

Also make sure you don't forget the `SubcommandBuilder` stuff. The `SubcommandBuilder` stuff would almost look the same as the `ArgBuilder` stuff.

libimagcliutil/src/name.rs
```diff
@@ -19,15 +22,25 @@ impl<'b> DerefMut for NameArgBuilder<'b> {
...
-    } 
+    }
+}
+
+impl<'a> NameArgBuilder<'a> {
+    pub fn arg_present(&self, arg: ArgMatches<'a>) -> bool {
+        arg.is_present("name")
+    }
+
+    pub fn fetch_value(&self, arg: ArgMatches<'a>) -> String {
```
matthiasbeyer: Returning an `Option<&'a str>` would be much more sane, wouldn't it?

And the lifetime for `ArgMatches` does not have to be the same as for the `NameArgBuilder` here, does it?

libimagcliutil/src/create.rs
```diff
@@ -0,0 +1,42 @@
...
+
+impl<'a> Default for CreateArgBuilder<'a> {
+    fn default() -> CreateArgBuilder<'a> {
+        CreateArgBuilder(ArgBuilder::new("create")
+            .with_short("c")
+            .with_long("create")
+            .with_helptext("creates a new element with name")
+            .with_takes_value(true)
+            .with_required(false)
+            .with_value_name("CREATE"))
```
matthiasbeyer: The value name here is `"NAME"`, ... please check all other builders as well.

libimagcliutil/src/create.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+impl<'a> CreateArgBuilder<'a> {
+    pub fn arg_present(&self, arg: ArgMatches<'a>) -> bool {
+        arg.is_present("create")
+    }
+}
```
matthiasbeyer: We need to get the name from this one as well... please check all other builders as well.

libimagcliutil/src/subcommand/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+            }
+
+            impl<'b> DerefMut for $typename<'b> {
+                fn deref_mut<'a>(&'a mut self) -> &'a mut SubCommandBuilder<'b> {
+                    &mut self.0
+                }
+            }
+
+            impl<'a> Default for $typename<'a> {
+
```
matthiasbeyer: Please remove this blank line.

libimagcliutil/src/subcommandbuilder.rs
```diff
@@ -0,0 +1,110 @@
+use clap::{App, Arg, SubCommand};
+
+
+//add
+//delete
+//list
+//view
+//edit
+//
```
matthiasbeyer: Please remove this comment.

matthiasbeyer:
You could also implement `Into<>` for `Arg` and `SubCommand` and the builders for them.

matthiasbeyer:
Also: The `SubCommandBuilder` could also get some Arg builders, right?

I'm not sure about this one, but it could be useful to have `SubCommandBuilder::with_arg(ArgBuilder)` - you understand what I mean?

matthiasbeyer:
You could also use [the derive_builder crate](https://crates.io/crates/derive_builder) which was announced [here](https://users.rust-lang.org/t/announcing-derive-builder-which-generates-setter-methods-for-your-struct/6990) to implement the builder types.

(I just found this, so sorry for not mentioning it beforehand).

matthiasbeyer:
Due to #652, please rebase on lastest master.

matthiasbeyer:
@asuivelentine are you ready for review? :smile:

matthiasbeyer:
@asuivelentine hey, anything happening here?

matthiasbeyer:
I'm closing this because of no progress in about 3/4 year. Feel free to reopen something like this based on current master, and even reuse patches from this PR if you can.

Issue #475, opened 2016-06-18
imag-view: Support: view-in: editor


matthiasbeyer:
`imag-view` cannot show/view entries in the editor, yet.

We should clearly add that soonish.

---

Note: The entry should be _shown_ and there should be **NO** way to edit the entry. I guess we should simply copying it to `/tmp` (and offer a way to copy it to another TMP directory) and then run the editor on it.

PR #603, opened 2016-07-31
get default editor

asuivelentine:
#475 
Not sure what I should open? Do I get a path from FLE or do I have to use the content of it and send it to stdin, ....?
```
editor.arg(what).spawn()
````

### DIFF
--- a/imag-view/src/editor.rs
+++ b/imag-view/src/editor.rs
@@ -1,0 +1,24 @@
+use libimagstore::store::FileLockEntry;
+use libimagrt::runtime::Runtime;
+use libimagentryview::builtin::editor::EditorView;
+use libimagentryview::viewer::Viewer;
+use libimagentryview::error::ViewError as VE;
+
+pub struct Editor<'a> {
+    rt: &'a Runtime<'a>,
+    fle: &'a FileLockEntry<'a>,
+}
+
+impl<'a> Editor<'a> {
+    pub fn new(rt: &'a Runtime, fle: &'a FileLockEntry) -> Editor<'a> {
+        Editor{
+            rt: rt,
+            fle: fle,
+        }
+    }
+
+    pub fn show(self) -> Result<(), VE> {
+        EditorView::new(self.rt).view_entry(self.fle)
+    }
+}
+
--- a/imag-view/src/main.rs
+++ b/imag-view/src/main.rs
@@ -34,8 +34,10 @@ use libimagentryview::builtin::versions::VersionsViewer;
 use libimagentryview::viewer::Viewer;
 
 mod ui;
+mod editor;
 
 use ui::build_ui;
+use editor::Editor;
 
 fn main() {
     let rt = generate_runtime_setup( "imag-view",
@@ -75,7 +77,10 @@ fn main() {
         } else if scmd.is_present("view-in-browser") {
             warn!("Viewing in browser is currently not supported, switch to stdout");
         } else if scmd.is_present("view-in-texteditor") {
-            warn!("Viewing in texteditor is currently not supported, switch to stdout");
+            if let Err(e) = Editor::new(&rt, &entry).show() {
+                error!("Cannot view in editor: {}", e);
+                trace_error_exit(&e, 1);
+            }
         } else if scmd.is_present("view-in-custom") {
             warn!("Viewing in custom is currently not supported, switch to stdout");
         }

### CODE REVIEW

**Comment by matthiasbeyer** [comment]
You should get the contents of the `FLE` and put it into a temporary file (there's a crate for temporary files, I guess it is named `tempfile` or similar).

Make sure to also put the header in the tempfile if the user requests so.

Then call the editor with this file as argument.

---

We make it this way because I'm pretty sure that not all editors support stdin content.

**Comment by asuivelentine** [comment]
To launch a new session I need the default Terminal emulator. Any idea how to get this information? Should I use the TERM environment?

**Comment by asuivelentine** [comment]
libimagrt's function [`editor`](https://github.com/matthiasbeyer/imag/blob/master/libimagrt/src/runtime.rs#L284) returns a Command. I guess we should change this implementaiton. We need a command that starts a new terminal emulator with the editor as argument, don't we?
Maybe we could offer both implementations?

File: imag-view/src/editor.rs [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+            c.wait().unwrap_or_else(|e| {
+                error!("Error: couldn't open editor. Make sure $EDITOR is set: {}", e);
+                exit(1);
+            });
+        };
+
+        file.close().unwrap();
+    }
+}
+
```
matthiasbeyer: I guess this whole functionality could be implemented in `libimagentryview` and therefor also implement the `Viewer` trait... so we can reuse it nicely.

asuivelentine: Yeah, you're right #610.
I've opened a new PR to offer a loose binding between the modules.

asuivelentine:
Ping... if your fine with that errorhandling it's meta/ready

matthiasbeyer:
Will have a look soonish.

imag-view/src/main.rs
```diff
@@ -75,7 +77,10 @@ fn main() {
         } else if scmd.is_present("view-in-browser") {
             warn!("Viewing in browser is currently not supported, switch to stdout");
         } else if scmd.is_present("view-in-texteditor") {
-            warn!("Viewing in texteditor is currently not supported, switch to stdout");
+            if let Err(e) = Editor::new(&entry, &rt).show() {
+                error!("Cannot view in editor: {}", e);
```
matthiasbeyer: I would like you to use `trace_error_exit()` here. Tracing the error is always better than using `error!()`, as we get the whole error chain printed.

imag-view/src/editor.rs
```diff
@@ -0,0 +1,26 @@
...
+impl<'a> Editor<'a> {
+    pub fn new(fle: &'a FileLockEntry, rt: &'a Runtime) -> Editor<'a> {
+        Editor{
+            fle: fle,
+            rt: rt,
+        }
+    }
+
+    pub fn show(&mut self) -> Result<(), VE>{
```
matthiasbeyer: There's a missing whitespace before the opening bracket `{`.

matthiasbeyer: I guess this function can also be made consuming `self` - what do you think? I have no problem if you do not want to make it consuming, though.

imag-view/src/editor.rs
```diff
@@ -0,0 +1,26 @@
...
+        EditorView::new(self.rt).view_entry(self.fle.deref())
```
matthiasbeyer: I'm not sure you need to call `deref` here - it is normally called implicitely. Try to compile it without the call to `deref` and fixup your commit if it compiles! :smile:

imag-view/src/editor.rs
```diff
@@ -0,0 +1,26 @@
...
+use libimagentryview::viewer::Viewer;
+use libimagentryview::error::ViewError as VE;
+
+pub struct Editor<'a> {
+    fle: &'a FileLockEntry<'a>,
+    rt: &'a Runtime<'a>,
+}
+
+impl<'a> Editor<'a> {
+    pub fn new(fle: &'a FileLockEntry, rt: &'a Runtime) -> Editor<'a> {
```
matthiasbeyer: I would say context goes first, so please let the arguments switch places here.

matthiasbeyer:
Due to #652, please rebase on lastest master.

imag-view/src/editor.rs
asuivelentine: Nope, your right. It's more intuitive.

matthiasbeyer:
You consider this ready, am I right?

PR #606, opened 2016-07-31
Libimagcounter/filter

matthiasbeyer:
My approach to add filter utilities to `libimagcounter`, so we can use them in the commandline.

Not yet successful. :cry:


matthiasbeyer:
Closing this, as it is long forgotten and will not even merge anymore.

PR #610, opened 2016-08-01
Libimagentryview/editor

asuivelentine:


matthiasbeyer:
You should implement `Viewer` (https://github.com/matthiasbeyer/imag/blob/master/libimagentryview/src/viewer.rs) instead of providing an own function!

libimagentryview/src/builtin/editor.rs
```diff
@@ -0,0 +1,29 @@
...
+        EditorView {
+            rt: rt,
+        }
+    }
+}
+
+impl<'a> Viewer for EditorView<'a> {
+    fn view_entry(&self, e: &Entry) -> Result<()> {
+        let ref mut entry = e.to_str().clone().to_string();
+        edit_in_tmpfile(self.rt, entry)
```
matthiasbeyer: Please do not use `ref` but pass `&entry`.

libimagentryview/src/builtin/editor.rs
```diff
@@ -0,0 +1,29 @@
...
+            .map_err(|e| VE::new(VEK::ViewError, Some(Box::new(e))))
+    }
```
matthiasbeyer: Add a comment here that `entry` gets `drop()`ed, even if it was modified... because we only _view_ things here.

libimagentryview/src/builtin/mod.rs
```diff
@@ -1,3 +1,4 @@
 pub mod plain;
 pub mod stdout;
 pub mod versions;
+pub mod editor;
```
matthiasbeyer: Can you sort the list of modules please (and create a `--fixup` commit)?

libimagentryview/src/builtin/editor.rs
```diff
@@ -0,0 +1,29 @@
...
+use libimagrt::edit::edit_in_tmpfile;
+
+use viewer::Viewer;
+use result::Result;
+use error::ViewErrorKind as VEK;
+use error::ViewError as VE;
+
+pub struct EditorView<'a> {
+    rt: &'a Runtime<'a>,
+}
```
matthiasbeyer: We could define this as newtype, right?

matthiasbeyer:
Ready here?

PR #611, opened 2016-08-01
Libimagstore/drop wlocks early

matthiasbeyer:
This was a hard one, thanks @TheNeikos for debugging and suggestions.

@TheNeikos would be nice if you could have _really_ close look on the diff in this PR!

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -428,16 +428,18 @@ impl Store {
                 .map_err_into(SEK::RetrieveCallError)
         }
 
-        let entry = try!(self.entries
-            .write()
-            .map_err(|_| SE::new(SEK::LockPoisoned, None))
-            .and_then(|mut es| {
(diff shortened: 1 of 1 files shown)


libimagstore/src/store.rs
```diff
@@ -426,7 +428,7 @@ impl Store {
                 .map_err_into(SEK::RetrieveCallError)
         }
 
-        self.entries
+        let entry = try!(self.entries
```
TheNeikos: This might not be enough, could you do it like this:

```rust
try!({
   /**/
});
```

matthiasbeyer: You mean like in acf7107 ?

TheNeikos: Yes

matthiasbeyer:
@TheNeikos ping.

TheNeikos:
LGTM :+1:

PR #624, opened 2016-08-06
Add workspace

matthiasbeyer:
Use cargo workspaces to speed up the build.

This is a test whether this actually works (on travis and locally).


matthiasbeyer:
I doubt that it worked the way we intended.

matthiasbeyer:
If I'm not mistaken the workspaces should also be used when compiling one of the crates locally. Unfortunately I do not get a `/target` when building in `/imag-bookmark`, so I assume cargo does not yet have workspaces - if I understand @steveklabnik right in [this reddit](https://www.reddit.com/r/rust/comments/4ri6kv/cargo_how_can_i_use_cargo_workspaces_in_my_project/) workspaces will be included in cargo 0.12 - I'm running 0.11 on my machine by now.

matthiasbeyer:
From what I can see, the workspace does not work yet, does it?

steveklabnik:
I'm confused, because you're just adding a `Cargo.toml` now?

matthiasbeyer:
@steveklabnik isn't that all that's needed for the workspaces? If not, I misunderstood the documentation for it...

steveklabnik:
It has to be a package generally; so you need all the rest of the stuff that goes in a Cargo.toml. That's why it's failing to parse

matthiasbeyer:
Ah, okay. I will add that tomorrow or something like that, am too tired right now after my long trip home.

Thanks for your support on this one!

steveklabnik:
No problem :smile:

matthiasbeyer:
:disappointed:

matthiasbeyer:
AFAICS workspaces didn't work in the latest build.

matthiasbeyer:
@steveklabnik sorry to bother you again, but can you think of something I got wrong here or why it didn't work yet?

steveklabnik:
No idea.

matthiasbeyer:
@TheNeikos do you think I should just merge this and as soon there is workspace support it will work automatically?

TheNeikos:
I don't know

steveklabnik:
From the workspace RFC: https://github.com/rust-lang/rfcs/blob/master/text/1525-cargo-workspace.md

-------

Some examples of layouts that will require extra configuration, along with the configuration necessary, are:

*    Trees without any root crate

```text
    crate1/
      Cargo.toml
      src/
    crate2/
      Cargo.toml
      src/
    crate3/
      Cargo.toml
      src/
```

these crates can all join the same workspace via a Cargo.toml file at the root looking like:

```text
    [workspace]
    members = ["crate1", "crate2", "crate3"]
```

-------

So I think that's what you need?

neithernut:
That's [already present](https://github.com/matthiasbeyer/imag/blob/72b64b65e72a7f40ce2a738f19a5c3739c0a63ea/Cargo.toml#L18).

steveklabnik:
Ah! I hadn't looked at this PR in a while, sorry for the noise!

matthiasbeyer:
From what I see, the workspaces feature does not work, yet. For example, [this build (from this PR)](https://travis-ci.org/matthiasbeyer/imag/jobs/162752856) used 

```
$ rustc --version
rustc 1.12.0-beta.6 (d3eb33ef8 2016-09-23)
$ cargo --version
cargo 0.13.0-nightly (109cb7c 2016-08-19)
```

And if I `grep` for `libimagerror` I can see that it is build multiple times. Or maybe the output occurs just multiple times because `cargo` wants to tell me that it would need to build cargo multiple times,... not sure.

neithernut:
I just noticed something: in the `Makefile`, the manifest-path is provided via the command line. Hence, f we use the `Makefile` for testing, the toplevel `Cargo.toml` might never be parsed by `cargo`. Maybe the `Makefile` should be rewritten to use the `--package` parameter rather then supplying the manifest.

matthiasbeyer:
Does still not work on travis, as far as I can see.
I guess we will have to move it to the next release ...

matthiasbeyer:
This does not generate error messages in my build jobs anymore, so I'll merge it.

Issue #625, opened 2016-08-06
travis build enhancements


matthiasbeyer:
If we build `imag-todo` we only have to build `libimagtodo` for tests (`travis-cargo -q test`).

Also, we do not have benchmarks, so we could remove that (`travis-cargo -q bench`).

Maybe some other speed improvements could be added.

matthiasbeyer:
Maybe we should refactor the build setup (for travis) to only build the required things. For example, if we change `libimagstore`, surely all the things have to be build. But if we change `libimagbookmark`, we do not have to build the world.

We should create a build script that

1. Fetches all crates in the repository that are affeced by the change. We could do this with some cargo stuff that gets us the dependency tree
2. Build all these crates and run the tests for them

That would improve build times a lot, I guess.

neithernut:
Personally, at this point I'd begin splitting things into sub-repositories, each with it's own build setup.

matthiasbeyer:
> Personally, at this point I'd begin splitting things into sub-repositories, each with it's own build setup.

Nope, I do not like that idea. This just creates overhead in multi-crate changes. After a `1.*` release, maybe. But not yet.

matthiasbeyer:
I guess I can close this, we have better build times already (30-40 min).

PR #630, opened 2016-08-06
Travis build only needed

matthiasbeyer:
Targeting and hopefully closing #625


matthiasbeyer:
Hm. So I'm not sure whether this actually worked the way we intended. I guess I will do the following: Squash this branch, make the history nice. Then I will squash it again into one TMP commit and randomly apply this commit to other branches to see whether they build correctly. (and of course revert these patches later on again).

matthiasbeyer:
@TheNeikos and @neithernut and maybe even @manuelmessner - could you have a close look? I hope this is correct... and yes, it is my bad bash stlye, but it ~~works~~ runs (as reported by travis)! :smile:

.travis_build_script_functions.sh
```diff
@@ -0,0 +1,124 @@
...
+    dbg "Getting metadata for $1, grepping $2"
+    cargo metadata --manifest-path "$1" | grep "\"$2\"" 2>/dev/null
+}
+
+# Gets all pathes to Cargo.toml files in changed directories in $*
+get_cargo_tomls_for_changed_dirs() {
+    dbg "Finding Cargo.toml in $*"
+    for d in $*; do
+        find "$d" -name "Cargo.toml" -type f
+    done
```
neithernut: You can pass `$*` to find directly

.travis_build_script.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+
+source ./.travis_build_script_functions.sh
+
+for changed_dir in $(get_changed_dirs); do
+    toml=$(get_cargo_tomls_for_changed_dirs $changed_dir);
+    tomls=$(reverse_dep_recursive $toml)
+    echo "$tomls"
+done
```
neithernut: This loops does anything else but produce output?

matthiasbeyer: Ah, nope. I will remove it.

neithernut:
Looks like you have dead code in `.travis_build_script.sh`.

matthiasbeyer:
> Looks like you have dead code in .travis_build_script.sh.

Why don't you annotate the dead code (I fail to see) - so I know where it is? :smile:

neithernut:
Mh... you are right. They are pulled in by `get_to_build_tomls()`.

matthiasbeyer:
It seems like I fu**** up here. I will close this unmerged - we already have a shorter build time,... should be enough for now.

PR #631, opened 2016-08-06
Libimagstore/remove storeid from pathbuf

matthiasbeyer:
This is an experiement.

I want to remove complexity from the `StoreId` type to see what builds break ... this could simplify the changes @TheNeikos wants to make.


matthiasbeyer:
Even after some changes in the rest of the store, this seems to be the wrong approach. I'll stop travis and close this PR, the latest commits were only pushed for documentation purposes.

PR #635, opened 2016-08-06
Use clap in bin/imag binary.

### LINKED ISSUE #566: Switch to `clap` in `imag` binary
We could and should use `clap` in the `imag` binary as well.

This was suggested by @mario-kr yesterday as well, if I remember correctly. At least I just posted something in #564 that this would be a good idea.

matthiasbeyer:
Closes #566 

Not tested yet.

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -5,6 +5,7 @@ extern crate clap;
 extern crate walkdir;
 
 extern crate libimagrt;
+extern crate libimagerror;
 
 use std::env;
 use std::process::exit;
@@ -14,10 +15,10 @@ use std::io::ErrorKind;
 
 use walkdir::WalkDir;
 use crossbeam::*;
-use clap::{Arg, App, AppSettings, SubCommand};
+use clap::{Arg, AppSettings, SubCommand};
 
 use libimagrt::runtime::Runtime;
-use libimagrt::setup::generate_runtime_setup;
+use libimagerror::trace::trace_error;
 
 fn help_text(cmds: Vec<String>) -> String {
     let text = format!(r#"
@@ -105,8 +106,14 @@ fn get_commands() -> Vec<String> {
     execs
 }
 
-pub fn build_ui<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
-    app
+
+fn main() {
+    let appname  = "imag";
+    let version  = &version!();
+    let about    = "imag - the PIM suite for the commandline";
+    let commands = get_commands();
+    let helptext = help_text(commands);
+    let app      = Runtime::get_default_cli_builder(appname, version, about)
         .settings(&[AppSettings::AllowExternalSubcommands])
         .arg(Arg::with_name("version")
              .long("version")
@@ -121,15 +128,14 @@ pub fn build_ui<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
              .multiple(false)
              .help("Get the versions of the imag commands"))
         .subcommand(SubCommand::with_name("help").help("Show help"))
-        .help(help_text(get_commands()))
-}
-
-fn main() {
-    let appname  = "imag";
-    let version  = &version!();
-    let about    = "imag - the PIM suite for the commandline";
-    let rt       = generate_runtime_setup(appname, version, about, build_ui);
-    let matches  = rt.cli();
+        .help(helptext.as_str());
+    let rt = Runtime::new(app)
+        .unwrap_or_else(|e| {
+            println!("Runtime couldn't be setup. Exiting");
+            trace_error(&e);
+            exit(1);
+        });
+    let matches = rt.cli();
 
     debug!("matches: {:?}", matches);
 
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -186,6 +186,8 @@ fn main() {
                     });
             }
 
+            debug!("Calling 'imag-{}' with args: {:?}", subcommand, subcommand_args);
+
             match Command::new(format!("imag-{}", subcommand))
                 .stdin(Stdio::inherit())
                 .stdout(Stdio::inherit())
@@ -196,12 +198,15 @@ fn main() {
             {
                 Ok(exit_status) => {
                     if !exit_status.success() {
+                        debug!("{} exited with non-zero exit code: {:?}", subcommand, exit_status);
                         println!("{} exited with non-zero exit code", subcommand);
                         exit(exit_status.code().unwrap_or(42));
                     }
+                    debug!("Successful exit!");
                 },
 
                 Err(e) => {
+                    debug!("Error calling the subcommand");
                     match e.kind() {
                         ErrorKind::NotFound => {
                             println!("No such command: 'imag-{}'", subcommand);
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -13,7 +13,7 @@ use std::io::ErrorKind;
 
 use walkdir::WalkDir;
 use crossbeam::*;
-use clap::{Arg, SubCommand};
+use clap::{Arg, AppSettings, SubCommand};
 
 use libimagrt::runtime::Runtime;
 
@@ -129,6 +129,8 @@ fn main() {
              .required(false)
              .multiple(false)
              .help("Show help"))
+        .subcommand(SubCommand::with_name("help").help("Show help"))
+        .settings(&[AppSettings::AllowExternalSubcommands])
         .get_matches();
 
     if matches.is_present("help") {
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -125,9 +125,22 @@ fn main() {
                       .takes_value(false)
                       .required(false)
                       .multiple(false)
-                      .help("Get the versions of the imag commands"));
+                      .help("Get the versions of the imag commands"))
+                 .arg(Arg::with_name("help")
+                      .long("help")
+                      .short("h")
+                      .takes_value(false)
+                      .required(false)
+                      .multiple(false)
+                      .help("Show help"));
+
 
     let matches = app.get_matches();
+
+    if matches.is_present("help") {
+        help(get_commands());
+        exit(0);
+    }
 
     if matches.is_present("version") {
         println!("imag {}", &version!()[..]);
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -126,10 +126,6 @@ pub fn build_ui<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
              .multiple(false)
              .help("Show help"))
         .subcommand(SubCommand::with_name("help").help("Show help"))
-        .subcommands(get_commands()
-            .iter()
-            .map(|cmd| SubCommand::with_name(cmd))
-        )
 }
 
 fn main() {
@@ -177,7 +173,7 @@ fn main() {
 
     match matches.subcommand() {
         (subcommand, Some(scmd)) => {
-            let subcommand_args : Vec<&str> = scmd.values_of(subcommand).unwrap().collect();
+            let subcommand_args : Vec<&str> = scmd.values_of("").unwrap().collect();
             debug!("Calling 'imag-{}' with args: {:?}", subcommand, subcommand_args);
 
             match Command::new(format!("imag-{}", subcommand))
@@ -202,6 +198,7 @@ fn main() {
                     match e.kind() {
                         ErrorKind::NotFound => {
                             println!("No such command: 'imag-{}'", subcommand);
+                            println!("See 'imag --help' for available subcommands");
                             exit(2);
                         },
                         ErrorKind::PermissionDenied => {
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -14,9 +14,10 @@ use std::io::ErrorKind;
 
 use walkdir::WalkDir;
 use crossbeam::*;
-use clap::{Arg, AppSettings, SubCommand};
+use clap::{Arg, App, AppSettings, SubCommand};
 
 use libimagrt::runtime::Runtime;
+use libimagrt::setup::generate_runtime_setup;
 
 fn help(cmds: Vec<String>) {
     println!(r#"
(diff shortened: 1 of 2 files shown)


asuivelentine:
Could you please change the Clap dependency to 2.*

matthiasbeyer:
As soon as your PR is merged, I will rebase and fix this one.

matthiasbeyer:
Waiting for #636

matthiasbeyer:
Does not work yet :cry:

matthiasbeyer:
@kbknapp Could you lead me in this one? What I try: Find all `imag-*` binaries in `$PATH` and add them as subcommands. If `imag foo` gets called (so `foo` is the subcommand), I `exec()` (via `std::process::Command`) the `imag-foo` binary.

Somehow clap calls the help for the `foo` subcommand if I call `imag foo` on the commandline... I'm not sure what gets into my way here... maybe you can give me some pointers?

matthiasbeyer:
I'm pretty sure my whole approach here is buggy... maybe we can do this in another way...

matthiasbeyer:
@kbknapp As there was no reaction from you, I ping you again one more time... do you have any suggestions how this could be implemented with clap?

kbknapp:
@matthiasbeyer appologies! This got lost in my notifications page! I'm looking over it now and will let you know!

kbknapp:
@matthiasbeyer I have a quick question; are you trying to add these external subcommands (`imag-*` binaries) *prior to* them being used, so that they appear inside the `imag` help output? 

If *not* what you can do is tell the `imag` binary that [external subcommands are allowed](http://kbknapp.github.io/clap-rs/clap/enum.AppSettings.html#variant.AllowExternalSubcommands), and then if clap determines one was potentially used, at that point check `$PATH` for the existince of the specific subcommand, i.e. if `clap` says the external subcommand was `bar` search for `imag-bar` only. If found, use `process::Command` to execute `imag-bar` and pass all the remaining arguments from `clap` to this `imag-bar` binary.

If that's what you're already doing, could you point me to where the problem lies because I'm misunderstanding it otherwise :smile:

kbknapp:
The only other thing you'll have to do is if you search `$PATH` for `imag-bar` and it *isn't* found, you'll have to exit and inform the user of their possible error. There are some ways to build a `ClapError` and call `exit()` on that to display consistent error messages as well, but they weren't quite designed to be used like that so it may take some massaging.

matthiasbeyer:
@kbknapp The problem (as far as I can remember, havn't worked on this in a couple of days) was also that if I call

```bash
imag validmodule validsubcommand --validarg
```

clap yields an error because `validsubcommand` is not defined in the setup for the `imag` binary but only in the setup for the `imag-validmodule` binary.

kbknapp:
Yep, that's the part that `AppSettings::AllowExternalSubcommands` should fix. It will use `validsubcommand` as the potential "external subcommand" and store `--validarg` to be passed on to the `imag-validsubcommand` binary.

kbknapp:
It will still be on you to call and pass the remaining args to the external binary, but what that setting does is supress that error and store the relevant args for you to make passing of the data easy(ier).

matthiasbeyer:
Hm... I'm not sure. If I do the following:

```bash
# build imag and imag-notes, I did it manually and copied them to /out,
# but it should be possible with
make bin imag-notes

cd out
PATH=$PATH:$(pwd) ./imag notes --help 
# get the help text from 'imag' but not from 'imag-notes' here

# also:
PATH=$PATH:$(pwd) ./imag help # prints the clap-helptext, despite I included a `help` subcommand.
PATH=$PATH:$(pwd) ./imag --help # prints the imag-helptext
```

I'll investigate further this weekend.

kbknapp:
I'll look at this further too, I can't actually clone the repo and test anything until I get home tonight though.

matthiasbeyer:
@kbknapp I do not want to annoy you,... still, any progress? :smiley:

mario-kr:
Whatever it is you did in those two comments, it works less than before:
With imag-todo, I got this output:
```
> imag todo
error: Found argument 'todo' which wasn't expected, or isn't valid in this context

USAGE:
    imag-todo [FLAGS] [OPTIONS] [SUBCOMMAND]

For more information try --help
todo exited with non-zero exit code
```
imag-todo gets called, but with the subcommand `todo` (Basically `imag-todo todo`). Did you forget to ignore the first argument? If I find it, I'll make a comment in the code.

bin/src/main.rs
```diff
@@ -97,100 +103,113 @@ fn get_commands() -> Vec<String> {
...
-        s => {
-            let mut subcommand_args = find_args(s);
-            if is_debug && subcommand_args.iter().find(is_debug_flag).is_none() {
-                subcommand_args.insert(0, String::from(DBG_FLAG));
+    matches.subcommand_name()
+        .map(|subcommand| {
+
+            let mut subcommand_args = vec![];
+
+            for arg in Runtime::arg_names() {
```
mario-kr: maybe here?

bin/src/main.rs
```diff
@@ -97,100 +103,113 @@ fn get_commands() -> Vec<String> {
...
+                        subcommand_args.push(value);
+                    });
             }
-            match Command::new(format!("imag-{}", s))
+
+            match Command::new(format!("imag-{}", subcommand))
                 .stdin(Stdio::inherit())
                 .stdout(Stdio::inherit())
                 .stderr(Stdio::inherit())
+                .arg(subcommand)
```
mario-kr: or here...

bin/src/main.rs
matthiasbeyer: Seems so yes.

mario-kr:
I did some testing (just a bit):
```bin> ./target/debug/imag counter help
error: Found argument 'help' which wasn't expected, or isn't valid in this context

USAGE:
    imag counter

For more information try --help
```
Basically, with the current setup, all commands are parsed, and then checked by clap, including the ones meant for the external command. clap doesnt know them, and you have an error.
I dont know what exactly was different before you used `generate_runtime_setup`, but i consider it possible that you'll have to revert to using clap directly.

matthiasbeyer:
@mario-kr according to @kbknapp the idea I want to implement is possible with clap. I'm just not there yet, as it seems. And unfortunately I do not know how to get there... :cry:

mario-kr:
I dont question clap. I question the usage of generate_runtime_setup, as it worked (at least kind of) before

matthiasbeyer:
> I question the usage of generate_runtime_setup

Nah, it is better to use this function. It does just abstract away a few pieces, as you can see [here](https://github.com/matthiasbeyer/imag/blob/master/libimagrt/src/setup.rs#L13) - it is basically the same as before, but we use a builder-function (`build_ui`) to build the commandline interface. And it also `exit()`s for us if the `Runtime` object cannot be built.

I doubt that the change from not using this function to using it messed things up for you.

mario-kr:
> I doubt that the change from not using this function to using it messed things up for you.
But it certainly looks that way. I'll read over the last 4-5 comments, but I dont think there anything else which could explain that.

bin/src/main.rs
```diff
@@ -102,15 +103,10 @@ fn get_commands() -> Vec<String> {
...
 
-fn main() {
-    let appname  = "imag";
-    let version  = &version!();
-    let about    = "imag - the PIM suite for the commandline";
-    let commands = get_commands();
-    let r        = Runtime::get_default_cli_builder(appname, version, about);
-    let matches  = commands
+pub fn build_ui<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
+    get_commands()
```
mario-kr: The order of subcommands is different, in the new version you're adding the imag-* stuff before help, version etc.
Though i dont actually think this is the error, the order *shouldnt* matter.

bin/src/main.rs
```diff
@@ -97,78 +103,92 @@ fn get_commands() -> Vec<String> {
...
-            for versionstring in result.into_iter().map(|handle| handle.join()) {
-                println!("{}", versionstring);
-            }
-        },
+        for versionstring in result.into_iter().map(|handle| handle.join()) {
+            println!("{}", versionstring);
+        }
+    }
+
+    matches.subcommand_name()
```
mario-kr: You could try this construct:
```
match m.subcommand() {
    (external, Some(ext_m)) => {
         let ext_args: Vec<&str> = ext_m.values_of(external).unwrap().collect();
    },
    _ => unreachable!()
}
```
As described by @kbknapp in [Allow unrecognized subcommands](https://github.com/kbknapp/clap-rs/issues/372).
It looks like you might save some loc with it.

matthiasbeyer: I'm not sure whether this would work. If you want, you can provide patches for this PR!

bin/src/main.rs
```diff
@@ -132,21 +128,32 @@ fn main() {
...
         .subcommand(SubCommand::with_name("help").help("Show help"))
         .settings(&[AppSettings::AllowExternalSubcommands])
-        .get_matches();
+}
+
+fn main() {
+    let appname  = "imag";
+    let version  = &version!();
+    let about    = "imag - the PIM suite for the commandline";
+    let rt       = generate_runtime_setup(appname, version, about, build_ui);
```
mario-kr: I did some (dumb) tests with `println!()` before and after `generate_runtime_setup`, and guess what, Only the output `before: generate_runtime_setup` was given. So either it actually makes a difference, or clap just started to get hungry and does the parsing, as well as the error printing and exiting, before we even get a chance to spawn our commands. I think it's the first one.
Is there an easy way to go back to a specific commit, build and test, and then go back to the current one? If yes, can you show me? I'd like to make sure the cause was the introduction of `generate_runtime_setup`

matthiasbeyer: > or clap just started to get hungry and does the parsing, as well as the error printing and exiting, before we even get a chance to spawn our commands

Yes. If you call a non-valid command, the clap parsing will print the help/error message and call `exit()`.

mario-kr:
By now, it should be OK to merge. But additional tests are always welcome!

matthiasbeyer:
@mario-kr feel free to write some basic bash tests for this branch.

matthiasbeyer:
@mario-kr for me it doesn't work - example:

```bash
PATH=$PATH:$(pwd)/imag-notes/target/debug/ ./bin/target/debug/imag --debug notes --help
[imag][WARN]: No config file found.
[imag][WARN]: Continuing without configuration file
Config: None
Store-config: None
[imag][DEBUG][/home/m/kernel/imag/libimagstore/src/store.rs][214]: Validating Store configuration
[imag][DEBUG][/home/m/kernel/imag/libimagstore/src/store.rs][219]: Building new Store object
[imag][WARN]: Implicitely creating store directory is denied
[imag][WARN]:  -> Either because configuration does not allow it
[imag][WARN]:  -> or because there is no configuration
Runtime couldn't be setup. Exiting

ERROR[   1]: Creating store directory implicitely denied
ERROR[   2]: File Error -- caused:
ERROR[   3]: IO Error -- caused:
ERROR[   4]: Could not instantiate
```

with config passed:

```bash
PATH=$PATH:$(pwd)/imag-notes/target/debug/ ./bin/target/debug/imag --debug --versions --config ./imagrc.toml 
[imag][DEBUG][/home/m/kernel/imag/libimagrt/src/configuration.rs][60]: Building configuration
[imag][DEBUG][/home/m/kernel/imag/libimagrt/src/configuration.rs][61]:   - verbosity  : false
[imag][DEBUG][/home/m/kernel/imag/libimagrt/src/configuration.rs][62]:   - editor     : None
[imag][DEBUG][/home/m/kernel/imag/libimagrt/src/configuration.rs][63]:   - editor-opts: 
Config: Some(Configuration { config: Table({"store": Table({"aspects": Table({"debug": Table({"mutable_hooks": Boolean(true), "parallel": Boolean(false)})}), "hooks": Table({"stdhook_debug": Table({"aspect": String("debug")})}), "implicit-create": Boolean(false), "post-create-hook-aspects": Array([String("debug")]), "post-delete-hook-aspects": Array([String("debug")]), "post-move-hook-aspects": Array([String("debug")]), "post-retrieve-hook-aspects": Array([String("debug")]), "post-update-hook-aspects": Array([String("debug")]), "pre-create-hook-aspects": Array([String("debug")]), "pre-delete-hook-aspects": Array([String("debug")]), "pre-move-hook-aspects": Array([String("debug")]), "pre-retrieve-hook-aspects": Array([String("debug")]), "pre-update-hook-aspects": Array([String("debug")]), "store-unload-hook-aspects": Array([String("debug")])})}), verbosity: false, editor: None, editor_opts: "" })
Store-config: Some(Table({"aspects": Table({"debug": Table({"mutable_hooks": Boolean(true), "parallel": Boolean(false)})}), "hooks": Table({"stdhook_debug": Table({"aspect": String("debug")})}), "implicit-create": Boolean(false), "post-create-hook-aspects": Array([String("debug")]), "post-delete-hook-aspects": Array([String("debug")]), "post-move-hook-aspects": Array([String("debug")]), "post-retrieve-hook-aspects": Array([String("debug")]), "post-update-hook-aspects": Array([String("debug")]), "pre-create-hook-aspects": Array([String("debug")]), "pre-delete-hook-aspects": Array([String("debug")]), "pre-move-hook-aspects": Array([String("debug")]), "pre-retrieve-hook-aspects": Array([String("debug")]), "pre-update-hook-aspects": Array([String("debug")]), "store-unload-hook-aspects": Array([String("debug")])}))
[imag][DEBUG][/home/m/kernel/imag/libimagstore/src/store.rs][214]: Validating Store configuration
[imag][DEBUG][/home/m/kernel/imag/libimagstore/src/store.rs][219]: Building new Store object
[imag][WARN]: Implicitely creating store directory is denied
[imag][WARN]:  -> Either because configuration does not allow it
[imag][WARN]:  -> or because there is no configuration
Runtime couldn't be setup. Exiting

ERROR[   1]: Creating store directory implicitely denied
ERROR[   2]: File Error -- caused:
ERROR[   3]: IO Error -- caused:
ERROR[   4]: Could not instantiate
```

mario-kr:
How about allowing it to implicetly create the store? We use `libimagrt::runtime::Runtime` now, so we need the store here.

matthiasbeyer:
You're right:

```bash
$ PATH=$PATH:$(pwd)/imag-notes/target/debug/ ./bin/target/debug/imag --debug --versions --config ./imagrc.toml --override-config store.implicit-create=true
``` 

works. I guess we can merge then, or do you want to test a bit more?

mario-kr:
I dont really get to writing tests until at least evening, so but i can put them in a follow-up PR. so yes, mergable.

matthiasbeyer:
Thanks! :smile:

PR #638, opened 2016-08-07
Cargo rustc codegen units

matthiasbeyer:
Lets see whether this speeds up our build.


matthiasbeyer:
[46 min/57 min](https://travis-ci.org/matthiasbeyer/imag/builds/150402282) does not look like that helped anything...

killercup:
I would try to set this using the `RUSTFLAGS` env variable. Also, the number (4) was just a guess on my side, I don't know how many cores Travis gives you.

There was some effort to make rust-lang/rust use an in-beta environment on travis to do much more checks. Maybe you can fine some useful information there: https://github.com/rust-lang/rust/pull/28500

matthiasbeyer:
So lets see whether this works. :smile:

matthiasbeyer:
Still no speedup from what I can see... Not sure whether the env variable thingy has worked, though.

killercup:
Hm, you'll see it with `cargo build --verbose` (i think).

Btw, did cargo workspaces land in stable? I think this will give you some
nice speed ups, but it landed pretty recently, and your nightly CI was
broken last I checked. So it might not have been observable.


Matthias Beyer <notifications@github.com> schrieb am So. 7. Aug. 2016 um
17:10:

matthiasbeyer:
> Btw, did cargo workspaces land in stable?

WIP in https://github.com/matthiasbeyer/imag/pull/624, doesn't seem to work either, though!

matthiasbeyer:
I'm not sure why the nightly build is broken, though. It seems like nightly cannot parse the Cargo.toml files - is this my fault? I don't think so...

killercup:
There were some problem with a cargo release a few days ago (but I don't
want to just blame everything on that 😄). Maybe try to pin the nightly to
something a little older like nightly-2016-07-20?


Matthias Beyer <notifications@github.com> schrieb am So. 7. Aug. 2016 um
17:31:

matthiasbeyer:
I don't think that helped, will close unmerged.

PR #639, opened 2016-08-07
travis: cache all target/ directories and $HOME/.cargo

matthiasbeyer:


matthiasbeyer:
@killercup did I get this right? Could you have a look?

killercup:
@matthiasbeyer, it's hard to say whether this works. But you can test it—just add another commit that changes one crate and see if cargo can use the cached results.

If you want to get workspaces working, you should also add `$TRAVIS_BUILD_DIR/target` as the workspace will probably use the parent directory.

matthiasbeyer:
I would say that it won't hurt to include that and as it is only a small patch we can revert it easily if something breaks. I will merge this if CI succeeds and we'll see...

matthiasbeyer:
I completely forgot about this one.

matthiasbeyer:
@mario-kr This should speed up the travis builds even more :smile:

mario-kr:
Hmm, no real benefits. `$HOME/.cargo` might be more important to cache, as all the 3rd party crates are in there. Though that would require a `cargo update` some times...

matthiasbeyer:
> Hmm, no real benefits. $HOME/.cargo might be more important to cache, as all the 3rd party crates are in there.

This is done by the `cargo: true` part.

PR #645, opened 2016-08-10
Libimagref/custom hash functions

matthiasbeyer:
Add custom hash function support for `libimagref`.


matthiasbeyer:
After that I can implement `libimagmail` with the `mailparse` crate.

Issue #647, opened 2016-08-12
Rewrite libimagentryfilter with filters crate


matthiasbeyer:
All I have to say: https://crates.io/crates/filters

I extracted this idea from imag (of course) and published it as crate. We can now use this crate to rewrite `libimagentryfilter` with it. And as far as I can see, we can reimplement it backwards-compatible.

Nevertheless a complexity/medium issue, I guess.

PR #648, opened 2016-08-13
libimagentryfilter: rewrite with filters crate

impo:
Starting to rewrite this crate using the `filters` crate as per #647.


impo:
Travis approves. Go ahead and look at the commits, see if it looks good.

matthiasbeyer:
Thanks a lot for your effort! (I'm on vacation now, so don't feel ignored :smile:)

PR #650, opened 2016-08-21
Try to fix build error on travis

matthiasbeyer:
Lets see what's breaking the build.


matthiasbeyer:
It seems that commit dcb31cb broke the build... though I do not see why that could be...

matthiasbeyer:
Hm. It seems that this wasn't the issue. I'm too tired right now to investigate, but will do so tomorrow.

matthiasbeyer:
So the build fails because of a compiler error and the fix for that is beeing backported to rust beta (said @steveklabnik on IRC in #rust-beginners) - so all we can do here is wait.

PR #651, opened 2016-08-22
Travis gh pages doc

matthiasbeyer:
We do not yet have a setup for github pages... this one implements it.

I'm not sure whether this succeeds, as we need `sudo` to `pip install` `ghp-import` - lets see how this can be fixed.


matthiasbeyer:
Lets see whether this worked... *merge*

PR #656, opened 2016-08-23
Cargo rustc codegen units

matthiasbeyer:
travis has two cores per container, so we can add this to improve performance, right?


matthiasbeyer:
From what I can see this didn't yield any performance gains:

* Build: https://travis-ci.org/matthiasbeyer/imag/builds/155887770
* Build of base (master): https://travis-ci.org/matthiasbeyer/imag/builds/155769956

If someone can come up with a change to this PR that gives us a better build time... feel free!

matthiasbeyer:
Okay, so at least this doesn't print error messages anymore, so I'll merge this now.

PR #657, opened 2016-08-23
Libimagmail/init

matthiasbeyer:
Mail module library. Heavily WIP.

### DIFF
--- a/libimagmail/Cargo.toml
+++ b/libimagmail/Cargo.toml
@@ -1,0 +1,6 @@
+[package]
+name = "libimagmail"
+version = "0.1.0"
+authors = ["Matthias Beyer <mail@beyermatthias.de>"]
+
+[dependencies]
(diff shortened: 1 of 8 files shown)


matthiasbeyer:
Features I want to implement:

Intermedia "Store" object which can hold `N` references to mails and then create links between all of them based on message-id, so we can "cache" the mail "threads".

This store should be loadable via `Store::retrieve_for_module()` and then filterable via `libimagfilter`.

This would be indeed a really nice use-case for `libimagfilter`/`filters`, as we barely want to load _all_ mails into RAM while processing them. Sometimes it makes more sense to only load mails from a certain subdirectory of the `Maildir`, for example... there can be a hell lot of filters here.

matthiasbeyer:
@TheNeikos could you have a quick look whether something looks nasty? If not I'd go the "hardcore" route and merge this, ...

TheNeikos:
I saw nothing wrong with it

PR #667, opened 2016-08-25
Rewrite storeid type

matthiasbeyer:
@TheNeikos Feel free to provide patches here

This might be a long-running one. It is intended to rewrite the `StoreId` type so we can rely on its properties.


matthiasbeyer:
@TheNeikos The very first commit in this series somehow slipped into the patchset when rebasing ... I guess I can remove it from the set, am I right?

TheNeikos:
Yes

matthiasbeyer:
I'll wait for #669 - after a rebase this might easier.

matthiasbeyer:
By now it looks pretty good in here, IMHO.

I try to fix the stuff so that it compiles - I'm pretty sure tests (both lib tests and binary tests) will fail after I got this stuff compiling...

matthiasbeyer:
I'm not sure whether I'm fine with https://github.com/matthiasbeyer/imag/pull/667/commits/5da270b9821dffa983f5f3cb07dc7cd7775716d7 (`impl IntoStoreId for PathBuf /* using StoreId::new_baseless() */`) though I guess we can pretty easily rip this one out later on.

matthiasbeyer:
Holy shit it compiled. 

@TheNeikos what do you think? Suggestions for more features or things that should be changed? 

Also I really have to test all these changes whether the binaries still work as expected. In the process I should add bash script tests for all the things as well... Because by now only the store binary is tested.

Follow-up PRs for the ugly parts should be added. Possible PRs towards this very PR as I only focused on getting this compiling.

mario-kr:
Probably not specifically related to the rewrite, but:
`imag store create -i /test/test1 entry -h header.test.int1=1234 -h header.test.str1="some string" -f -` does not fail, even though i gave an absolute path.

mario-kr:
imag-link:
```
$> imag link internal add -f test/test1 -t test/test2 // no probs
$> imag link internal add -f test/test2 -t test/test1

ERROR[   1]: StoreId 'id' part is absolute (starts with '/') which is not allowed
ERROR[   2]: StoreId handling error
link exited with non-zero exit code
```
Of course the mistake might be in imag-link, but why this false error, when i've simply linked them before?

matthiasbeyer:
@mario-kr awesome! I will have a look asap.

mario-kr:
```
imag store retrieve -c -h -i test/test1 filter-header -w header.test.int1=1234
[imag][WARN]: Filtering via header specs is currently now supported.
[imag][WARN]: Will fail now!
thread 'main' panicked at 'not yet implemented', imag-store/src/retrieve.rs:43
note: Run with `RUST_BACKTRACE=1` for a backtrace.
store exited with non-zero exit code
```

matthiasbeyer:
@mario-kr Ah, of course. well... the retrieve and filtering with imag-store is not implemented on `master` either, I'm pretty sure. So that last one is not an issue.

matthiasbeyer:
@TheNeikos I decided I will file PRs onto this very PR with cleanup for the individual tools.

Would you please review this very PR (I will fixup-rebase now) and tell me whether this looks fine. If yes, I will start filing PRs to clean up some things.

matthiasbeyer:
@TheNeikos ping

> Would you please review this very PR [..] and tell me whether this looks fine. If yes, I will start filing PRs to clean up some things.

TheNeikos:
Please also remove the version part in the macro

matthiasbeyer:
@TheNeikos and now? :smile:

matthiasbeyer:
As review is unnecessarily slow here, I will now start filing PRs onto this PR with cleanup patches for all things.

matthiasbeyer:
@TheNeikos @neithernut @mario-kr 

I'd like to get some opinions here. Please review [this subset of the PR really carefully](https://github.com/matthiasbeyer/imag/pull/667/files#diff-dfdfa6d3dc43e100116cdfd6df832223) - this file is the core change. There was also some things added as features in here, for example `StoreId::is_in_collection()`, which is just a nice addition to the new behaviour.

What I want to do: Merge this ASAP. If you can review this small subset, it would be enough for me for the first step. CI ensures all the things still compile and we also have some tests in place that guarantee us that at least the idea works.

I would merge to contrinue development. If things do not work out that well with this PR merged, we can still revert it. But I'd vote for progress here and just merge it.

Don't get me wrong. If you'd watch over the rest of the PR that'd be nice, too! But the named subset is enough IMHO.

Some technical details follow, to simplify your review process.

---

So the purpose of the PR was to rewrite the `StoreId´ type, so we can rely on these properties:

* A `StoreId` is _always_ relative from the path of the store itself.
* Joining `Store::path()` and the StoreId "local-part" yields a valid path and hence a `PathBuf`

Because of these

* A `StoreId` cannot be `Into<PathBuf>` anymore, but an explicit function call is needed for this. This is not included in the PR, but we are prepared to rip out the `Into<PathBuf>` implementation in a follow-up PR. 
* We still need the `StoreId::to_str() -> Result<String>` function. The current implementation of `StoreId::to_str()` panics at `unwrap()` if the `base` part of the StoreId is not given. A follow-up PR is needed to fix this and return `Err(_)` if the base part is not given (or some default behaviour? I'm not sure)
* We can still call `exists()` because of `impl Into<PathBuf> for StoreId` - this is convenience and will possibly vanish if `Into<PathBuf>` vanishes
* Both `StoreId::is_file()` and `StoreId::is_dir()` are still there (and the implementation is just a dummy as a `StoreId` _always_ points to a file), but a follow-up PR will remove these
* `StoreId::is_in_collection()` is a nice addition which enables the user to check whether a `StoreId` points to a specific store subdirectory.

Other things:

* The trait `IntoStoreId` yields a `Result<StoreId>` now. Because of this we can implement it for `PathBuf` and still rely on the properties `StoreId` should hold.
* `StoreId::new` and `StoreId::new_baseless` yield a `Result<StoreId>` I'm not sure whether this is clean, as `::new()` should always yield an object of the type itself, not a `Result<T>`, but for now that'd do. Maybe a follow-up PR will rename these functions to `build` and `build_baseless` respectively.

---

Basically, this PR is more about _removing_ things than _changing_ things. The PR removes the following pieces of code:

* `Deref for StoreId`
* `From<String> for StoreId`
* `AsRef<Path> for StoreId`
* `Borrow<Path> for StoreId`
* `build_entry_path`
* The version-part in the `StoreId`, hence the `version` argument for the `module_entry_path_mod!()` vanished

libimagstore/src/storeid.rs
```diff
@@ -1,140 +1,158 @@
...
-        &self.0
+    pub fn exists(&self) -> bool {
+        let pb : PathBuf = self.clone().into();
+        pb.exists()
     }
 
-}
-
-impl From<PathBuf> for StoreId {
+    pub fn is_file(&self) -> bool {
```
mario-kr: is this kept for compatibility reasons? `pub fn foo() -> bool { true }` looks kinda dangerous. Please comment (or make a TODO: implement or just something). The same applies to the next function.

matthiasbeyer: Yes, it is kept for compatibility (as is `is_dir()`) and will be removed (see my commit above).

libimagstore/src/storeid.rs
```diff
@@ -1,140 +1,158 @@
...
+    ///
+    /// A "collection" here is simply a directory. So `foo/bar/baz` is an entry which is in
+    /// collection ["foo", "bar"].
+    ///
+    /// # Warning
+    ///
+    /// The collection specification _has_ to start with the module name. Otherwise this function
+    /// may return false negatives.
+    ///
+    pub fn is_in_collection(&self, colls: &[&str]) -> bool {
```
neithernut: You use a list of strings? Wouldn't be a list of pathbufs, or a pathbuf for the directory itself be more sensible?

libimagstore/src/storeid.rs
```diff
@@ -1,140 +1,158 @@
...
+    pub fn components(&self) -> Components {
+        self.id.components()
+    }
 
-    fn from(string: String) -> StoreId {
-        StoreId(string.into())
+    /// Check whether a StoreId points to an entry in a specific collection.
+    ///
+    /// A "collection" here is simply a directory. So `foo/bar/baz` is an entry which is in
+    /// collection ["foo", "bar"].
```
TheNeikos: Hm, does the order matter? Otherwise an Array would be wrong type with which to show this

libimagstore/src/storeid.rs
TheNeikos: The problem is that a Path/PathBuf is not necessarily UTF-8 so they cannot be just converted into each other

mario-kr: So I'm not the only one getting stuck on this function :smile: 
Though my problem actually is introducing a concept called "collection" to a type which works with Paths (a lot).

libimagstore/src/storeid.rs
```diff
@@ -1,140 +1,158 @@
...
+        use std::path::Component;
+
+        self.id
+            .components()
+            .zip(colls)
+            .map(|(component, pred_coll)| match component {
+                Component::Normal(ref s) => s.to_str().map(|ref s| s == pred_coll).unwrap_or(false),
+                _ => false
+            })
+            .all(|x| x) && colls.last().map(|last| !self.id.ends_with(last)).unwrap_or(false)
```
mario-kr: Everytime I see `|x| x` I assume it's unnecessary.
I can't find any real problems in this code, so if the expected functionality (create, retrieve, get, update, delete) still works as expected, there should be no problems merging (from my side).

libimagstore/src/storeid.rs
matthiasbeyer: The idea is to check for directory names, so a `&str` perfectly works, as we can check for `["foo", "bar", "baz"]`. Feel free to suggest another interface definition, though.

libimagstore/src/storeid.rs
matthiasbeyer: Yes, it does. `["foo", "bar"]` checks whether the StoreId points to a collection inside `foo/bar/` - I see that this interface is not ideal.

libimagstore/src/storeid.rs
matthiasbeyer: Good point, actually. We can replace the `map` with `all` and erase the `all(|x| x)` therefor.

libimagstore/src/storeid.rs
neithernut: What about passing `PathBuf`s and using `PathBuf::starts_with()`?

matthiasbeyer: That might do, yes. I'm not sure, though, whether it would be more or less convenient on the users side.

I guess I change it to do what @neithernut suggested and if we need another interface as well we can add it later on.

neithernut: You may as well omit this change and have it introduced in another PR. If you don't use it anywhere already, that is.

matthiasbeyer: I'm already using it in two places and ripping it out of the history would be too hard. I just started a cleanup branch and will file another PR onto this one with my cleanup work.

libimagstore/src/storeid.rs
TheNeikos: I would then at least put a hint in the doc string!

neithernut: The interface already changed in #690.

matthiasbeyer:
How about merging this now?

@TheNeikos, @neithernut, @mario-kr ?

imag-diary/src/list.rs
```diff
@@ -21,11 +19,11 @@ pub fn list(rt: &Runtime) {
...
-    fn location_to_listing_string(id: &StoreId, base: &PathBuf) -> String {
-        id.strip_prefix(base)
+    fn entry_to_location_listing_string(e: &Entry) -> String {
+        e.get_location().clone()
+            .without_base()
+            .to_str()
             .map_err(|e| trace_error(&e))
-            .ok()
-            .and_then(|p| p.to_str().map(String::from))
             .unwrap_or(String::from("<<Path Parsing Error>>"))
```
mario-kr: This looks... dangerous. It would require the caller of the function to check, if he really got a String with isn't equal to `<<Path Parsing Error>>`
Would it hurt to return a Result?
Also: Does the or-case ever occur, if the error already got mapped?

imag-link/src/main.rs
```diff
@@ -186,6 +206,11 @@ fn handle_external_linking(rt: &Runtime) {
     if entry.is_err() {
         trace_error_exit(&entry.unwrap_err(), 1);
     }
+    let entry = entry.unwrap();
+    if entry.is_none() {
+        warn!("Entry nt found: {:?}", entry_name);
```
mario-kr: typo

imag-store/tests/001-create_test.sh
```diff
@@ -4,15 +4,15 @@ source $(dirname ${BASH_SOURCE[0]})/../../tests/utils.sh
 source $(dirname ${BASH_SOURCE[0]})/utils.sh
 
 test_call() {
-    imag-store create -p /test-call~0.2.0
+    imag-store create -p test-call~0.2.0
```
mario-kr: version is still included. I believe it shouldnt be needed anymore.

libimagtodo/src/task.rs
```diff
@@ -196,34 +198,31 @@ impl<'a> IntoTask<'a> for TTask {
...
+            .into_storeid()
+            .map_err_into(TodoErrorKind::StoreIdError)
+            .and_then(|id| {
+                store.retrieve(id)
+                    .map_err_into(TodoErrorKind::StoreError)
+                    .and_then(|mut fle| {
+                        {
+                            let mut hdr = fle.get_header_mut();
+                            if try!(hdr.read("todo").map_err_into(TodoErrorKind::StoreError))
+                                .is_none()
```
mario-kr: this line break is killing me. Sorry, but could you either write the condition in a single line or do line breaks for every chained function like you did many times above?

imag-diary/src/list.rs
matthiasbeyer: It isn't dangerous. This function is used to pass to the `CoreLister` below and its signature is as the `CoreLister` wants it.

The function _must_ return a String and if we cannot parse the location to a `String` it is fine to return the `"<<Path Parsing Error>>"` as this returned String will be printed to the screen directly by the `CoreLister`. So this functions purpose is to generate something that is _immediately_ shown to the user - if the location cannot be turned into a `String`, there's no other way than returning something that indicates that.

`trace_error()`ing the error is just another convenience for the user, but we still need to return a `String` object here.

matthiasbeyer:
@mario-kr Thanks for the review. Your notes are fixed up with the latest push.

imag-store/tests/001-create_test.sh
```diff
@@ -72,7 +72,7 @@ zzz = "z"
 EOS
 )
 
-    imag-store create -p /test-std-header-plus-custom~0.2.0 entry -h zzz.zzz=z
+    imag-store create -p test-std-header-plus-custom entry -h zzz.zzz=z
     local result=$(cat ${STORE}/test-std-header-plus-custom~0.2.0)
```
mario-kr: thats what travis failed on.

imag-store/tests/002-retrieve_test.sh
```diff
@@ -49,9 +49,9 @@ test_retrieve_content() {
...
 
 test_retrieve_header() {
-    local id="test-retrieve_simple~0.2.0"
+    local id="test-retrieve_simple"
 
-    imag-store create -p /${id} || { err "create failed"; return 1; }
+    imag-store create -p ${id} || { err "create failed"; return 1; }
 
     out "Going to test the retrieve functionality now"
     local header_out="$(retrieve --id /${id} --header)"
```
mario-kr: I dont know exactly what you are doing here, but the `/` before `${id}` might be wrong.
There's also a double negative in the following if clause.

imag-diary/src/list.rs
neithernut: This may still end up being an inconvenience for peaople calling imag from scripts where they (try to) parse the output.

matthiasbeyer: I don't care about this at this point.

PR #675, opened 2016-08-29
Add targets for the imag binary itself

matthiasbeyer:
@mario-kr Would be nice if you could review.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,10 +21,30 @@ CLEAN_TARGETS=$(foreach x,$(TARGETS),$(x)-clean)
 all: $(TARGETS)
 	@$(ECHO) "\t[ALL    ]"
 
-release: $(RELEASE_TARGETS)
+imag-bin:
+	@$(ECHO) "\t[IMAG   ][BUILD  ]"
+	@$(CARGO) build --manifest-path ./bin/Cargo.toml
+
+imag-bin-release:
+	@$(ECHO) "\t[IMAG   ][RELEASE]"
+	@$(CARGO) release --manifest-path ./bin/Cargo.toml
+
+imag-bin-update:
+	@$(ECHO) "\t[IMAG   ][UPDATE ]"
+	@$(CARGO) update --manifest-path ./bin/Cargo.toml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -21,10 +21,26 @@ CLEAN_TARGETS=$(foreach x,$(TARGETS),$(x)-clean)
...
+
+imag-bin-clean:
+	@$(ECHO) "\t[IMAG   ][CLEAN  ]"
+	@$(CARGO) clean --manifest-path ./bin/Cargo.toml
+
+release: $(RELEASE_TARGETS) imag-bin-release
 	@$(ECHO) "\t[RELEASE]"
 
-bin: $(BIN_TARGETS)
+bin: $(BIN_TARGETS) imag-bin
```
mario-kr: Rather than adding it here, Isn't it possible to add `imag-bin` to the `$(BIN_TARGETS)` variable itself, which would result in it being included in the release, update, install and clean targets automatically? Or does this introduce errors? I can't test it right now, so I'll leave it up to you.

mario-kr:
Please also edit the README.md, I'm sure i made a remark there that there's no target yet for imag-bin. Also fix the examples then.

Makefile
matthiasbeyer: That won't work as `imag-bin` is a target and `BIN_TARGETS` is a list of paths...

matthiasbeyer:
@mario-kr what do you mean by 

> Also fix the examples then.

?

mario-kr:
the ones in the README, instead of `cargo install --path /bin` nothing, because make install now does it for you.

Makefile
mario-kr: Did you try it?

mario-kr: ah, forget it.

matthiasbeyer: > Did you try it?

Nope

> ah, forget it.

ok.

matthiasbeyer:
@mario-kr can I merge?

mario-kr:
yes.

PR #690, opened 2016-09-02
General cleanup for #667

matthiasbeyer:
Here go some general cleanup patches for #667

### DIFF
--- a/libimagentrylink/src/external.rs
+++ b/libimagentrylink/src/external.rs
@@ -91,8 +91,10 @@ pub trait ExternalLinker : InternalLinker {
 
 /// Check whether the StoreId starts with `/link/external/`
 pub fn is_external_link_storeid(id: &StoreId) -> bool {
+    use std::path::PathBuf;
+
     debug!("Checking whether this is a link/external/*: '{:?}'", id);
-    id.is_in_collection(&["link", "external"])
+    id.is_in_collection(&PathBuf::from("link/external"))
 }
 
 fn get_external_link_from_file(entry: &FileLockEntry) -> Result<Url> {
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@neithernut, @mario-kr, @TheNeikos Here go the cleanup patches. Feel free to suggest more cleanup in #667 and I will add them here.

mario-kr:
To maybe spar you having to look at travis, the tests are still unchanged (still taking the array of strings for instance)

matthiasbeyer:
Jup, forgot this one. Will fix maybe tonight or tomorrow...

libimagentrylink/src/external.rs
```diff
@@ -91,8 +91,10 @@ pub trait ExternalLinker : InternalLinker {
 
 /// Check whether the StoreId starts with `/link/external/`
 pub fn is_external_link_storeid(id: &StoreId) -> bool {
+    use std::path::PathBuf;
+
     debug!("Checking whether this is a link/external/*: '{:?}'", id);
-    id.is_in_collection(&["link", "external"])
+    id.is_in_collection(&PathBuf::from("link/external"))
```
matthiasbeyer: We could use `lazy_static!()` here but I guess it is not worth the effort pulling in a dependency here, is it?

libimagstore/src/storeid.rs
```diff
@@ -86,27 +86,11 @@ impl StoreId {
...
-        self.id
-            .components()
-            .zip(colls)
-            .map(|(component, pred_coll)| match component {
-                Component::Normal(ref s) => s.to_str().map(|ref s| s == pred_coll).unwrap_or(false),
-                _ => false
-            })
-            .all(|x| x) && colls.last().map(|last| !self.id.ends_with(last)).unwrap_or(false)
+    /// This function calls `PathBuf::starts_with()` on the _local_ part of `self`.
+    pub fn is_in_collection(&self, collspec: &PathBuf) -> bool {
```
matthiasbeyer: Maybe we should rename the function, should we?

I mean, it was intended to check for _folders_, but as of now we can also check for the filename of the local part.

Maybe we should just have a function returning an `&PathBuf` which is named `local()` or something like that, and then do the `starts_with()` call on the users side.

@mario-kr, @TheNeikos, @neithernut what do you think?

PR #691, opened 2016-09-03
Bin/clap patches

mario-kr:
I actually noticed what we were doing different than @kbknapp in his examples.
We were actually defining the imag-* subcommands. This way, even with `AppSettings::AllowExternalSubcommands` enabled, clap would try to parse every argument, because it doesn't consider it an external subcommand (it's defined, so it's internal). If we just let go of defining it, it works.
@matthiasbeyer Of course this needs review and some testing from a side different than mine.
The downside is, of course, that we can't use clap to tell the users that a subcommand does not exist. What we would be able to do, though, is to check if the given SubCommand is in our list of commands, so in the case of a wrong subcommand, the `$PATH` will not be searched, probably saving us some valuable CPU cycles.


mario-kr:
Second downside:
`imag help` doesnt show our subcommands, because it is generated by clap. (and as it seems, a user-defined subcommand called `help` is ignored)

mario-kr:
Update 3:
I can't add a help text as thinks are now. I think I could, if we were not using `generate_runtime_setup` or, not any wrappers at all. For the help-text, I would need to build the app directly in `main`, so the app as well as the &str share the same lifetime.
I'll try by falling back to`get_default_cli_builder()`

mario-kr:
with `default_cli_builder` it works. @matthiasbeyer feel free to review and tell me what else may need to be done.

bin/src/main.rs
```diff
@@ -56,8 +59,10 @@ fn help(cmds: Vec<String>) {
     imag is free software. It is released under the terms of LGPLv2.1
 
     (c) 2016 Matthias Beyer and contributors"#);
+    text
```
matthiasbeyer: I would actually use the `format!()` macro here to build the `String`. You can use named formatters to make it more readable (E.G.: `format!("{value}", value=4);      // => "4"`).

mario-kr: Isn't `format!` meant for string slices?

matthiasbeyer:
@mario-kr You have a lot of chunks in your commits that do not change but only the coding style changed. This is a bit of a hassle when reviewing. Can you remove the patch for the formatting foo and submit it as another PR (maybe even later when things already hit master)?

bin/src/main.rs
matthiasbeyer: `format!()` returns a `String` AFAIK.

mario-kr: Will using `format!()` in the for-loop suffice? I dont like the idea of using this macro with these large junks of text.

matthiasbeyer: I thought of:

```rust
format!(#"
big heaps of text

{imag_binaries}

some more text

"#,
cmds.iter_into().map(|cmd| cmd_to_string(cmd)).join(somehow)
)
```

bin/src/main.rs
```diff
@@ -127,20 +133,19 @@ pub fn build_ui<'a>(app: App<'a, 'a>) -> App<'a, 'a> {
...
-    let appname  = "imag";
-    let version  = &version!();
-    let about    = "imag - the PIM suite for the commandline";
-    let rt       = generate_runtime_setup(appname, version, about, build_ui);
-    let matches  = rt.cli();
+        .help(helptext.as_str());
+    let rt = Runtime::new(app)
+        .unwrap_or_else(|e| {
+            println!("Runtime couldn't be setup. Exiting");
+            trace_error_dbg(&e);
```
matthiasbeyer: Now that I'm thinking about it: This line will not print anything. If the Runtime could not be setup, this means that the logger is not set up (not in all cases, but maybe...) so this doesn't yield anything. I'd opt for a straigt `println!()` here, actually!

matthiasbeyer: Ah nope, a `trace_error()` would do!

mario-kr: I think I took `libimagrt::setup::generate::runtime::setup` as the example for this. Please look and I'll leave it to you to fix it.

matthiasbeyer:
> It seems clap will be changed to not allow the override of the --help flag with a custom text anymore. Or do I interpret wrong? clap Issue 536

I'm sure that it will support the custom help stuff and continue to do so. That's also why I do not understand the changes in the help-text function - we can tell clap to not print the help itself by [using this function](https://kbknapp.github.io/clap-rs/clap/struct.App.html#method.get_matches_from_safe) instead of the default we use.

mario-kr:
I was wrong with the clap issue, yes. But I'm actually not sure if you linked the right function...
As I see now, clap supports custom help-texts with `.help()` function, Which prevents the generation of help output by clap. By overwriting the help-text, we get the same output with `imag help`, `--help` and `-h`, which is desirable, is it not?

matthiasbeyer:
It definitively is, yes. Ok, I will accept that part of the patchset.

And I can see now why you did not use the `generate_runtime_setup` function anymore, which also makes sense to me.

---

Besides, it would be nice if you could fix your git setup some time. Some commits say "Mario Krehl" and have a science-computing.de mail adress, others say "mario" and have your private mail adress. At least having your name right in all commits would be really nice!

---

Please to a `rebase` and:

* remove 36ea138 and a1b70e6 from the patchset
* squash 19af680 and 0661c98 into a5201d1 and provide a good commit message
* squash 544e64c into ca86ac1

---

Also would be awesome if you could rewrite all the commit message afterwards to at least start with a capital letter (I'm just trying to teach you here - not to annoy you)!

---

Besides all this: Thanks for providing patches for this PR! I appreciate it very much!

mario-kr:
rebase was successful i think. How does it look, @matthiasbeyer ?

matthiasbeyer:
Looks good! Will merge if CI succeeds!

Issue #686, opened 2016-08-31
libimagerror: Add utilities to `Result::map_err_trace()`


matthiasbeyer:
Would be helpful in some cases to have the `Result` type extended with a `map_err_trace()` function which does simply do:

```rust
Err(e).map_err(|e| { trace_error(&e); e })
```

Variants for tracing with `trace_error_maxdepth()` and `trace_error_dbg()` might be added.

---

Cleanup patches might be included in a PR for this.

PR #694, opened 2016-09-04
libimagtodo: Debug output

### LINKED ISSUE #628: imag-todo/libimagtodo: more debug output
We should add more debugging output in `libimagtodo` and also `imag-todo`.

matthiasbeyer:
(Waiting on merge of ~~#686~~ #692 )

Add some debug output in `libimagtodo`.

---

Closes #628

### DIFF
--- a/libimagtodo/src/task.rs
+++ b/libimagtodo/src/task.rs
@@ -11,6 +11,8 @@ use task_hookrs::import::{import_task, import_tasks};
 
 use libimagstore::store::{FileLockEntry, Store};
 use libimagstore::storeid::{IntoStoreId, StoreIdIterator, StoreId};
+use libimagerror::trace::MapErrTrace;
+use libimagutil::debug_result::DebugResult;
 use module_path::ModuleEntryPath;
 
 use error::{TodoError, TodoErrorKind, MapErrInto};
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@mario-kr review pls!

libimagtodo/src/task.rs
```diff
@@ -32,6 +34,8 @@ impl<'a> Task<'a> {
         r.read_line(&mut line);
         import_task(&line.as_str())
             .map_err_into(TodoErrorKind::ImportError)
+            .map_dbg_err_str("Error while importing task")
+            .map_err_dbg_trace()
```
mario-kr: how about a `map_err_dbg_trace_str()` function?

mario-kr:
The only change is additional output via Result helper functions, I see no problems here.
It builds, so let's merge.
Also: Maybe leave such easy, low priority tasks for new contributors? The more so in imag core libs like store, link, and the like

libimagtodo/src/task.rs
matthiasbeyer: Nah, that would be overkill IMHO. One has to draw a line somewhere.

matthiasbeyer:
> Maybe leave such easy, low priority tasks for new contributors? 

Yes, definitively. But I kind of fear the git hook PR actually, so I keep myself busy with the small things :smile:

PR #695, opened 2016-09-04
libimagutil: Add warn_exit() convenience helper

matthiasbeyer:
Can be helpful in some `imag-*` implementations.


neithernut:
The documentation could be improved. I had to look at the code to get what it does.

PR #706, opened 2016-09-05
Libimagstore/remove storeid into pathbuf

### LINKED ISSUE #703: There should be no Into<PathBuf> for StoreId, but an explicit function
This issue is a follow-up for #667 which is not included in #667 for size reasons of this PR.

matthiasbeyer:
Closes #703

libimagstore is already fixed, other crates will be in this PR (hence WIP).

### DIFF
--- a/libimagref/src/reference.rs
+++ b/libimagref/src/reference.rs
@@ -15,6 +15,7 @@ use libimagstore::storeid::StoreId;
 use libimagstore::storeid::IntoStoreId;
 use libimagstore::store::Store;
 use libimagerror::into::IntoError;
+use libimagerror::trace::MapErrTrace;
 
 use toml::Value;
 
@@ -230,11 +231,18 @@ impl<'a> Ref<'a> {
 
     /// Get the hash from the path of the ref
     pub fn get_path_hash(&self) -> Option<String> {
-        let pb : PathBuf = self.0.get_location().clone().into();
-        pb.file_name()
-            .and_then(|osstr| osstr.to_str())
-            .and_then(|s| s.split("~").next())
-            .map(String::from)
+        self.0
+            .get_location()
+            .clone()
+            .into_pathbuf()
+            .map_err_trace()
+            .ok() // TODO: Hiding the error here is not so nice
+            .and_then(|pb| {
+                pb.file_name()
+                    .and_then(|osstr| osstr.to_str())
+                    .and_then(|s| s.split("~").next())
+                    .map(String::from)
+            })
     }
 
     /// Get the hash of the link target which is stored in the ref object
--- a/libimagstore/src/error.rs
+++ b/libimagstore/src/error.rs
@@ -44,6 +44,7 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     StoreIdHandlingError    => "StoreId handling error",
     StoreIdLocalPartAbsoluteError => "StoreId 'id' part is absolute (starts with '/') which is not allowed",
     StoreIdBuildFromFullPathError => "Building StoreId from full file path failed",
+    StoreIdHasNoBaseError   => "StoreId has no 'base' part",
 
     CreateCallError            => "Error when calling create()",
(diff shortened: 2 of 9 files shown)


matthiasbeyer:
I'd be ready for merge here. @mario-kr review appreciated!

libimagref/src/reference.rs
```diff
@@ -230,11 +231,18 @@ impl<'a> Ref<'a> {
...
-        pb.file_name()
-            .and_then(|osstr| osstr.to_str())
-            .and_then(|s| s.split("~").next())
-            .map(String::from)
+        self.0
+            .get_location()
+            .clone()
+            .into_pathbuf()
+            .map_err_trace()
+            .ok() // TODO: Hiding the error here is not so nice
```
mario-kr: If `None` is returned in this specific case, then hiding the Error shouldnt be that big a deal. Otherwise, you'd have to change the return type to `Result<Option<String>>'

libimagstore/src/error.rs
```diff
@@ -44,6 +44,7 @@ generate_custom_error_types!(StoreError, StoreErrorKind, CustomErrorData,
     StoreIdHandlingError    => "StoreId handling error",
     StoreIdLocalPartAbsoluteError => "StoreId 'id' part is absolute (starts with '/') which is not allowed",
     StoreIdBuildFromFullPathError => "Building StoreId from full file path failed",
+    StoreIdHasNoBaseError   => "StoreId has no 'base' part",
```
mario-kr: I thought the base had become optional?

libimagstore/src/storeid.rs
```diff
@@ -62,9 +62,17 @@ impl StoreId {
...
+        let mut base = try!(self.base.ok_or(SEK::StoreIdHasNoBaseError.into_error()));
+        base.push(self.id);
+        Ok(base)
+    }
+
     pub fn exists(&self) -> bool {
-        let pb : PathBuf = self.clone().into();
-        pb.exists()
+        // TODO: hiding error here.
+        self.clone().into_pathbuf().map(|pb| pb.exists()).unwrap_or(false)
```
mario-kr: `Pathbuf::exists` returns a boolean, so calling `unwrap_or` on this expression is weird. But it builds. Do I miss a `Result` here?

libimagstorestdhook/src/flock.rs
```diff
@@ -17,25 +15,54 @@ use libimagstore::storeid::StoreId;
...
+    generate_error_imports!();
+    generate_error_types!(FlockError, FlockErrorKind,
+        IOError                    => "IO Error",
+        StoreIdPathBufConvertError => "Error while converting StoreId to PathBuf",
+        FileOpenError              => "Error on File::open()",
+        LockError                  => "Error while lock()ing",
+        UnlockError                => "Error while unlock()ing"
+    );
+}
+use self::error::FlockError as FE;
```
mario-kr: If `Flock` stands for `FileLock`, shouldn't the correct camelcase be `FLock`?

libimagref/src/reference.rs
matthiasbeyer: Yes, that would need another refactor. Feel free to open an PR for this. For now this is okay, but we need to change the return type of this function, yes.

libimagstore/src/error.rs
matthiasbeyer: Have a look where this error kind is used - sometimes we need the base part.

libimagstore/src/storeid.rs
matthiasbeyer: Yes. `into_pathbuf()` returns a `Result`, hence the `map` call.

libimagstorestdhook/src/flock.rs
matthiasbeyer: Feel free to file a PR about this. I just keep consistency here, as the hook is also named "Flock". Also, this is an internal type never meant to be exported, so I'd let that through. Not another hour on travis, you know! :-)

mario-kr:
Well, you explained every point i had, so green light from me.

PR #709, opened 2016-09-05
Fix absolute StoreId specification in imag-store test

matthiasbeyer:
We missed them in #667 - why didn't the test suite fail.

@mario-kr I'd like you to investigate. Go to master (and, if this is already merged by then, revert this patch) and find out why the test suite did not fail.

And, of course, provide patches to fix the testing-suite.

### DIFF
--- a/imag-store/tests/002-retrieve_test.sh
+++ b/imag-store/tests/002-retrieve_test.sh
@@ -23,7 +23,7 @@ test_retrieve_nothing() {
     imag-store create -p ${id} || { err "create failed"; return 1; }
 
     out "Going to test the retrieve functionality now"
-    local zero_out="$(retrieve --id ${id})"
+    local zero_out="$(retrieve --id ${id})" || return 1
     out "Retrieving for zero_out finished"
 
     if [[ ! -z "$zero_out" ]]; then
@@ -39,7 +39,7 @@ test_retrieve_content() {
 
     out "Going to test the retrieve functionality now"
 
-    local content_out="$(retrieve --id ${id} --content)"
+    local content_out="$(retrieve --id ${id} --content)" || return 1
     out "Retrieving for content_out finished"
 
     if [[ ! -z "$content_out" ]]; then
--- a/imag-store/tests/002-retrieve_test.sh
+++ b/imag-store/tests/002-retrieve_test.sh
@@ -23,7 +23,7 @@ test_retrieve_nothing() {
     imag-store create -p ${id} || { err "create failed"; return 1; }
 
     out "Going to test the retrieve functionality now"
-    local zero_out="$(retrieve --id /${id})"
+    local zero_out="$(retrieve --id ${id})"
     out "Retrieving for zero_out finished"
 
     if [[ ! -z "$zero_out" ]]; then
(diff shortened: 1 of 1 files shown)


mario-kr:
> and, if this is already merged by then, revert this patch

What do you mean with "this"? This commit in this PR here?

matthiasbeyer:
Yes.

mario-kr:
I have these suggestions, which i will implement if you "ok" them:
* Add a function `fatal()` which is the same as `err()`, but exits with `exit 1`, which should only be used for tests which are required to pass for all other tests in the same file to even be run
* Add in each script a variable called `rv` (*R*eturn *V*alue), which is set to `1` every time a test fails, and the script exits (at the end) via `exit $rv`.
* As alternative to the second one, We could define the variable in `$imag/tests/utils.sh` and set it to `1` in the `err` function. This is my favorite, as test-writers don't need to bother with setting it themselves (which leads to 10+ `rv=1` statements), but only need to be sure to use `exit $rv` or `return $rv` at the end of the script.

matthiasbeyer:
> fatal

ACK

> rv (Return Value)

Not sure about this one. It sounds good from what I understand, but I'm not sure. I will have a look at the patch.

> set it to 1 in the err function

Nope. the `err` function is for output. Nothing more.
Having a function `error_happened()` which increments the (global?) `rv` variable should be a better idea.

---

Overall I'm not quite happy with the tests. Abusing bash for tests like this is ... not nice. No, let me put it straight: it's ugly. If you can come up with another idea, feel free to open an issue and propose a new setup - or even better: a PR! :smile:

mario-kr:
Another choice would be to make test functions in rusts which construct Commands, but that's still ugly

mario-kr:
hmm, weird. in `call_test` you already cheack for return value and exit on error. And all your functions return != 0 on error. hmmm.

mario-kr:
Ok, I couldn't help but chuckle when i found the error.
The code looks like this:
```sh
1  test_retrieve_nothing() {
2      local id="test-retrieve_nothing"

3      imag-store create -p ${id} || { err "create failed"; return 1; }

4      out "Going to test the retrieve functionality now"
5      local zero_out="$(retrieve --id /${id})"
6      out "Retrieving for zero_out finished"

7      if [[ ! -z "$zero_out" ]]; then
8          err "Expected zero output, got '$zero_out'"
9          return 1
10    fi
11 }
```
In line 5, you are calling imag, but are not checking for the return value. You only later check for some expected text. However, `$zero_out` captures stdout, and not stderr of imag, So you don't notice if an error happened.
In this case, I would suggest function internal tracking of the return code, or simply returning on error:
```sh
1  test_retrieve_nothing() {
2      local id="test-retrieve_nothing"

3      imag-store create -p ${id} || { err "create failed"; return 1; }

4      out "Going to test the retrieve functionality now"
5      local zero_out="$(retrieve --id /${id})" || return 1
6      out "Retrieving for zero_out finished"

7      if [[ ! -z "$zero_out" ]]; then
8          err "Expected zero output, got '$zero_out'"
9          return 1
10    fi
11 }
```

matthiasbeyer:
Nice! I will fix this tomorrow.

matthiasbeyer:
@mario-kr can I merge?

mario-kr:
looks fine to me.

PR #711, opened 2016-09-05
Add initial test suite for imag-ref

matthiasbeyer:
Blocked by #710 by now.


matthiasbeyer:
Here's a `Makefile` missing, that's why the tests succeed.

matthiasbeyer:
Closed because we can test this with rust code because we have the store-mem-backend for testing.

PR #713, opened 2016-09-05
Use warn exit helper

matthiasbeyer:
Followup for #695 - use the new helper function whenever possible.

### DIFF
--- a/imag-tag/src/main.rs
+++ b/imag-tag/src/main.rs
@@ -8,6 +8,7 @@ extern crate libimagstore;
 extern crate libimagrt;
 extern crate libimagentrytag;
 extern crate libimagerror;
+extern crate libimagutil;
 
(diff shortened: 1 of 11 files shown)


imag-tag/src/main.rs
```diff
@@ -103,10 +103,7 @@ fn list(id: PathBuf, rt: &Runtime) {
 
     let entry = match rt.store().get(path.clone()) {
         Ok(Some(e)) => e,
-        Ok(None) => {
-            info!("No entry found.");
-            exit(1);
-        },
+        Ok(None) => warn_exit("No entry found.", 1),
```
mario-kr: you changed info to warn. Intentional?

matthiasbeyer: Yes, I guess it is okay to warn here.

PR #717, opened 2016-09-06
Use result trace helper

### LINKED ISSUE #714: Use Result::map_* helpers from #692
We introduced some helper functions in #692 which could be used in various places.

matthiasbeyer:
Closes #714 

I guess there will be a lot of CI errors as I didn't test-compile this at all.


mario-kr:
> I guess there will be a lot of CI errors as I didn't test-compile this at all.

Really?

matthiasbeyer:
> > I guess there will be a lot of CI errors as I didn't test-compile this at all.
> Really?

Yes, because this is a multi-crate change and I think it is better to stress CI than stressing my device! :smile:

PR #718, opened 2016-09-07
Bin/clap debug

mario-kr:
This PR contains some fixes, and some misc stuff:
* There was a panic when no arguments at all or no additional arguments to the subcommand were given
* Exit codes were probably decided on while the person was bored
* It wasn't checked if subcommand existed before constructing a child-process, which would then fail because the required executable wouldnt exist.
* the `unreachable!()` at the end seemed to have been reachable - by passing no subcommand, but a argument which wouldnt exit (in this case, `--versions`)
* Some comments were added (The whole file was nearly a comment-free zone)
* `--versions` Only produced error output, because it tried to call `imag-<>` without the `imag-`
* `--versions` output was prettied a tiny little bit (out of all the commits I consider this one to be optional, so I wouldnt mind reverting it)

### DIFF
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -114,7 +114,7 @@ fn main() {
     let commands = get_commands();
     let helptext = help_text(commands);
     let app      = Runtime::get_default_cli_builder(appname, version, about)
-        .settings(&[AppSettings::AllowExternalSubcommands])
+        .settings(&[AppSettings::AllowExternalSubcommands, AppSettings::ArgRequiredElseHelp])
         .arg(Arg::with_name("version")
              .long("version")
              .takes_value(false)
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -170,7 +170,11 @@ fn main() {
 
     match matches.subcommand() {
         (subcommand, Some(scmd)) => {
-            let subcommand_args : Vec<&str> = scmd.values_of("").unwrap().collect();
+            debug!("Calling with subcommand: {}", subcommand);
+            let subcommand_args : Vec<&str> = match scmd.values_of("") {
+                Some(values) => values.collect(),
+                None => Vec::new()
+            };
             debug!("Calling 'imag-{}' with args: {:?}", subcommand, subcommand_args);
 
             match Command::new(format!("imag-{}", subcommand))
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -153,7 +153,7 @@ fn main() {
     if matches.is_present("versions") {
         debug!("Showing versions");
         let mut result = vec![];
-        for command in get_commands().iter() {
+        for command in commands.iter() {
             result.push(crossbeam::scope(|scope| {
                 scope.spawn(|| {
                     let v = Command::new(format!("imag-{}",command)).arg("--version").output();
@@ -186,7 +186,7 @@ fn main() {
             };
             
             // Typos happen, so check if the given subcommand is one found in $PATH
-            if !commands.clone().contains(&String::from(subcommand)) {
+            if !commands.contains(&String::from(subcommand)) {
                 println!("No such command: 'imag-{}'", subcommand);
                 println!("See 'imag --help' for available subcommands");
                 exit(2);
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -189,7 +189,7 @@ fn main() {
                     if !exit_status.success() {
                         debug!("{} exited with non-zero exit code: {:?}", subcommand, exit_status);
                         println!("{} exited with non-zero exit code", subcommand);
-                        exit(exit_status.code().unwrap_or(42));
+                        exit(exit_status.code().unwrap_or(1));
                     }
                     debug!("Successful exit!");
                 },
@@ -208,7 +208,7 @@ fn main() {
                         },
                         _ => {
                             println!("Error spawning: {:?}", e);
-                            exit(1337);
+                            exit(1);
                         }
                     }
                 }
--- a/bin/src/main.rs
+++ b/bin/src/main.rs
@@ -169,7 +169,8 @@ fn main() {
         }
 
         for versionstring in result.into_iter().map(|handle| handle.join()) {
-            print!("{}", versionstring);
+            // The amount of newlines may differ depending on the subprocess
+            println!("{}", versionstring.trim());
         }
     }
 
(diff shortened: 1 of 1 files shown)


bin/src/main.rs
```diff
@@ -151,10 +156,10 @@ fn main() {
         for command in get_commands().iter() {
             result.push(crossbeam::scope(|scope| {
                 scope.spawn(|| {
-                    let v = Command::new(command).arg("--version").output();
+                    let v = Command::new(format!("imag-{}",command)).arg("--version").output();
                     match v {
                         Ok(v) => match String::from_utf8(v.stdout) {
-                            Ok(s) => format!("{} -> {}", command, s),
+                            Ok(s) => format!("{:10} -> {}", command, s),
```
matthiasbeyer: Why are you adding the `:10` formatting here (I'm not even sure what it does)?
I guess a comment would be nice here as well! :smile:

bin/src/main.rs
```diff
@@ -164,15 +169,31 @@ fn main() {
         }
 
         for versionstring in result.into_iter().map(|handle| handle.join()) {
-            println!("{}", versionstring);
+            print!("{}", versionstring);
```
matthiasbeyer: Why not `println!()`? Does the string already contain a newline character? Either way: Comment would be nice!

bin/src/main.rs
```diff
@@ -164,15 +169,31 @@ fn main() {
...
+            // Get all given arguments and further subcommands to pass to
+            // the imag-<> binary
+            // Providing no arguments is OK, and is therefore ignored here
+            let subcommand_args : Vec<&str> = match scmd.values_of("") {
+                Some(values) => values.collect(),
+                None => Vec::new()
+            };
+            
+            // Typos happen, so check if the given subcommand is one found in $PATH
+            if !commands.clone().contains(&String::from(subcommand)) {
```
matthiasbeyer: Are you sure you need this `clone()` here?

matthiasbeyer:
Awesome, I like your changes. I will test this after you fixed my annotations.

bin/src/main.rs
mario-kr: It's padding the string with spaces

bin/src/main.rs
mario-kr: The output of the subprocesses already contains a newline, yes. Do I really have to comment this?

bin/src/main.rs
matthiasbeyer: Ah, okay. No need for fixing this up, thanks!

bin/src/main.rs
mario-kr: I don't need it in the last usage of `commands`. In the ones before this, I use `.clone()` instead of calling `get_commands()` again, because I figured copying a `Vec<&str>` is less work than traversing `$PATH` every time to search for binaries.

bin/src/main.rs
matthiasbeyer: Would be nice if you'd do that, yes. You can, of course, squash the commits then.

bin/src/main.rs
matthiasbeyer: By the way: If there would be a single commit for this change, and the comment message telling that we can use `print!()` here instead of `println!()` because the string contains the newline... then I would say no, you don't need to. But as your commits are not that detailed, I ask you to add this comment.

---

I'm trying to teach you good commit messages, you noticed? :smile:

bin/src/main.rs
matthiasbeyer: By the way: If there would be a single commit for this change, and the comment message telling that we can use print!() here instead of println!() because the string contains the newline... then I would say no, you don't need to. But as your commits are not that detailed, I ask you to add this comment.

---

I'm trying to teach you good commit messages, you noticed? :smile:

bin/src/main.rs
matthiasbeyer: I'm pretty sure you don't need it here either, because `contains()` takes a reference to `self` (as in `&self`) as first argument...

bin/src/main.rs
```diff
@@ -164,15 +169,32 @@ fn main() {
...
     }
 
+    // Matches any subcommand given
     match matches.subcommand() {
         (subcommand, Some(scmd)) => {
-            let subcommand_args : Vec<&str> = scmd.values_of("").unwrap().collect();
+            // Get all given arguments and further subcommands to pass to
+            // the imag-<> binary
+            // Providing no arguments is OK, and is therefore ignored here
+            let subcommand_args : Vec<&str> = match scmd.values_of("") {
```
matthiasbeyer: I'm not sure whether this works properly, passing `""` as arg for `values_of()` - did you test this?

matthiasbeyer:
From what I see I would merge this.

I added one annotation we might want to talk about, but I would merge anyways and you're welcome to file a follow-up PR after the merge! What do you think?

bin/src/main.rs
mario-kr: i tested with several subcommands, e.g.:
```
> cargo run -- todo list
     Running `target/debug/imag todo list`

> cargo run -- store create -i testxy~0.2.0
     Running `target/debug/imag store create -i testxy~0.2.0`

cargo run -- store get -h -c -i testxy~0.2.0
     Running `target/debug/imag store get -h -c -i testxy~0.2.0`

[imag]
links = []
version = "0.2.0"

> cargo run -- link internal add -f testxy~0.2.0 -t testyz~0.2.0
     Running `target/debug/imag link internal add -f testxy~0.2.0 -t testyz~0.2.0`

> cargo run -- store get -h -c -i testxy~0.2.0
     Running `target/debug/imag store get -h -c -i testxy~0.2.0`

[imag]
links = ["/home/mkrehl/.imag/store/testyz~0.2.0"]
version = "0.2.0"
```
So yes, it works.

mario-kr: see the documentation of clap regarding the `AppSetting::AllowExternalSubcommands`, I got it from there. It's basically a catch-all for args.

matthiasbeyer: Wow, awesome! You don't need the `~0.2.0` part anymore, though! :smile:

mario-kr: the storeid-rewrital isnt merged yet (i think), so i still needed it to test on this branch

matthiasbeyer: > the storeid-rewrital isnt merged yet 

What do you mean (which PR)?

mario-kr: oops, my bad. it's already merged. So either bin/clap wasn't rebased yet or I forgot to rebuild the other binaries before testing :smile:

PR #719, opened 2016-09-07
Fix imag-bin-target

mario-kr:
I noticed that it failed and fixed it.


mario-kr:
ok, i better close this, looks like i got a wrong branch to checkout from...

matthiasbeyer:
@mario-kr You can tell github to use another branch as base right here in the web UI. But if you checked out the wrong branch locally, it'd be better you redo the PR, yes.

PR #721, opened 2016-09-07
libimagstorestdhook: git update hook impl

matthiasbeyer:
Compiles for me, so I'll file it. We can possibly start refactoring some things out here.


matthiasbeyer:
Merged.

PR #723, opened 2016-09-08
libimagstorestdhook: git ci fixes

matthiasbeyer:
It is not working yet, but I hope that @TheNeikos can tell me what is wrong here.

---

( @TheNeikos see your mail account, I've written a mail how to reproduce the issue I'm experiencing )

### DIFF
--- a/libimagstorestdhook/src/vcs/git/update.rs
+++ b/libimagstorestdhook/src/vcs/git/update.rs
@@ -170,10 +170,16 @@ impl StoreIdAccessor for UpdateHook {
         let message = try!(commit_message(cfg, StoreAction::Update)
                 .map_dbg_err_str("Failed to get commit message"));
 
-        repo.commit(Some("HEAD"), &signature, &signature, &message, &tree, &parents)
+        try!(repo.commit(Some("HEAD"), &signature, &signature, &message, &tree, &parents)
             .map_dbg_str("Committed")
             .map_dbg_err_str("Failed to commit")
             .map_err_into(GHEK::RepositoryCommittingError)
+            .map_into_hook_error()
+        );
+
+        index.write()
+            .map_err_into(GHEK::RepositoryIndexWritingError)
+            .map_dbg_err_str("Failed to write tree")
             .map_into_hook_error()
             .map(|_| ())
     }
--- a/libimagstorestdhook/src/vcs/git/update.rs
+++ b/libimagstorestdhook/src/vcs/git/update.rs
@@ -172,6 +172,7 @@ impl StoreIdAccessor for UpdateHook {
                 .map_dbg_err_str("Failed to get commit message"));
 
         repo.commit(Some("HEAD"), &signature, &signature, &message, &tree, &parents)
+            .map_dbg_str("Committed")
             .map_dbg_err_str("Failed to commit")
             .map_err_into(GHEK::RepositoryCommittingError)
             .map_into_hook_error()
--- a/libimagstorestdhook/src/vcs/git/update.rs
+++ b/libimagstorestdhook/src/vcs/git/update.rs
@@ -99,14 +99,6 @@ impl StoreIdAccessor for UpdateHook {
         let repo      = try!(self.runtime.repository(&action));
         let mut index = try!(fetch_index(repo, &action));
 
-        let path = try!(
-            id.clone()
-            .into_pathbuf()
-            .map_err_into(GHEK::StoreIdHandlingError)
-            .map_dbg_err_str("Failed to StoreId.into_pathbuf()")
-            .map_into_hook_error()
-        );
-
         let signature = try!(
             repo.signature()
                 .map_err_into(GHEK::MkSignature)
@@ -123,7 +115,7 @@ impl StoreIdAccessor for UpdateHook {
 
         let file_status = try!(
             repo
-                .status_file(&path)
+                .status_file(id.local())
                 .map_dbg_err_str("Failed to fetch file status")
                 .map_dbg_err(|e| format!("\t->  {:?}", e))
                 .map_err_into(GHEK::RepositoryFileStatusError)
@@ -141,7 +133,7 @@ impl StoreIdAccessor for UpdateHook {
         };
 
         try!(
-            index.add_all(&[path], ADD_DEFAULT, Some(cb as &mut IndexMatchedPath))
+            index.add_all(&[id.local()], ADD_DEFAULT, Some(cb as &mut IndexMatchedPath))
                 .map_err_into(GHEK::RepositoryPathAddingError)
                 .map_dbg_err_str("Failed to add to index")
                 .map_into_hook_error()
--- a/libimagstorestdhook/src/vcs/git/update.rs
+++ b/libimagstorestdhook/src/vcs/git/update.rs
@@ -103,30 +103,29 @@ impl StoreIdAccessor for UpdateHook {
             id.clone()
             .into_pathbuf()
             .map_err_into(GHEK::StoreIdHandlingError)
+            .map_dbg_err_str("Failed to StoreId.into_pathbuf()")
             .map_into_hook_error()
-        );
-
-        let tree_id = try!(
-            index.write_tree()
-                .map_err_into(GHEK::RepositoryIndexWritingError)
-                .map_into_hook_error()
         );
 
         let signature = try!(
             repo.signature()
                 .map_err_into(GHEK::MkSignature)
+                .map_dbg_err_str("Failed to fetch signature")
                 .map_into_hook_error()
         );
 
         let head = try!(
             repo.head()
                 .map_err_into(GHEK::HeadFetchError)
+                .map_dbg_err_str("Failed to fetch HEAD")
                 .map_into_hook_error()
         );
 
         let file_status = try!(
             repo
                 .status_file(&path)
+                .map_dbg_err_str("Failed to fetch file status")
+                .map_dbg_err(|e| format!("\t->  {:?}", e))
                 .map_err_into(GHEK::RepositoryFileStatusError)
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
Well, maybe I should've explained my problem here as well, so I'll paste the mail in here (some details omitted as they are not relevant here):

---

What I want to do:

In the Post-Update-Hook I get the modified files and add them to the
index, producing a commit with them.

The problem is, though, that git tells me that the file does not exist
on the filesystem and therefor the git hook fails to commit the
changes.

---

You can reproduce with the following steps (the test setup isn't
adapted yet):

```bash
cd imag-store && cargo build
mkdir /tmp/store
pushd /tmp/store
touch .gitkeep
git init
git add .gitkeep
git commit .gitkeep -m 'init'
popd
./target/debug/imag-store --rtp /tmp/ --config ./imagrc.toml --debug create -p test
```

You'll see that the post-update-hook gets executed and a trace is
printed:

```
ERROR[   1]: Attempt to get status of nonexistent file '/tmp/store/test'
ERROR[   2]: Error while getting file status -- caused:
ERROR[   3]: Hook exec error -- caused:
ERROR[   4]: Hook exec error -- caused:
ERROR[   5]: Post-Hook execution error -- caused:
ERROR[   6]: Hook execution error -- caused:
ERROR[   7]: Error when calling update()
```

I modified the store hook executing in 978c7b58d784dab7ff8c55c69fa89d3e0ca8784f to call the update hooks in Store::_update() rather than in the Store::update() function. I guess this is appropriate, as the update hooks should always run, not only if the user requests them.

From what I understand, the line

```rust
    try!(se.write_entry(&entry.entry));
```

(from the `Store`) causes the entry to be written to disk, am I right? But apparently it doesn't.

Would be awesome of you if you could investigate this a bit and tell me what the problem is - feel free to submit patches, of course.

matthiasbeyer:
What I can think of (from the top of my head) is that the IO actions are buffered and therefor the file does not appear on the filesystem, yet. As soon as imag shuts down, though, the kernel syncs the stuff to the disk (or as soon as the FD vanishes - I'm not sure).

This would mean that the file would appear _right after_ the `drop()` - too late for us to catch it with a hook, though.

@TheNeikos and maybe @neithernut what do you think?

mario-kr:
For testing purposes:
* If the entry is written successfully, 
* wait for 3 secs
* try the update hook
And shouldnt the create-hook handle it if you call `create`?

matthiasbeyer:
@mario-kr No, the update hook, because the `Store::create()` function is lazy and does not write to disk at all. It only creates the in-memory object. The `Store::_update()` function takes care of writing to disk.

matthiasbeyer:
Okay, the error was simple and embarrassing: git2 wants the path to the file to add relative to the repository root. Thankfully that's not that much of a change in the codebase.

mario-kr:
is there a reason "test" got deleted?

matthiasbeyer:
@mario-kr That's exactly what I'm asking. I created test and wanted it to add to the git repository and commit it. This is the state I end up with - which is clearly wrong.

mario-kr:
what about the test setup? the same as before or with the test script?

matthiasbeyer:
@mario-kr The test setup has to be adapted to the git hook implementation, as the hook implementation does not initialize the repository. It has to `git init` basically... I'm not even sure whether there has to be a commit in the repsitory before we can use the git hooks... and I also think this is stuff for later PRs, for now I would simply adapt the test setup to `git init` and `git add/commit` a `.gitkeep` file... and I did exactly that in https://github.com/matthiasbeyer/imag/pull/723/commits/a3c8500a399a45db7af430930320574229caf47c

tests/utils.sh
```diff
@@ -88,7 +88,15 @@ prepare_store_directory() {
     __git_commit 'Initial commit: .gitkeep'
 }
 
+init_git_setup() {
+    git config --global user.email "imag@imag-pim.org"
+    git config --global user.name "Imag CI"
+}
```
mario-kr: you do realize people call these test scripts with their usual user account?

matthiasbeyer: Nope. Whooooops.

libimagstorestdhook/src/vcs/git/update.rs
```diff
@@ -96,21 +99,50 @@ impl StoreIdAccessor for UpdateHook {
...
+        let file_status = try!(
+            repo
+                .status_file(id.local())
+                .map_dbg_err_str("Failed to fetch file status")
+                .map_dbg_err(|e| format!("\t->  {:?}", e))
+                .map_err_into(GHEK::RepositoryFileStatusError)
+                .map_into_hook_error()
+        );
+
+        let cb = &mut |path: &Path, _matched_spec: &[u8]| -> i32 {
```
mario-kr: what is `_matched_spec` used for? I dont see it being used.
cb isnt really descriptive for a closure definition.

libimagstorestdhook/src/vcs/git/update.rs
```diff
@@ -96,21 +99,50 @@ impl StoreIdAccessor for UpdateHook {
...
+                debug!("[GIT CREATE HOOK]: File is modified/new: {}", path.display());
+                0
+            } else {
+                debug!("[GIT CREATE HOOK]: Ignoring file: {}", path.display());
+                1
+            }
+        };
+
+        try!(
+            index.add_all(&[id.local()], ADD_DEFAULT, Some(cb as &mut IndexMatchedPath))
```
mario-kr: where is this index used?

libimagstorestdhook/src/vcs/git/update.rs
matthiasbeyer: Its a callback for `git2::Index::add_all`. It decides what to add to the index and what not.

I did name it `cb` as it might get refactored rather soon. I'm not even sure how, but I have some things in mind... but everything after the basic git support was merged.

libimagstorestdhook/src/vcs/git/update.rs
matthiasbeyer: What do you mean?

mario-kr: Couldnt find the usage before, but it seems to be used in Line 143 for writing the tree

mario-kr:
I ran it via strace, and couldnt find anything that looked like a deletion of a file. So basically it should be a problem with the git DB in the store, which leads us straight to this code (Unfortunately). I have absoultely not idea how git works in the background, so i cant really debug it to see what might go wrong.

mario-kr:
```
> cat /tmp/store/test; git status;
---
[imag]
links = []
version = "0.2.0"
---
Auf Branch master
zum Commit vorgemerkte Änderungen:
  (benutzen Sie "git reset HEAD <Datei>..." zum Entfernen aus der Staging-Area)

        gelöscht:       test

Unversionierte Dateien:
  (benutzen Sie "git add <Datei>...", um die Änderungen zum Commit vorzumerken)

        test

```

matthiasbeyer:
Nonono!

From what I understand, I do the actions in a wrong order: I add the file to the index, write the object yielded from the index to the repository and end up with an index which isn't correctly written to the repository or something. That's why I asked @alexcrichton above.

If you look at the `git log`, @mario-kr, you will see that there are indeed commits in the repository after the hook was executed. I just end up with an dirty repository and I don't know why...

mario-kr:
btw, did you read the git2-rs doc?

matthiasbeyer:
Partly, why are you asking? (Be reminded that this is an issue tracker and not a chat - feel free to talk to me about details and such in the imag IRC channel or via the mailinglist).

libimagstorestdhook/src/vcs/git/update.rs
```diff
@@ -96,21 +99,50 @@ impl StoreIdAccessor for UpdateHook {
...
+                .map_err_into(GHEK::RepositoryPathAddingError)
+                .map_dbg_err_str("Failed to add to index")
+                .map_into_hook_error()
+        );
+
+        let tree_id = try!(
+            index.write_tree()
```
mario-kr: write tree writes the index back to disc, and return the root-oid needed for a commit.

libimagstorestdhook/src/vcs/git/update.rs
```diff
@@ -130,16 +163,19 @@ impl StoreIdAccessor for UpdateHook {
...
+                .map_dbg_err_str("Failed to get commit message"));
 
         repo.commit(Some("HEAD"), &signature, &signature, &message, &tree, &parents)
+            .map_dbg_str("Committed")
+            .map_dbg_err_str("Failed to commit")
             .map_err_into(GHEK::RepositoryCommittingError)
             .map_into_hook_error()
             .map(|_| ())
-
     }
```
mario-kr: and here, after the commit, the index isnt written anymore to disk.

mario-kr: `try!(index.write())` should suffice

matthiasbeyer: You're right!

matthiasbeyer:
I guess this will work now (I already tried on my VM) and if CI succeeds completely I will merge this into the main branch and then implement custom commit messages as first step, I guess.

mario-kr:
There should be a way to check if the store is already a repository, and if no, `repo.init()` it. I think having this functionality in this PR here would be good.

matthiasbeyer:
@mario-kr Nope, not in this PR. This is certainly a thing we have to implement somewhere, somehow, but for the first steps I'm fine with having to `git init` by hand.

Later - if basic git functionality is merged - we can add configurations to auto-init a repository if there is no repo yet and `git add` everything with a `Initial import` commit.

But not now in these first steps. It is on my todo-list, though.

PR #730, opened 2016-09-09
Travis matrix builds

matthiasbeyer:
Try out matrix builds on travis, for a speed up of some things.

But first convert to use the `Makefile` for building the things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,49 +20,29 @@ before_script:
   export PATH=$HOME/.local/bin:$PATH
 script:
 - |
-  travis_cargo_run_in() {
-      echo ":: Trying to run cargo in $1"
-      [[ -d "$1" ]] &&
-      cd "$1" &&
-      {
-        {
-          if [[ $(echo "$1" | grep lib) ]]; then
-            travis-cargo -q test
-          else
-            travis-cargo -q build
-          fi
-        } &&
-        cd -
-      } || exit 1
-  }
-
-  run_sh_test() {
-      echo "--- Running test script: '$1'"
-        bash $1 || { echo "--- Test failed. Exiting"; exit 1; }
-      echo "--- Test script $1 executed successfully"
-  }
-
-  echo "<< Changes in ./doc are not build by CI >>"
-
+  libs_to_test="$(find . -maxdepth 1 name "libimag*" -printf "test-%f ")"
+  bins_to_build="$(find . -maxdepth 1 name "imag-*" -printf "%f ")"
+  make $libs_to_test &&  \
+  make $bins_to_build && \
   for d in $(find -name "Cargo.toml" | grep -vE "^./Cargo.toml$"); do
-      echo ":: Working on $d"
       dir=$(dirname $d)
       { \
-        echo -e "\n--- Running in $d ---\n" &&    \
-          travis_cargo_run_in $dir &&             \
-          echo "--- Running test scripts ---" &&  \
-          for testsh in $(find $dir -iname "*test.sh"); do
-            run_sh_test $testsh
-          done && \
-          echo -e "--- Done with test scripts ---\n\n"
+        echo "--- Running test scripts ---" &&  \
+        for testsh in $(find $dir -iname "*test.sh"); do
+          echo "--- Running test script: '$1'"
+            bash $1 || { echo "--- Test failed. Exiting"; exit 1; }
+          echo "--- Test script $1 executed successfully"
+        done && \
       } || true
   done
+
 addons:
   apt:
     packages:
     - libcurl4-openssl-dev
-    - libelf-dev
     - libdw-dev
+    - libelf-dev
+    - make
 after_success:
 - |
   pushd .imag-documentation &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,21 +20,20 @@ before_script:
   export PATH=$HOME/.local/bin:$PATH
 script:
 - |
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
I wonder how much we can do here. Doing each crate in one container would be overkill, also because things depend on eachother.

As soon as cargo workspaces are supported, more splitting up would maybe even yield _longer_ builds as we would have to wait for more machines ... 

splitting between library- and binary-crates is good for the first step, I guess.

@TheNeikos or maybe @mario-kr do you have suggestions for more split-ups?

matthiasbeyer:
Seems to work now. Wonderful. Then I can start my vacation! :smile:

matthiasbeyer:
I would even move the test-script executing to the Makefile, actually! @mario-kr Feel free to play around with the Makefile to execute tests via it! :smile: 

It could be hard, though.

matthiasbeyer:
WOW :open_mouth: 

The first test is from 50 minutes down to 14 minutes... I guess the library tests will take a bit longer, but this is awesome!

matthiasbeyer:
22 minutes for the library test build. CI didn't succeed yet because the rust-stable machines didn't spin up soon enough, so the overall time will be around 40 minutes for _this_ build, I assume. But if the machines spin up together, we have build times of less than half an hour again!

This is really awesome! I guess as soon as we have some more code in here, we can split per-crate.

matthiasbeyer:
as the builds now succeed (should also after the rebase, I guess): @TheNeikos , @neithernut and @mario-kr a short review would be appreciated!

.travis.yml
```diff
@@ -20,49 +20,32 @@ before_script:
...
-  run_sh_test() {
-      echo "--- Running test script: '$1'"
-        bash $1 || { echo "--- Test failed. Exiting"; exit 1; }
-      echo "--- Test script $1 executed successfully"
-  }
-
-  echo "<< Changes in ./doc are not build by CI >>"
+  if [[ "$TEST_SUITE" == "binaries" ]]; then
+    bins_to_build="$(find . -maxdepth 1 -name "imag-*" -printf "%f ")"
+    make $bins_to_build && \
```
neithernut: Would include a `-type d` (directories only) for good measure. And if you don't use `bins_to_build` later on, you might as well just use the `-exec` parameter, or `xargs`. I'd prefer the latter because the `find -exec program {} +` doesn't fail if `program` does (unlike `find -exec program {} \;`).

Doesn't really change anything functional, so...

.travis.yml
```diff
@@ -20,49 +20,32 @@ before_script:
...
+    for d in $(find -name "Cargo.toml" | grep -vE "^./Cargo.toml$"); do
+        dir=$(dirname $d)
+        { \
+          echo "--- Running test scripts ---" && \
+          for testsh in $(find $dir -iname "*test.sh"); do \
+            echo "--- Running test script: '$testsh'"; \
+              bash $testsh || { echo "--- Test failed. Exiting"; exit 1; }; \
+            echo "--- Test script $testsh executed successfully"; \
+          done
+        } || true
```
neithernut: The result isn't used?

matthiasbeyer: We `exit 1` if the test fails.

.travis.yml
matthiasbeyer: good point.

matthiasbeyer: As we pass the whole result from the `find` call, `xargs` wouldn't make sense at all. The variable is overhead, though.

.travis.yml
neithernut: Yes, but the result from the entire block is not used

matthiasbeyer: Because it is not needed, is it? If everything succeeds, we return `true`, else we `exit 1` - so this is perfectly fine!

neithernut: My point is that the result of the entire construct is irrelevant. You can just omit the `|| true`.

matthiasbeyer:
@neithernut here we go! :smile: Have a nice week!

matthiasbeyer:
@neithernut , @TheNeikos I would like you to approve these changes before merging them. Could you please... ? :smile:

.travis.yml
```diff
@@ -20,49 +20,30 @@ before_script:
...
-        bash $1 || { echo "--- Test failed. Exiting"; exit 1; }
-      echo "--- Test script $1 executed successfully"
-  }
-
-  echo "<< Changes in ./doc are not build by CI >>"
+  if [[ "$TEST_SUITE" == "binaries" ]]; then
+    make $(find . -maxdepth 1 -name "imag-*" -type d -printf "%f ") && \
+    for d in $(find -name "Cargo.toml" | grep -vE "^./Cargo.toml$"); do
+        dir=$(dirname $d)
+        { \
```
neithernut: You can also remove this line

.travis.yml
```diff
@@ -20,49 +20,30 @@ before_script:
...
+          echo "--- Running test scripts ---" && \
+          for testsh in $(find $dir -iname "*test.sh"); do \
+            echo "--- Running test script: '$testsh'"; \
+              bash $testsh || { echo "--- Test failed. Exiting"; exit 1; }; \
+            echo "--- Test script $testsh executed successfully"; \
+          done
+        }
```
neithernut: And this one

neithernut:
The changes look ok to me.

matthiasbeyer:
Will merge after CI succeeded.

PR #734, opened 2016-09-16
Execute tests via Makefile

matthiasbeyer:
@neithernut is this a good idea?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ release: $(RELEASE_TARGETS) imag-bin-release
 bin: $(BIN_TARGETS) imag-bin
 	@$(ECHO) "\t[ALLBIN ]"
 
+bin-test: $(BIN_TARGET_TESTS)
+
 lib: $(LIB_TARGETS)
 	@$(ECHO) "\t[ALLLIB ]"
 
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,9 @@ $(TARGETS): %: .FORCE
 
 $(BIN_TARGET_TESTS): %-test: % .FORCE
 	@$(ECHO) "\t[BINTEST]:\t$@"
-	find $(subst -test,,$@) -name "tests" -type d -exec $(MAKE) -j 1 -C {} \;
+	if [[ -f $(subst -test,,$@)/tests/Makefile ]]; then \
+		$(MAKE) -C $(subst -test,,$@)/tests || exit 1;\
+	fi;
 
 $(RELEASE_TARGETS): %: .FORCE
 	@$(ECHO) "\t[RELEASE]:\t$(subst -release,,$@)"
--- a/Makefile
+++ b/Makefile
@@ -4,12 +4,14 @@ bin = $@/target/debug/$@
 doc-crate-toml=./.imag-documentation/Cargo.toml
 
 ECHO=$(shell which echo) -e
+BASH=$(shell which bash)
 CARGO=$(shell which cargo)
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -4,12 +4,15 @@ bin = $@/target/debug/$@
 doc-crate-toml=./.imag-documentation/Cargo.toml
 
 ECHO=$(shell which echo) -e
+MAKE=$(shell which make)
```
neithernut: AFAIK make already sets the MAKE variable by default.

Makefile
```diff
@@ -65,6 +68,10 @@ $(TARGETS): %: .FORCE
 	@$(ECHO) "\t[CARGO  ]:\t$@"
 	@$(CARGO) build --manifest-path ./$@/Cargo.toml
 
+$(BIN_TARGET_TESTS): %: .FORCE
+	@$(ECHO) "\t[BINTEST]:\t$@"
+	find $(subst -test,,$@) -name "tests" -type d -exec $(MAKE) -j 1 -C {} \;
```
neithernut: Whoa.. a `find` in this place is rater... odd. The right way to do it would be to build the dependencies and let `make` do the rest for you. The way it is now diminishes the benefits of parallel builds and adds a dependency on a working `find`. (I happen to have encountered environments where find was broken/did not behave as expected.)

neithernut:
Well, if you ship a Makefile and if you have tests it is only sensible to provide a `test` target.

Makefile
matthiasbeyer: I had it without this line and it failed because `make` wasn't set... so I guess it isn't set.

Makefile
matthiasbeyer: Feel free to provide the appropriate patches to let `make` handle the things. I failed to implement it.

neithernut: I'll give it a try. Later this evening.

Makefile
```diff
@@ -4,12 +4,15 @@ bin = $@/target/debug/$@
...
 ECHO=$(shell which echo) -e
+MAKE=$(shell which make)
+BASH=$(shell which bash)
 CARGO=$(shell which cargo)
 
 BINS=$(shell find -maxdepth 1 -name "imag-*" -type d)
 LIBS=$(shell find -maxdepth 1 -name "libimag*" -type d)
 
 BIN_TARGETS=$(patsubst imag-%,,$(BINS))
+BIN_TARGET_TESTS=$(foreach x,$(BIN_TARGETS),$(x)-test)
```
neithernut: Another thing I noticed is that it's `test-x` for libraries but `x-test` for binaries. Should be the same for both.

matthiasbeyer: Yep, noticed myself as well.

Makefile
neithernut: Mh... I assume that you didn't want to just require the "tests" directory be `x/tests`?

neithernut: Why don't you just norm the location of the `tests` directory?

Makefile
neithernut: Reading this up, at least GNU make's documentation explicitly says `$(MAKE)` would point to the executable.

matthiasbeyer:
I have to rebase this as things changed on master and CI fails because of this.

matthiasbeyer:
@neithernut fine with merging this?

matthiasbeyer:
But now we should be ready... @neithernut what do you think?

neithernut:
LGTM

PR #742, opened 2016-09-17
Quality-of-life changes for testing via `make`

neithernut:
Little changes improving user experience.


neithernut:
I noticed yet another thing: `find` prints relative paths starting with `./`. This is nothing new. However, since `find` is also used to get the list of binaries and libraries, this messes up target names when prefixed (e.g. with `test-`). A quick fix would be to have `test` as a postfix for both libraries' and binaries' testing targets.
However, I still dislike relying on `find`, especially for this, as the directories are in the top-level directory anyway. You can easily do the same with a little piece of shell-code. @matthiasbeyer, what do you think?

matthiasbeyer:
Feel free to add the appropriate patches in this PR! :smile:

neithernut:
With that last commit, I assumed that it's ok to only consider `x/tests` directories (not looking any deeper in the directory tree).

Makefile
```diff
@@ -68,17 +72,19 @@ $(TARGETS): %: .FORCE
 	@$(ECHO) "\t[CARGO  ]:\t$@"
 	@$(CARGO) build --manifest-path ./$@/Cargo.toml
 
-$(BIN_TARGET_TESTS): %: .FORCE
+$(BIN_TARGET_TESTS): %-test: % .FORCE
 	@$(ECHO) "\t[BINTEST]:\t$@"
-	find $(subst -test,,$@) -name "tests" -type d -exec $(MAKE) -j 1 -C {} \;
+	if [[ -f $(subst -test,,$@)/tests/Makefile ]]; then \
+		cd $(subst -test,,$@)/tests && $(MAKE) || exit 1;\
```
matthiasbeyer: Please don't `cd` here but use `make -C $directory` instead.

neithernut:
I'll squash now...

neithernut:
Done squashing

matthiasbeyer:
Will merge as soon as CI succeeds.

PR #698, opened 2016-09-05
libimagstore: Add Store testing

matthiasbeyer:
This PR (WIP) adds tests for store functions.

---

Here goes a list what should be implemented in the tests IMO. This list will be updated as soon as more ideas come into my mind.

- [x] `move` (and variants) do alter the `FileLockEntry` location and also the `Store.entries` map
- [x] Hooks
    - [x] Working hooks
        - [x] working pre-hooks get executed
        - [x] working post-hooks get executed
    - [x] Erroring hooks
        - [x] Erroring pre-hooks get executed
        - [x] Erroring pre-hooks deny the store operations and hence do not alter the `Store.entries` map
            - [x] create
            - [x] retrieve
            - [x] update
            - [ ] move (variants)
        - [x] working post-hooks get executed, but error is returned instead of entry
            - [x] create
            - [x] retrieve
            - [x] update
            - [ ] move (variants)
    - [x] Allowed erroring hooks
        - [x] Erroring pre-hooks get executed
        - [x] Erroring pre-hooks do not deny the store operations
            - [x] create
            - [x] retrieve
            - [x] get
            - [x] update
            - [ ] move (variants)
        - [x] working post-hooks get executed (but error is returned instead of entry)
            - [x] create
            - [x] retrieve
            - [x] update
            - [ ] move (variants)

---

**Contributions to this PR are welcome!**

PR #750, opened 2016-09-18
Libimagstore/store: tests move by id test

matthiasbeyer:
@TheNeikos am I right that [here](https://github.com/matthiasbeyer/imag/blob/master/libimagstore/src/store.rs#L659) is a negation error?

This test seems to fail because of exactly this negation error I suspect here.

PR on #698.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -667,7 +667,7 @@ impl Store {
                 Ok(m)  => m,
             };
 
-            if hsmap.contains_key(&old_id) {
+            if !hsmap.contains_key(&old_id) {
                 return Err(SE::new(SEK::EntryAlreadyBorrowed, None));
             } else {
                 let old_id_pb = try!(old_id.clone().with_base(self.path().clone()).into_pathbuf());
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -650,7 +650,6 @@ impl Store {
 
     /// Move an entry without loading
     pub fn move_by_id(&self, old_id: StoreId, new_id: StoreId) -> Result<()> {
-
         let new_id = new_id.with_base(self.path().clone());
         let old_id = old_id.with_base(self.path().clone());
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@TheNeikos The line moved: https://github.com/matthiasbeyer/imag/blob/master/libimagstore/src/store.rs#L655 (caused by merge I guess...)

matthiasbeyer:
@TheNeikos please review. Especially	caa214f (see commit message).

The test succeeds now, though I'm not sure my "negation-error-fix" is actualy a fix or not.

matthiasbeyer:
@TheNeikos ping.

matthiasbeyer:
@TheNeikos ping.

matthiasbeyer:
@TheNeikos I suspect another negation error here: https://github.com/matthiasbeyer/imag/blob/libimagstore/store-tests-move-by-id-test/libimagstore/src/store.rs#L533

```rust
assert!(se.is_borrowed(), "Tried to update a non borrowed entry.");
```

Should it be `!se.is_borrowed()` or is the error message wrong?

matthiasbeyer:
@TheNeikos The latest commit tests successfully for me. Tell me what you think!

libimagstore/src/store.rs
```diff
@@ -662,21 +661,28 @@ impl Store {
...
-            } else {
-                let old_id_pb = try!(old_id.clone().with_base(self.path().clone()).into_pathbuf());
-                let new_id_pb = try!(new_id.clone().with_base(self.path().clone()).into_pathbuf());
-                match FileAbstraction::rename(&old_id_pb, &new_id_pb) {
-                    Err(e) => return Err(SEK::EntryRenameError.into_error_with_cause(Box::new(e))),
-                    Ok(_) => {
-                        debug!("Rename worked");
-                    }
+            if hsmap.contains_key(&new_id) {
+                return Err(SEK::EntryAlreadyExists.into_error());
```
matthiasbeyer: @TheNeikos I'm not sure here. The new entry could already exist, just not in the internal hashmap.

If this happens, the `FileAbstraction::rename()` call _should_ error because renaming wouldn't work if the destination path exists already, am I right?

Not sure whether this is the nicest way of doing this, though.

TheNeikos:
> Should it be !se.is_borrowed() or is the error message wrong?

Nope, that is correct. You don't want to update a non-borrowed entry

libimagstore/src/store.rs
TheNeikos: I was thinking that it would behave similarly to `mv` where it just overwrites the destination

matthiasbeyer: Nah, I wouldn't implicitely override content... this could result in a lot of issues (think of a file which is linked to another file, and now one gets overridden - links are two-way, so one would now have a dangling pointer... not so nice!).

matthiasbeyer:
Jup, I forgot that this is an `assert!()` - What do you think of the current diff?

libimagstore/src/store.rs
TheNeikos: You could invoke the deletion hooks for it?

matthiasbeyer: I would actually rather introduce a new set of hooks for this problem, but no... I like this behaviour much more, actually. Maybe we can make it configurable? Or offer alternative approaches via different functions?

matthiasbeyer:
@TheNeikos ping.

PR #751, opened 2016-09-18
libimagstore/store: tests succeeding hook tests

matthiasbeyer:
Tests for running store hooks that succeed.

PR on #698.


matthiasbeyer:
@TheNeikos try to run `cargo test` multiple times - you will see that sometimes only the test added in this PR fails and sometimes several fail.

What the hell happens here?

PR #759, opened 2016-09-20
libimagstorestdhook/git: commit on drop

### LINKED ISSUE #732: git hooks: `git-add` on `update` and `git-commit` on `Store::drop`
This should be a configuration file setting (and `on` by default IMHO).

We should only add files to the index if they are changed and commit them when the store is dropped.

This way we could have a multi-file changing `imag` call which results in _one_ commit, not `N` commits for one call to `imag`. For example: change all files which are tagged with `foo` to be tagged with `bar` - this would create `N` commits for `N` files.

matthiasbeyer:
Closes #732.

Commit on drop via store configuration.

### DIFF
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -58,6 +58,10 @@ try_checkout_ensure_branch = true
 # Commit configuration
 [store.hooks.stdhook_git_update.commit]
 
+# Enable committing here. If not enabled, the "stdhook_git_storeunload" hook
+# will commit all changes in one commit when the store is closed.
+enabled = false
+
 # Whether to do the commit interactively
 interactive = false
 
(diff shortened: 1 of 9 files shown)


matthiasbeyer:
Waiting on reply in https://github.com/alexcrichton/git2-rs/issues/159 - lets see whether I overlook something.

matthiasbeyer:
Rebased because of license header.

matthiasbeyer:
Now things should work.

@mario-kr Could you play around and test these changes (everything after 9326a83 will be refactoring and cleanup).

matthiasbeyer:
From #732:

> As the add-on-update and commit-on-drop scheme seems a bit harder to implement than I thought, I chose the do-nothing-on-update and add-and-commit-on-drop scheme, which is still okay, I guess.

PR #760, opened 2016-09-20
libimagstore/store-tests: extend create delete get test

matthiasbeyer:
WIP PR on #698 

@TheNeikos do you see the error here? The tests panic [on line 2244](https://github.com/matthiasbeyer/imag/blob/libimagstore/store-tests-extend-create-delete-get-test/libimagstore/src/store.rs#L2244) which indicates that the `Store::delete()` operation reports a false positive and the entries are not removed from the internal hashmap.

(See also the fix for the `Store::get` function I introduced to confirm my above statement is right).

### DIFF
--- a/libimagstore/src/file_abstraction.rs
+++ b/libimagstore/src/file_abstraction.rs
@@ -62,7 +62,8 @@ mod fs {
             };
         }
 
-        pub fn remove_file(_: &PathBuf) -> Result<(), SE> {
+        pub fn remove_file(path: &PathBuf) -> Result<(), SE> {
+            MAP.lock().unwrap().remove(path);
             Ok(())
         }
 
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
I see the error now: https://github.com/matthiasbeyer/imag/blob/master/libimagstore/src/file_abstraction.rs#L63 . :facepunch: @TheNeikos :tired_face:

TheNeikos:
Haha yeah, clear case of evolving codebase. The reasoning for that is simple, it was purely a fs abstraction at first rather than just a test abstraction, so it only made sense for it to not remove anything.

matthiasbeyer:
Nice that our tests work.

With the fix for the `Store::get` function I failed, because when checking the internal hashmap, the entry might be there but simply not yet loaded.

So I guess we have to first check the hashmap and if that fails we can check the FS.

PR #762, opened 2016-09-21
imag-diary: move viewer impl

### LINKED ISSUE #615: Move imag-diary functionality to libimagdiary
We have way too much functionality of the diary implemented in `imag-diary` - we should move it to `libimagdiary` where it makes sense.

I guess this will be a non-trivial task, though.

matthiasbeyer:
Closes #615 

I guess more cant be moved to the library, as all other code has commandline processing involved from what I see.


matthiasbeyer:
We could even abstract this viewer to contain other viewers than plain viewers, I guess... But for now this would do...

PR #770, opened 2016-09-27
Use the `--package` parameter rather then supplying the manifest path

neithernut:
This PR is based on #624. It tries to make `cargo` use the workspace feature when run via the `Makefile`.


neithernut:
Hey, now we at least get an error from `cargo` trying to parse the `Cargo.toml`.

neithernut:
The manual actually says that this isn't necessary.

PR #771, opened 2016-09-27
Clap validators

### LINKED ISSUE #511: Use clap validators (to validate whether imag-ref gets pointed to actual files/directories)
https://github.com/kbknapp/clap-rs/blob/master/examples/15_custom_validator.rs#L17 can be used to implement a validator.

---

I'm pretty sure we can use this elsewhere as well.

---

If the validator function goes into `libimagutil`, one needs to implement a functionality to build the function with a custom error message. This could be rather complex, but a feature we want to have IMHO. Because of this, I will tag this as complex task.

matthiasbeyer:
Closes #511

### DIFF
--- a/doc/src/04020-module-tag.md
+++ b/doc/src/04020-module-tag.md
@@ -2,6 +2,8 @@
 
 The Tagging module.
 
+A valid tag matches the regex `[a-zA-Z][0-9a-zA-Z]*`.
+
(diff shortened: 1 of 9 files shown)


matthiasbeyer:
I'm pretty sure travis does not fail because of my changes. :smile:

PR #761, opened 2016-09-20
libimagstore/store-tests: more succeeding hook tests

matthiasbeyer:
Blocked by #760 

PR on #698 

WIP

PR #772, opened 2016-09-30
libimagstore/store-tests: erroring hooks

matthiasbeyer:
Erroring-Hook-tests for `libimagstore`.

Not ready yet and should also be rebased after #761 is merged.


matthiasbeyer:
I guess I merged to early :tada: .. there are no tests yet.

PR #774, opened 2016-09-30
Copyright notice

matthiasbeyer:
I guess this is a good idea, isn't it.

@neithernut what do you think about the text? Is it appropriate?

### DIFF
--- a/.imag-documentation/src/lib.rs
+++ b/.imag-documentation/src/lib.rs
@@ -1,3 +1,22 @@
+//
+// imag - the personal information management suite for the commandline
+// Copyright (C) 2015, 2016 Matthias Beyer <mail@beyermatthias.de> and contributors
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; version
+// 2.1 of the License.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+
 //! # imag
 //!
 //! This is the _developer_ documentation for the imag personal information management suite for the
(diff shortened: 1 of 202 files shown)


matthiasbeyer:
And I have a typo. Fix incoming.

matthiasbeyer:
And I marked this hard-required as I want this in 0.2.0 included.

neithernut:
Mh... are you using a standard copyright notice template? I don't recognize it.

neithernut:
Also, I'd have to have a look at the LGPL again. Some licenses only require a notice, others want it in a specific form, yet others (short ones) want themselves copied to every single file.

matthiasbeyer:
> Mh... are you using a standard copyright notice template? I don't recognize it.

Nope.

> Also, I'd have to have a look at the LGPL again. Some licenses only require a notice, others want it in a specific form, yet others (short ones) want themselves copied to every single file.

Good point.

neithernut:
Right at the beginning of the actual license text, it says:

>  This License Agreement applies to any software library or other
program which contains a notice placed by the copyright holder or
other authorized party saying it may be distributed under the terms of
this Lesser General Public License (also called "this License").

We should have added a notice/that we should add a notice as soon as possible. The `LICENSE` file also contains a section *after* the terms and conditions, containing some recommendations regarding the notice:

> To apply these terms, attach the following notices to the library.  It is
safest to attach them to the start of each source file to most effectively
convey the exclusion of warranty; and each file should have at least the
"copyright" line and a pointer to where the full notice is found.
>
> [...]

You should just copy that template, fill out the fields and prepend it to every single source file.

matthiasbeyer:
@neithernut What do you think?

neithernut:
Mh... I'm not sure about the `... and contributors` part in the copyright info. Normally, one would list all the contributors (one per line) or one abstract entity (for projects steered by foundations). In the latter case, you'd have to "collect" the copyright from all the contributors.

Maybe you should have a look at other projects using the LGPL.

matthiasbeyer:
I'm pretty sure it doesn't matter. Listing all contributors would be way too verbose IMHO.

As we do not relicense, I'm pretty sure I do not have to ask people for acknowledgement here.

So I'd say "and contributors" or nothing :smile:

neithernut:
The copyright notice is, in a way, a separate concern from the license. So, it still can bite you.

For now, `and contributors` might be OK. However, maybe you'll be forced to change the copyright notice at some point.

matthiasbeyer:
@neithernut is this a 

> TL;DR: Ack?

? :smile:

neithernut:
It's more a disclaimer than an ACK.

matthiasbeyer:
If there are no more points left, I'll merge it now.

neithernut:
No more objections from me.

PR #775, opened 2016-09-30
libimaginteraction: readline support

matthiasbeyer:
Readline support via the rustyline crate.

### DIFF
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -11,11 +11,20 @@
 #
 [ui.cli]
 
-# History file path for readline
+# History file path for readline. Will be created by imag if it does not exist.
 readline_history_file = "/tmp/readline.history"
+
+# Number of lines to safe in the history file
 readline_history_size = 100
+
+# Ignore duplicated lines
 readline_history_ignore_dups = true
+
+# Tell if lines which begin with a space character are saved or not in the
+# history list.
 readline_history_ignore_space = true
+
+# The prompt string to use
 readline_prompt = ">> "
 
 [store]
--- a/imagrc.toml
+++ b/imagrc.toml
@@ -1,6 +1,23 @@
 # This is a example configuration file for the imag suite.
 # It is written in TOML
 
+#
+# Configuration options for the user interface
+#
+[ui]
+
+#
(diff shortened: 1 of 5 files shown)


matthiasbeyer:
@TheNeikos and maybe @neithernut What do you think: Merging this early and fixing it when we start using it (if there are errors), or waiting until some use-case appears (there clearly are, but not implemented yet) and pull in this PR then into the feature-branch of the using feature?

I would simply merge this and go on then.

matthiasbeyer:
I will rebase this now and add the license-header for the new files.

imagrc.toml
```diff
@@ -1,6 +1,32 @@
...
+#
+[ui]
+
+#
+# Configuration options for the commandline user interface
+#
+[ui.cli]
+
+# History file path for readline. Will be created by imag if it does not exist.
+readline_history_file = "/tmp/readline.history"
```
TheNeikos: Hm, I don't think that using `/tmp/` as a default is a good thing, something like `XDG_RUNTIME_DIR` might be better

matthiasbeyer: We do not (yet?) have ENV-variable expansion in the codebase... so specifying `$XDG_RUNTIME_DIR` won't work.

I'd go for this or something like `~/.imag_readline_history`, though I'm also not sure whether `~` is expanded automatically... and add support for variables later on.

matthiasbeyer:
@TheNeikos Do you want me to edit the config or can I merge this and we wait for env-var expansion support (feel free to file an issue)?

TheNeikos:
The problem with `/tmp/readline.history` is that it is not multi-user, the moment a second user tries to do anything it will error as the file already exists and is not owned by them.

matthiasbeyer:
So we should go for a `~/.imag_readline_history` or something like that. I will change this, but not today! :smile:

matthiasbeyer:
I'm not even sure whether `~` will be expanded correctly... :smiley:

matthiasbeyer:
I will go with the current settings and file an issue for more advanced things (env-variable expansion) as issue.

PR #783, opened 2016-10-03
Rewrite README.md

### LINKED ISSUE #773: Rewrite repository Readme file
We should rewrite our `README.md` file as it is a bit outdated.

matthiasbeyer:
@neithernut feel free to start some discussion on this!

Closes #773

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ outdated.
 
 ## Please contribute!
 
-We are looking work contributors!
+We are looking for contributors!
 
 There is always a number of
 [complexity/easy tagged issues](https://github.com/matthiasbeyer/imag/issues?q=is%3Aopen+is%3Aissue+label%3Acomplexity%2Feasy)
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ $> make install
 ```
 
 The installation root of the binaries may not yet be in your $PATH.
-To see, where this installation root is, check out `man cargo-install`.
+To see where this installation root is check out `man cargo-install`.
 To change the $PATH in bash:
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -1,12 +1,117 @@
 # imag - [imag-pim.org](http://imag-pim.org)
 
-Imag is a CLI PIM suite you can
-integrate in your tools of choice (Editor, MUA, RSS reader, etc etc).
+`imag` is a commandline personal information management suite.
+
+**This application is in early development. There are _some_ things that work,
+but we do not consider anything stable or usable at this moment. Feel free to
+play around anyways.**
 
 [![Build Status](https://travis-ci.org/matthiasbeyer/imag.svg?branch=master)](https://travis-ci.org/matthiasbeyer/imag)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -155,20 +102,30 @@ To test, simply add `--help` to one of the above commands:
...
+[pandoc](http://pandoc.org).
 Developer documentation is also available
 [online on github.io](https://matthiasbeyer.github.io/imag/imag_documentation/index.html).
 
 Please note that the documentation is work in progress as well and may be
 outdated.
 
+## Please contribute!
+
+We are looking work contributors!
```
mario-kr: I'm still not sure if this sentence is valid. Isn't it "We are looking for contributors!"?

README.md
```diff
@@ -1,148 +1,95 @@
...
+
 ```
 $> make install
 ```
-The installation root of the binaries (a.k.a. where they are installed to), may
-not yet be in your $PATH. To see, where this installation root is, check out
-`man cargo-install`. To change the $PATH in bash:
+
+The installation root of the binaries may not yet be in your $PATH.
+To see, where this installation root is, check out `man cargo-install`.
```
mario-kr: The comma after "To see" seems wrong. it looks like it was originally my mistake (oops).

README.md
```diff
@@ -155,20 +102,30 @@ To test, simply add `--help` to one of the above commands:
...
+
+There is always a number of
+[complexity/easy tagged issues](https://github.com/matthiasbeyer/imag/issues?q=is%3Aopen+is%3Aissue+label%3Acomplexity%2Feasy)
+available in the issue tracker you can start with and we are open to questions!
+
+Feel free to open issues for asking questions, suggesting features or other
+things!
+
+Also have a look at [the CONTRIBUTING.md file](./CONTRIBUTING.md)!
+
```
mario-kr: Add a paragraph what is currently WIP, and what you want to have implemented/for what features you are especially looking for contributors? Though this would need updates on a weekly(?) basis.

matthiasbeyer: @mario-kr I kind of like that suggestion. I really do, though we would need weekly updates or every other week maybe.

---

We have my blog articles every second week, though... so what would be better: Adding the current status here in two short sentences or simply linking the category in my blog: http://beyermatthias.de/tags/imag.html

---

Maybe we could add a link to all issues that are currently WIP... but I'm not sure whether I would want such things in the README, which is IMHO just a general overview over the project and not a detailed one.

mario-kr: then how about linking to your blog here?
Like
"For general updates on the project, as well as Open Issues and WIP, see my blog bm.de"

matthiasbeyer: Jup... I guess that's the way to go. I thought about moving the blog updates to `imag-pim.org`, though it is a lot to do IMHO and I'm not sure whether I want to do it, yet. Maybe some time, maybe I will just push articles about releases to `imag-pim.org` or something like this... We'll see.

matthiasbeyer:
@mario-kr I just added a section on where to get general updates. What do you think?

mario-kr:
looks good to me

PR #786, opened 2016-10-07
libimagstore/store: tests erroring hooks

matthiasbeyer:
PR on #698 

Adds erroring hooks testing, whether the store behaves correctly.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -879,12 +879,22 @@ impl<'a> DerefMut for FileLockEntry<'a> {
     }
 }
 
+#[cfg(not(test))]
 impl<'a> Drop for FileLockEntry<'a> {
     /// This will silently ignore errors, use `Store::update` if you want to catch the errors
     fn drop(&mut self) {
         let _ = self.store._update(self);
     }
 }
+
+#[cfg(test)]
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@TheNeikos I'm pretty sure I have found a bug.

Look at this test, please: https://github.com/matthiasbeyer/imag/pull/786/files#diff-927fe00a45c65abaa57681aa75731d58R2687

[The first assert()](https://github.com/matthiasbeyer/imag/pull/786/files#diff-927fe00a45c65abaa57681aa75731d58R2693) succeeds, so I'm positive I'm doing nothing wrong in the test itself. The second assert, though fails.

And here's what I think: We `create()` a `FileLockEntry`. The we `update()` it. The `update()` call fails, because the error is _not_ borrowed. This _cannot_ be, as the store just set the internal status to `Borrowed` in the `create()` function. Something is _really, really broken_ here, or I'm simply missing something.

matthiasbeyer:
OH FUCK.

I just found the issue, I guess:

1. We `update()`, what does the following:
  1. It `_update()`s, what sets the status to `Present`
  1. It `drop()`s the FLE, what `_update()`s the FLE, but it is `Present`, so not `Borrowed` -> the `assert!(se.is_borrowed())` fails.

PR #790, opened 2016-10-09
libimagstore/store: tests allowed erroring hooks

matthiasbeyer:
PR on #698

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -2860,5 +2860,86 @@ aspect = "test"
         assert!(store.update(fle).is_err());
     }
 
+    fn get_store_with_allowed_error_hook_at_pos(pos: HP) -> Store {
+        let mut store = get_store_with_config();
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
@TheNeikos Can you have a look please? The CI fails with a `PoisonError` - but why? I cannot reproduce localy!

matthiasbeyer:
For the sake of documentation: The first execution of the CI caused 16 tests in the erroring job. The second one caused 26 to fail. Lets see what the third iteration gives us.

matthiasbeyer:
Okay, this scares the shit out of me. The third CI run succeeded.

@TheNeikos What do you think should we do? Run the CI another 100 times? This scares me because it seems to be not that reproducible.

TheNeikos:
Well, if it just happens randomly and only on CI then we can't do much. It's pretty weird but I guess there is not much we can do?

matthiasbeyer:
Okay. Lets see what the merge commit build of this tells us...

PR #794, opened 2016-10-11
libimagstore/store-tests: fix poisonerror panic

matthiasbeyer:
PR on #698 

---

What I debugged so far:

Randomly tests fail. On the latest commit it _seems_ to be the `Store::move_by_id()` that returns an `FileNotFound` error whenever it has to return an `Ok(())` in the failing tests.

What I guess:

1. The tests run all on the same internal HashMap
1. The tests run all with the same store-id / file names
1. One test (one thread) moves the "test" storeid away and succeeds
1. Another test (thread) tries to move the "test" storeid and does not find it, because it is already moved.

If this is the error, the obvious fix would be to name all StoreIds used in tests uniquely.

I will try this now and then start my test script (`while [[ true ]]; do RUST_BACKTRACE=1 cargo test -j 16 || break; done`) on my private testing machine.

### DIFF
--- a/libimagstore/src/file_abstraction.rs
+++ b/libimagstore/src/file_abstraction.rs
@@ -69,14 +69,14 @@ mod fs {
 
         pub fn copy(from: &PathBuf, to: &PathBuf) -> Result<(), SE> {
             let mut map = MAP.lock().unwrap();
-            let a = map.get(from).unwrap().clone();
+            let a = try!(map.get(from).cloned().ok_or(SEK::FileNotFound.into_error()));
             map.insert(to.clone(), a);
             Ok(())
         }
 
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
It seems that this was indeed the problem, @TheNeikos.

I had a machine running from yesterday ~21:30 until now (08:30) with the following script:

```bash
while [[ true ]]; do RUST_BACKTRACE=1 cargo test -j 16 || break; done
```

And that script breaks from time to time on the `libimagstore/store-tests` branch, but not on this one.
I will now rebase and remove the fixup commits.

TheNeikos:
Hm, that is definitely possible since we only lock upon retrieval and not while it is 'borrowed'

PR #801, opened 2016-10-13
libimagannotation: init

matthiasbeyer:
The initial PR for the libimagannotation library, a combination of `libimagnotes` and `libimagentrylink` to easily create annotations for entries.

The library will be a convenience layer over the libraries mentioned above.


matthiasbeyer:
Moving this to 0.3.0 as the `libimaglink::internal::LinkIter` feature will be in 0.3.0.

matthiasbeyer:
@TheNeikos can you see why this fails? You wrote the `module_entry_path_mod!()` macro, so I'm asking you here.

TheNeikos:
Ya, so what is going on is that it tries to 'pull' the constant into the macro, and when it tries to resolve it it can't find it due to macro hygiene. Maybe using something like `$crate::MODULE_PATH_NAME` might help.

matthiasbeyer:
@TheNeikos Do you mean in the call or in the implementation of the macro?

TheNeikos:
The Call

matthiasbeyer:
And how does this work? I cannot do a `module_entry_path!($crate::VARIABLE)`, neither can I do a `module_entry_path!(libimaglink::VARIABLE)` as I would need to import `libimaglink` right there in `libimaglink` ... or am I misunderstanding something?

TheNeikos:
Why can't you do `module_entry_path!($crate::VARIABLE)` ?

matthiasbeyer:
Hm, that's pretty bad... as I cannot see how to work around that problem...

matthiasbeyer:
No, I will change the implementation so we do not run into this problem.

---

Notes to self:

We don't need a new subcollection in "notes/annotations" - we _really_ should _only_ provide a "annotation" command that does "note --create && link", _nothing_ more!

Issue #802, opened 2016-10-13
Rewrite libimagentrylink interface


matthiasbeyer:
Some things that come up:

* The `get_internal_links()` (and other) function(s) should return iterators
* There should be iterators that do special stuff, for example
    * An iterator that `Store::delete()`s `FileLockEntry`s that do not contain any link to other FLEs anymore
    * An iterator that filters for items that are linked to a certain set of entries
    * ... possibly more

That would help greatly in #733 !

I will mark that as hard required feature for the next (0.3.0) release therefor.

---

This issue should be solved in ~~two~~ multiple steps:

- [x] Refactor the interface of `libimagentrylink` to return Iterator objects rather than `Vec`s
- [x] Add more iterator types to do special things, like GC and so on.
- [x] Refactor the interface of `libimagentrylink` to take iterators where possible

matthiasbeyer:
I checked the last two boxes.

We have now special iterators which do things with links,... so that's done.

I also checked the interface, and I do not think that we can take iterators as input parameters... so I checked that box, too.

PR #803, opened 2016-10-13
libimagentrylink: iter interface

matthiasbeyer:
First part of #802

### DIFF
--- a/libimagentrylink/src/internal.rs
+++ b/libimagentrylink/src/internal.rs
@@ -17,8 +17,6 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 //
 
-use std::cmp::Ordering;
-
 use libimagstore::storeid::StoreId;
 use libimagstore::store::Entry;
 use libimagstore::store::EntryHeader;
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
@TheNeikos Also have a quick look here, please! :smile:

libimagentrylink/src/internal.rs
```diff
@@ -89,8 +147,8 @@ impl InternalLinker for Entry {
         add_foreign_link(link, self.get_location().clone())
             .and_then(|_| {
                 self.get_internal_links()
-                    .and_then(|mut links| {
-                        links.push(new_link);
+                    .and_then(|links| {
+                        let links = links.chain(LinkIter::new(vec![new_link]));
```
TheNeikos: You can only put one type of objects into a vec, is this what you want here?

matthiasbeyer: Hm, not sure what you mean... `get_internal_links()` returns an iterator and `Iterator::chain()` adapts two iterators to become one... that's what it does here...

TheNeikos: Never mind, I misread ^^

TheNeikos:
Looks alright to me

PR #804, opened 2016-10-14
Initial add iterators for link library

matthiasbeyer:
Next step in #802 - more iterators for more fun!

:smile: 

---

Suggestions welcome!

### DIFF
--- a/libimagentrylink/src/internal.rs
+++ b/libimagentrylink/src/internal.rs
@@ -61,12 +61,23 @@ pub mod iter {
     use toml::Value;
     use itertools::Itertools;
 
+    use libimagstore::store::Store;
+    use libimagstore::store::FileLockEntry;
+
     pub struct LinkIter(IntoIter<Link>);
 
     impl LinkIter {
 
         pub fn new(v: Vec<Link>) -> LinkIter {
             LinkIter(v.into_iter())
+        }
+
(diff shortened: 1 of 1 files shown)


libimagentrylink/src/internal.rs
```diff
@@ -104,6 +115,98 @@ pub mod iter {
...
+            self.0.next().and_then(|id| match self.1.get(id).map_err_into(LEK::StoreReadError) {
+                Ok(None)    => None,
+                Ok(Some(x)) => Some(Ok(x)),
+                Err(e)      => Some(Err(e)),
+            })
+        }
+
+    }
+
+    /// An Iterator that `Store::get()`s the Entries from the store while consumed
```
matthiasbeyer: Documentation wrong.

Also: We delete all entries that are in the iterator. We should really include a hook that denies that action if there's a link left in the entry... what would make this iterator pointless, wouldn't it?

matthiasbeyer:
I guess this CI run will fail now because I didn't even care to test compile this...I will do more work here... and I just pushed as backup.

matthiasbeyer:
@TheNeikos I could use some help here, actually.. I get the `F: FnOnce` things wrong and don't know why.

matthiasbeyer:
I really should merge this after the 0.2.0 release, am I right?

PR #809, opened 2016-10-15
libimagstorestdhook: deny delete of linked entries

matthiasbeyer:
This adds a hook which denies `Store::delete()` if there are still links in the entry.

### DIFF
--- a/libimagstorestdhook/src/denylinkeddelete.rs
+++ b/libimagstorestdhook/src/denylinkeddelete.rs
@@ -32,7 +32,8 @@ use libimagentrylink::internal::InternalLinker;
 mod error {
     generate_error_imports!();
     generate_error_types!(NoLinksLeftCheckerHookError, NoLinksLeftCheckerHookErrorKind,
-        LinksLeft => "The entry has links and therefor cannot be deleted."
+        ReadInternalLinksError  => "Error while reading internal links of entry",
+        LinksLeft               => "The entry has links and therefor cannot be deleted."
     );
 }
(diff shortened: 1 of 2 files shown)


libimagstorestdhook/src/denylinkeddelete.rs
```diff
@@ -0,0 +1,116 @@
...
+        debug!("[NO LINKS LEFT CHECKER HOOK] {:?}", fle.get_location());
+
+        let n = fle
+            .get_internal_links()
+            .map(|i| i.count())
+            .map_warn_err_str("[NO LINKS LEFT CHECKER HOOK]: Cannot get internal links")
+            .map_warn_err_str("[NO LINKS LEFT CHECKER HOOK]: Assuming 1 to automatically abort")
+            .map_dbg_err_str("[NO LINKS LEFT CHECKER HOOK]: Printing trace now")
+            .map_err_trace()
+            .unwrap_or(1);
```
neithernut: You assume one link exists if you failed to retrieve links?

matthiasbeyer: Yes, that automatically causes the hook to abort the delete action, as we cannot be sure that there are links. If we assume `0` and there are links left, we end up with malformed/inconsistent data. If we assume `1`, we get the error messages from above and everything stops.

That's what I thought of. What I not thought of: we might be more safe to just return the error, right? :smile:

neithernut: Yep.

PR #810, opened 2016-10-15
Single crate for all binaries

mario-kr:
I tried making all of imag into one crate, but only succeeded partially.
The project [plantex](https://github.com/OsnaCS/plantex) could be considered my source of inspiration.

What I have right now is:
- A Cargo.toml specifying every binary as a part of `imag-project` (Thats only a temporary name)
- The same Cargo.toml containing each and every dependency of all the imag-binaries

With this, I can call for example `cargo build` in the project root, and all binaries would be built.

What is a problem:
- the libraries are not includable as the binaries are, therefore not part of `imag-project`, but still single crates. This leads to
- `cargo test` tests only the binaries, not the libraries. And
- Adding a `workspace` containing the libraries does not solve this problem (I don't really know how workspaces work, I just figured I'd give it a shot)

With this, the individual binaries' `Cargo.toml` might be considered obsolete, as it is still possible to build single binaries via `cargo build --bin imag-store` for example.

I only open this PR as a suggestion, and for discussion. Therefore, some things are simply not done yet, like:
- Deleting the unneeded `Cargo.toml`s
- Updating the Makefile
- Updating the README.md/other Documentation

As of the creation of this PR, `cargo build`, `cargo install` and `cargo test` work as described above. (A minor fix in `libimagstore/src/store.rs` was needed and is part of this PR)


mario-kr:
Minor Update:
`cargo doc` works too. Every library in the `Cargo.toml` will get documented, so the `.imag-documentation` crate might become obsolete.

matthiasbeyer:
Hm. I'm not sure whether we want something like this, as we lose the ability to install imag without certain parts (e.g. "install all of it, but not imag-mail").

I'd welcome discussion, though.

mario-kr:
You could still keep the makefile and change the install command to `cargo install --bin imag-<...>`.
I'd actually recommend this, so there's still a central way to to do all the stuff, including testing of libraries.

matthiasbeyer:
Well, feel free to continue to play around. This might be a way to release a single imag crate. Though I almost decided to rather release each crate as one crate via a script...

neithernut:
Another thing to consider: one day, if `imag` internal interfaces are somewhat stable, we might split it up into its individual libraries and binaries.

mario-kr:
There's a different solution in place for releasing imag. I will close this PR but keep the branch.

PR #812, opened 2016-10-17
libimagentrylink/filter predicates

matthiasbeyer:
This PR adds some types for filtering with operations on the links of an entry as filter predicate.

What I want to implement:
- [ ] Filter on entries where the predicate function gets the links of an entry
- [ ] Filter on Link objects

### DIFF
--- a/libimagentrylink/src/internal.rs
+++ b/libimagentrylink/src/internal.rs
@@ -17,9 +17,11 @@
 // Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 //
 
-use std::collections::BTreeMap;
 #[cfg(test)]
 use std::path::PathBuf;
+
+use std::collections::BTreeMap;
+use std::cmp::Ordering;
 
 use libimagstore::storeid::StoreId;
 use libimagstore::storeid::IntoStoreId;
@@ -391,6 +393,7 @@ pub mod pred {
         //! for links
 
         use std::error::Error;
+        use std::cmp::Ordering;
 
         use filters::filter::Filter;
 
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@TheNeikos I could also need some help here... I do not get the call for the closure right and I really don't know why...

libimagentrylink/src/internal.rs
```diff
@@ -265,6 +265,83 @@ pub mod iter {
...
+        use std::error::Error;
+
+        use filters::filter::Filter;
+
+        use libimagstore::store::Entry;
+        use libimagerror::trace::trace_error;
+
+        use super::super::InternalLinker;
+
+        pub enum LinkCountOp {
```
matthiasbeyer: Replace this by `std::cmp::Ordering`.

libimagentrylink/src/internal.rs
```diff
@@ -265,6 +265,83 @@ pub mod iter {
...
+        }
+
+        impl Filter<Entry> for FilterLinkCount {
+            fn filter(&self, entry: &Entry) -> bool {
+                match entry.get_internal_links() {
+                    Err(e)    => (self.errfn)(&e),
+                    Ok(links) => match self.op {
+                        LinkCountOp::LT => links.count() < self.n,
+                        LinkCountOp::EQ => links.count() == self.n,
+                        LinkCountOp::GT => links.count() > self.n,
```
matthiasbeyer: With `std::cmp::Ordering` we can break this into a oneliner: `a.cmp(b) == ord`.

matthiasbeyer:
Closing as this was basically forgotten since it started.

Issue #813, opened 2016-10-18
Iterator for Iterator<Item = Result<T, Error>> tracing


matthiasbeyer:
That would be a rather nice playground for some beginners to play around with iterators.

What I want is basically a generic iterator that is able to trace (as in `libimagerror::trace::*`).

I want to be able to call

```rust
iterator.trace_unwrap(); 
// The above is a: Iterator<Item = Result<T, Error>> -> Iterator<Item = T>
// which `trace_error()`s the errors and then filters them out in the output iterator

iterator.on_err(|e| debug!("Error happened: {:?}", e))
// The above returns the iterators items unmodified, but calls the closure
// on each error from the iterator

iterator.unwrap_with(|e| debug!("Error happened: {:?}", e))
// The above is a: Iterator<Item = Result<T, Error>> -> Iterator<Item = T>
// which calls the passed closure for each `Error` and filters them out in the output iterator
```

Function names are just made up and one could probably come up with better names.

---

I mark this as `complexity/easy` - it's not for complete rust beginners, though, but I guess it is good for someone who already knows Rust a bit and wants to play with traits and Iterators.

All the work will be done in `libimagerror`, in a new submodule `iter`, I guess.

Will submit this issue to TWIR.

Thiez:
Would something like [this](https://github.com/Thiez/imag/commit/1065f8ad33e96d5b3bc03c9162039d9be96176fb) (unpolished, no documentation) represent the semantics you have in mind?

matthiasbeyer:
@Thiez this looks really good, yes. Before you submit a PR, please make sure you do have a feature branch and do not submit a PR from master, please!

Besides that, I'm looking forward to see your PR and I'm eager to see what the CI tells us on this! :smile:

matthiasbeyer:
Whops, should close this :smile:

PR #820, opened 2016-10-20
Add Iterator for Iterator<Item = Result<T, Error>> tracing

Thiez:
Adds a new trait `libimagerror::iter::TraceIterator` that provides the methods as described in #813.

### DIFF
--- a/libimagerror/src/iter.rs
+++ b/libimagerror/src/iter.rs
@@ -1,0 +1,244 @@
+//
+// imag - the personal information management suite for the commandline
+// Copyright (C) 2016 the imag contributors
+//
+// This library is free software; you can redistribute it and/or
+// modify it under the terms of the GNU Lesser General Public
+// License as published by the Free Software Foundation; version
+// 2.1 of the License.
+//
+// This library is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+// Lesser General Public License for more details.
+//
+// You should have received a copy of the GNU Lesser General Public
+// License along with this library; if not, write to the Free Software
+// Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+//
+
+use std::error::Error;
+
+/// An iterator that maps `f` over the `Error` elements of `iter`, similar to `std::iter::Map`.
+///
+/// This `struct` is created by the `on_err()` method on `TraceIterator`. See its
+/// documentation for more information.
+#[must_use = "iterator adaptors are lazy and do nothing unless consumed"]
+#[derive(Clone)]
+pub struct OnErr<I, F>{
+    iter: I,
+    f: F
+}
+
+impl<I, F, T, E> Iterator for OnErr<I, F> where
+    I: Iterator<Item = Result<T, E>>,
+    F: FnMut(&E)
+{
+    type Item = Result<T, E>;
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
+            Some(Err(e)) => {
+                (self.f)(&e);
+                Some(Err(e))
+            },
+            other => other
+        }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+}
+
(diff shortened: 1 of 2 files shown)


libimagerror/src/iter.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+    }
+
+    #[inline]
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        self.iter.size_hint()
+    }
+}
+
+impl<I, F> ExactSizeIterator for OnErr<I, F> where
```
matthiasbeyer: And why this one?

libimagerror/src/iter.rs
```diff
@@ -0,0 +1,244 @@
...
+    OnErr<I, F>: Iterator
+{
+}
+
+impl<I, F, T, E> DoubleEndedIterator for OnErr<I, F> where
+    I: DoubleEndedIterator<Item = Result<T, E>>,
+    F: FnMut(&E)
+{
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
```
matthiasbeyer: ..and why this one? :smile:

matthiasbeyer:
This looks awesome! I do not understand why you implement the `ExactSizedIterator` and `DoubleEndedIterator`, though.

I do not mind these lines, I am just curious what the advantage will be (okay,... badly worded, the advanatage is that we can iterate backwards of course... but ... why do we need it?).

libimagerror/src/iter.rs
```diff
@@ -0,0 +1,244 @@
...
+
+    #[inline]
+    fn next(&mut self) -> Option<Self::Item> {
+        match self.iter.next() {
+            Some(Err(e)) => {
+                (self.f)(&e);
+                Some(Err(e))
+            },
+            other => other
+        }
```
matthiasbeyer: I'd prefer a `self.iter.next().map(|el| el.map_err(|e| { (self.f)(&e); e }))` here, actually.

libimagerror/src/iter.rs
```diff
@@ -0,0 +1,244 @@
...
+{
+    #[inline]
+    fn next_back(&mut self) -> Option<Self::Item> {
+        match self.iter.next_back() {
+            Some(Err(e)) => {
+                (self.f)(&e);
+                Some(Err(e))
+            },
+            other => other
+        }
```
matthiasbeyer: I'd prefer a `self.iter.next().map(|el| el.map_err(|e| { (self.f)(&e); e }))` here, actually.

matthiasbeyer:
Wonderful PR, thanks a lot for it! Also :+1: for adding tests! :heart:

libimagerror/src/iter.rs
Thiez: For the same reason, because we might as well implement it when we know it's applicable.

libimagerror/src/iter.rs
Thiez: Same reason. It's all the rage, just look at the stdlib iterator implementations 😋!

libimagerror/src/iter.rs
Thiez: As would I, but people are always calling me crazy when I'm using all those nice adapter methods, so I figured I'd do the explicit `match` for once 😄

Thiez:
Okay, so in conclusion: yank out all the trait implementations that aren't strictly necessary, and rewrite `<OnErr as Iterator>::next` method?

libimagerror/src/iter.rs
matthiasbeyer: Awesome. I will merge these things.

(That's also why I didn't include these comments in my review... because I didn't knew about the advantages. TIL!)

libimagerror/src/iter.rs
matthiasbeyer: Well, ... I don't mind them, actually.

Do you want to keep it this way (for the sake of the people yelling at you :smiley: ) or rewrite it?
I guess I would merge both. :+1:

matthiasbeyer:
> Okay, so in conclusion: yank out all the trait implementations that aren't strictly necessary, and rewrite <OnErr as Iterator>::next method?

Nonono, keep the additional trait impls, I'm fine with them! I just did not know the advantages. About the other things: If you want, you can rewrite them. If you care. If not... I guess I'll fine with the current impl.

Thiez:
I've modified `next` and `next_back` as suggested, I also like them better like this. As far as I'm concerned we can go ahead with the merge, provided you and travis are happy :smile:

PR #824, opened 2016-10-25
libimagmail/linker

matthiasbeyer:
PR for the linker module that creates inter-mail-ref links based on message ids and in-reply-to header fields.

First implementation, not optimized and also not usable by now... we need some infrastructure code for this as well.


matthiasbeyer:
@TheNeikos any ideas how I can abstract away the double-mut-borrow here?

TheNeikos:
I am not sure what you mean? The error is about a trait missing

matthiasbeyer:
Nutin' said!

matthiasbeyer:
@TheNeikos now you look! :smile:

PR #826, opened 2016-10-28
Bake in shell-compl generation

### LINKED ISSUE #495: Use clap to generate bash compl?
This might be complicated, actually, as we have a multi-crate setup here.

But it might be worth a try.

It would us require to update clap to `2.9.*`.

matthiasbeyer:
Therefor we need to have at least clap 2.16.1.

We bake this into libimagrt::runtime::Runtime, so all binary crates
should automatically have it.

---

Closes #495

Or does it? We then have shell compl for `imag-foo` but not yet for `imag` itself. At least a step into the right direction, IMHO.

### DIFF
--- a/libimagrt/src/runtime.rs
+++ b/libimagrt/src/runtime.rs
@@ -85,8 +85,9 @@ impl<'a> Runtime<'a> {
         match matches.value_of(Runtime::arg_generate_compl()) {
             Some(shell) => {
                 debug!("Generating shell completion script, writing to stdout");
-                let shell = shell.parse::<Shell>().unwrap(); // clap has our back here.
-                cli_spec.gen_completions_to("fakename", shell, &mut stdout());
+                let shell   = shell.parse::<Shell>().unwrap(); // clap has our back here.
+                let appname = String::from(cli_spec.get_name());
+                cli_spec.gen_completions_to(appname, shell, &mut stdout());
             },
             _ => debug!("Not generating shell completion script"),
         }
--- a/libimagrt/src/runtime.rs
+++ b/libimagrt/src/runtime.rs
@@ -57,6 +57,9 @@ impl<'a> Runtime<'a> {
      */
     pub fn new(cli_spec: App<'a, 'a>) -> Result<Runtime<'a>, RuntimeError> {
         use std::env;
+        use std::io::stdout;
+
+        use clap::Shell;
 
         use libimagstore::hook::position::HookPosition as HP;
         use libimagstore::hook::Hook;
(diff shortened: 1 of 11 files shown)


libimagrt/src/runtime.rs
```diff
@@ -79,6 +82,15 @@ impl<'a> Runtime<'a> {
 
         Runtime::init_logger(is_debugging, is_verbose, colored);
 
+        match matches.value_of(Runtime::arg_generate_compl()) {
+            Some(shell) => {
+                debug!("Generating shell completion script, writing to stdout");
+                let shell = shell.parse::<Shell>().unwrap(); // clap has our back here.
+                cli_spec.gen_completions_to("fakename", shell, &mut stdout());
```
mario-kr: maybe `cli_spec : App` has a function to retrieve the app name? So we could have `imag-<foo>` instead of `fakename`

libimagrt/src/runtime.rs
```diff
@@ -79,6 +82,15 @@ impl<'a> Runtime<'a> {
...
+            },
+            _ => debug!("Not generating shell completion script"),
```
mario-kr: more like: `unreachable!()`. clap makes sure you have an argument, and it even makes sure you got a valid one. I can't think of a way to reach there. And if it was possible, `warn!` or `err!` might be more appropriate.

libimagrt/src/runtime.rs
```diff
@@ -302,6 +323,10 @@ impl<'a> Runtime<'a> {
         "editor"
     }
 
+    pub fn arg_generate_compl() -> &'static str {
```
mario-kr: A static string stored/accessible in/through a function. Can't really understand it. Please explain or change.

mario-kr: Ok, you probably want to add it to the `arg_names` function.

libimagrt/src/runtime.rs
matthiasbeyer: Not in this context. If you do not provide the argument, you do not generate the compl scripts, and `unreachable!()` would yield a panic.

mario-kr: yeah, youre right.

libimagrt/src/runtime.rs
matthiasbeyer: Yes, we need that definitively, though `App` does not have a function for this. We have to do it in another way...

matthiasbeyer:
Okay, this works now but we have to wait for https://github.com/kbknapp/clap-rs/pull/726 as we need to pass the actual imag binary name to the generator. With https://github.com/kbknapp/clap-rs/pull/726 we can do this, if it gets rejected, we have to rewrite some abstractions in imag.

@mario-kr FYI.

libimagrt/src/runtime.rs
mario-kr: Wouldn't it be possible to use `<app>.p.meta.bin_name` or `<app>.p.meta.name` directly, at least until youre PR is in an official release (if it doesn't get rejected)?

matthiasbeyer:
Cool! We have the feature in clap, we must wait for a release of clap and then we can have it... this would push imag to a new clap version, ... but I'm pre-1.0, so what will you do???

matthiasbeyer:
Awesome, clap 2.17 was released, so we can use it to get the app name from the `clap::App` object and pass it back to `App::gen_completions_to()`! :smile:

matthiasbeyer:
@mario-kr I would still like to merge this feature here, even if #832 covers some aspects of this very PR as well. Could you do a short review?

---

Maybe we revert changes from this PR as soon as #832 is merged. I don't know yet.

matthiasbeyer:
@mario-kr feel free to play around before we merge this! :-)

PR #832, opened 2016-11-02
Add shell completion

mario-kr:
I present a solution for the "shell-completion problem", though the "how" may be a bit... discussable.
It somehow works, which is kind of surprising, as it literally is my first macro in rust.
There definitely is space for improval, as there is still duplicate code (9 times `.subcommand(...)`).
Testing is definitely welcome, as this work is in the "10 minutes ago I ran it and it actually worked (on my machine)" kind of state.
Also this kind of solution will require maintenance when adding new binaries to imag. Furthermore, the completion is generated at compiletime, which might or might not be considered a downside.

### DIFF
--- a/bin/build.rs
+++ b/bin/build.rs
@@ -7,16 +7,35 @@ extern crate libimagutil;
 use clap::Shell;
 use libimagrt::runtime::Runtime;
 
-macro_rules! gen_types_buildui {
-    ($(($p:expr, $n:ident)$(,)*)*) => (
+/// This macro generates mods with the given '$modulename',
+/// whose content is the file given with `$path`.
+/// In this case, It is used specifically to include the
+/// `ui.rs` files of the imag binaries.
+/// The imag project (accidentally?) followed the convention
+/// to write a `ui.rs` containing the function
+/// `fn build_ui(app : App) -> App`.
+/// This macro allows us to use the same named functions by
+/// putting them each into their own module.
+macro_rules! gen_mods_buildui {
+    ($(($path:expr, $modulename:ident)$(,)*)*) => (
         $(
-            mod $n {
-                include!($p);
+            mod $modulename {
+                include!($path);
             }
          )*
         )
 }
 
+/// This macro reduces boilerplate code.
+///
+/// For example: `build_subcommand!("counter", imagcounter)`
+/// will result in the following code:
+/// ```ignore
+/// imagcounter::build_ui(Runtime::get_default_cli_builder(
+///     "counter",
+///     &version!()[..],
+///     "counter"))
+/// ```
 macro_rules! build_subcommand {
     ($name:expr, $module:ident) => (
         $module::build_ui(Runtime::get_default_cli_builder(
@@ -26,7 +45,8 @@ macro_rules! build_subcommand {
     )
 }
 
-gen_types_buildui!(
+// Actually generates the module.
+gen_mods_buildui!(
     ("../imag-bookmark/src/ui.rs",  imagbookmark),
     ("../imag-counter/src/ui.rs",   imagcounter),
     ("../imag-diary/src/ui.rs",     imagdiary),
@@ -40,10 +60,12 @@ gen_types_buildui!(
 );
 
 fn main() {
+    // Make the `imag`-App...
     let mut app = Runtime::get_default_cli_builder(
         "imag",
         &version!()[..],
         "imag")
+        // and add all the subapps as subcommands.
         .subcommand(build_subcommand!("bookmark",   imagbookmark))
         .subcommand(build_subcommand!("counter",    imagcounter))
         .subcommand(build_subcommand!("diary",      imagdiary))
@@ -55,6 +77,7 @@ fn main() {
         .subcommand(build_subcommand!("todo",       imagtodo))
         .subcommand(build_subcommand!("view",       imagview));
 
+    // Actually generates the completion files
     app.gen_completions("imag", Shell::Bash, env!("OUT_DIR"));
     app.gen_completions("imag", Shell::Fish, env!("OUT_DIR"));
     app.gen_completions("imag", Shell::Zsh, env!("OUT_DIR"));
--- a/bin/build.rs
+++ b/bin/build.rs
@@ -17,6 +17,15 @@ macro_rules! gen_types_buildui {
         )
 }
 
+macro_rules! build_subcommand {
+    ($name:expr, $module:ident) => (
+        $module::build_ui(Runtime::get_default_cli_builder(
+                $name,
+                &version!()[..],
+                $name))
+    )
+}
+
 gen_types_buildui!(
     ("../imag-bookmark/src/ui.rs",  imagbookmark),
     ("../imag-counter/src/ui.rs",   imagcounter),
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Wow. 

I do not know what implications this brings with it... I will play around on a dedicated machine with it. Great work!!!

bin/build.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn main() {
+    let mut app = Runtime::get_default_cli_builder(
+        "imag",
+        &version!()[..],
+        "imag")
+        .subcommand(
+            imagbookmark::build_ui(Runtime::get_default_cli_builder(
+                    "bookmark",
+                    &version!()[..],
```
matthiasbeyer: I'm pretty sure we have to pass the version of the binary we build the Runtime for here. So for imag-bookmark in this case.

matthiasbeyer: You could also abstract away the whole `build_ui` calling by building everything inside the `subcommand()` call via a macro. This way you could then do `subcommand(ui_build_for!(imagbookmark, "bookmark"))` as far as I can see..

bin/build.rs
```diff
@@ -0,0 +1,93 @@
...
+            imagview::build_ui(Runtime::get_default_cli_builder(
+                    "view",
+                    &version!()[..],
+                    "view")))
+        ;
+    app.gen_completions("imag", Shell::Bash, env!("OUT_DIR"));
+    app.gen_completions("imag", Shell::Fish, env!("OUT_DIR"));
+    app.gen_completions("imag", Shell::Zsh, env!("OUT_DIR"));
+
+}
```
matthiasbeyer: Missing newline at end of file :smile:

bin/build.rs
```diff
@@ -0,0 +1,93 @@
...
+    ("../imag-counter/src/ui.rs",   imagcounter),
+    ("../imag-diary/src/ui.rs",     imagdiary),
+    ("../imag-link/src/ui.rs",      imaglink),
+    ("../imag-notes/src/ui.rs",     imagnotes),
+    ("../imag-ref/src/ui.rs",       imagref),
+    ("../imag-store/src/ui.rs",     imagstore),
+    ("../imag-tag/src/ui.rs",       imagtag),
+    ("../imag-todo/src/ui.rs",      imagtodo),
+    ("../imag-view/src/ui.rs",      imagview)
+    );
```
matthiasbeyer: Don't indent the closing parentheses here please.

matthiasbeyer:
I'm also a bit insecure as we now _implicitely_ depend on all `imag-*` crates here. Also, we build the autocompl for _all_ `imag-*` crates, but we do not know whether the user wants them all installed.

I'd love if you could investigate whether there is a possibility to run the `build.rs` "script" with flags,... this way we could add only the compl we need. :heart:

matthiasbeyer:
If I type `imag --rtp /hom<tab>`, I get `imag --rtp <runtimepath>` - not sure if this is a bug in clap or in our setup here? @kbknapp

kbknapp:
@matthiasbeyer This is because it's no longer shell completing, but using the script to get completion options. And in the bash completion script in particular it'll fill `--rtp` with a placeholder value `<runtimepath>`

There is work ongoing in kbknapp/clap-rs#644 to allow using a custom completion value. I.e. instead of just a generic `<runtimepath>` placeholder, you could say something like, "complete `--rtp` with directories" or something like that.

matthiasbeyer:
Awesome! So I will for https://github.com/kbknapp/clap-rs/issues/644 before merging this.

bin/build.rs
mario-kr: Please keep in mind that we're generating shell completions here, so the version simply does not matter, and cannot appear in the output this script generates.
I'll look into your other idea though. Most likely, I will keep the first macro, and make a call to `$x::build_ui(...)` in the one you suggested.

mario-kr:
The only input which can be given to a build script are environment variables, and in the given list [on the bottom of this page](http://doc.crates.io/environment-variables.html) I couldn't find a suitable one. The (imo) most promising one was `CARGO_FEATURE_<name>`, until i was reminded that a feature needed to be present so we would be able to `build --features <feature_list>`

mario-kr:
With the motto "We have to go deeper" in mind, we might as well generate this build-script via a bash-script like `gen_compl_script.sh ./ bookmark counter store view > build.rs && cargo build`

kbknapp:
Actually kbknapp/clap-rs#568 may also be applicable

matthiasbeyer:
> generate this build-script via a bash-script 

Where the bash script itself is created from a piece of Ruby which falls out of a Python script that was generated by C++ code which was code-to-code compiled from Rust.

Nope, sorry, this doesn't look like a good solution to me! :-)

mario-kr:
Thought so :smile: 
I was thinking of pulling this stuff into the main.rs, und determine with the `get_commands` function which imag binaries are currently in the `$PATH`. I won't look at it until evening though.

mario-kr:
@matthiasbeyer I need some help on this one: If I use `get_commands`, I have a list of `String`s, but how can I translate those to module names, preferably without doing something ugly? The ugly idea currently is in my head like this:
```
fn add_subcommand(s : String, app : &mut App) -> mut App {
  match s {
    "bookmark" => app.subcommand(imagbookmark::build_ui(App::new())),
    "counter"  => app.subcommand(imagcounter::build_ui(App::new())),
    ...
    // silent error: if nothing matched, return the unchanged original app
    _ => app,
  }
}
...
pub fn main() {
  ...
  match gen_completion {
    for command in commands {
      add_subcommand(command, app);
    }
    app.gen_completions_to(...);
  }
  ...
}
```

matthiasbeyer:
I would do it in another way:

```rust
app.subcommand(build_subcommand!("bookmark", imagbookmark))
      .subcommand(build_subcommand!("notes" imagnotes))
...
```

just as suggested above. This way you don't have the problem you're facing.

bin/build.rs
```diff
@@ -0,0 +1,86 @@
...
+            mod $modulename {
+                include!($path);
+            }
+         )*
+        )
+}
+
+/// This macro reduces boilerplate code.
+///
+/// For example: `build_subcommand!("counter", imagcounter")`
```
matthiasbeyer: Here's a typo: `imagcounter"` should be `imagcounter`.

bin/build.rs
```diff
@@ -0,0 +1,86 @@
...
+/// imagcounter::build_ui(Runtime::get_default_cli_builder(
+///     "counter",
+///     &version!()[..],
+///     "counter"))
+/// ```
+macro_rules! build_subcommand {
+    ($name:expr, $module:ident) => (
+        $module::build_ui(Runtime::get_default_cli_builder(
+                $name,
+                &version!()[..],
```
matthiasbeyer: We still use the version of the `imag` crate itself here, but we really should use the version of the crate we include. Maybe you can provide a function in each `imag-*` binary `version()` which calls `version!()` inside, so we can get the actual `imag-*` version passed to `build_ui` here?

mario-kr: As written above, simply not necessary. not at all. I might as well write an empty string there

matthiasbeyer: Ah, yes. I'm a bit tired I guess... sorry!

matthiasbeyer: You could add a short comment about that, so I do not "accidentially" fix this in the future :-)

mario-kr:
I just `rebase`d and `sqash`ed a bit. The commit history should be nice and clean now.

PR #834, opened 2016-11-04
Don't re-open temporary file.

Stebalien:
`Write` is implemented on both `&NamedTemporaryFile` and `&File` so you don't actually need a mutable reference. I wish there were a better way to do this but such is life.


matthiasbeyer:
Good catch! Thanks a lot!

PR #835, opened 2016-11-14
libimagstore: extract toml functionality

matthiasbeyer:
Extract the TOML query functionality to another module, as extension on `toml::Value`. The first step for a `toml-rs-querying` crate! :smile:


matthiasbeyer:
@TheNeikos any objections? The next PR would be to remove the `struct EntryHeader` entirely and make it a `type EntryHeader = Value`.

Some functionality from `impl EntryHeader` can be removed (including some slight refactoring where it gets used) and some functionality must be reimplemented using traits.

That'd go into another PR, as I want to keep this PR targeting only one crate, not multiple.

TheNeikos:
I can only comment the Idea, it's way too much code for me to comment on that.

It seems like a good change though (if I understood correctly)

matthiasbeyer:
@TheNeikos Basically I'm extracting the `toml::Value` query functionality to a new trait which is then implemented on `toml::Value`. Before it was implemented on a wrapper type `EntryHeader`.

The next step is to either submit it upstream (this functionality) if wanted (I already have such a PR running, but not completed yet) or provide an extension crate for `toml-rs`.

The implementation has to be optimized, though, as I do some nasty stuff in there. Follow-Up PRs will come.

matthiasbeyer:
@TheNeikos The ultimate goal would be to eliminate the `EntryHeader` type as a whole and rely on `toml::Value`. Maybe I'm fine with a `type EntryHeader = Value`.

TheNeikos:
Alright, doesn't sound bad

PR #836, opened 2016-11-15
libimagstore: Eliminate EntryHeader type

matthiasbeyer:
Followup for #835 

This eliminates the `EntryHeader` type by moving the functionality to traits implemented on `toml::Value`.

Refactoring is about to be done.

### DIFF
--- a/libimagentrylink/src/external.rs
+++ b/libimagentrylink/src/external.rs
@@ -38,6 +38,7 @@ use libimagstore::store::FileLockEntry;
 use libimagstore::store::Store;
 use libimagstore::storeid::StoreId;
 use libimagstore::storeid::IntoStoreId;
+use libimagstore::toml_ext::TomlValueExt;
 use libimagutil::debug_result::*;
 
 use error::LinkError as LE;
--- a/libimagentrytag/src/tagable.rs
+++ b/libimagentrytag/src/tagable.rs
@@ -19,8 +19,9 @@
 
 use itertools::Itertools;
 
-use libimagstore::store::{Entry, EntryHeader};
+use libimagstore::store::Entry;
 use libimagerror::into::IntoError;
+use libimagstore::toml_ext::TomlValueExt;
 
 use error::TagErrorKind;
 use error::MapErrInto;
@@ -43,7 +44,7 @@ pub trait Tagable {
 
 }
 
-impl Tagable for EntryHeader {
+impl Tagable for Value {
 
     fn get_tags(&self) -> Result<Vec<Tag>> {
         let tags = try!(self.read("imag.tags").map_err_into(TagErrorKind::HeaderReadError));
--- a/libimagref/src/reference.rs
+++ b/libimagref/src/reference.rs
@@ -33,6 +33,7 @@ use libimagstore::store::FileLockEntry;
 use libimagstore::storeid::StoreId;
 use libimagstore::storeid::IntoStoreId;
 use libimagstore::store::Store;
+use libimagstore::toml_ext::TomlValueExt;
 use libimagerror::into::IntoError;
 
 use toml::Value;
--- a/libimagstore/Cargo.toml
+++ b/libimagstore/Cargo.toml
@@ -40,7 +40,27 @@ default = []
 verify  = []
 
 # Enable panic!()s if critical errors occur.
-# Can be used to debug the store more intensly via the imag-store commandline
-# application
+#
+# # Howto
+#
+# To enable this, put
+#
+# ```toml
+# [features]
+# early-panic = [ "libimagstore/early-panic" ]
+# ```
+#
+# In the crate depending on this library and compile your crate with
+# `cargo build --features early-panic`. This way, the `libimagstore`
+# implementation fails via `panic!()` instead of propagating errors which have
+# to be printed somewhere to be visible.
+#
+# # WARNING
+#
+# The behaviour of the store implementation might be broken with this, resulting
+# in partially written store entries and/or worse, so this is
+#
+#    _NOT INTENDED FOR PRODUCTION USE_!
+#
 early-panic=[]
 
--- a/libimagstore/Cargo.toml
+++ b/libimagstore/Cargo.toml
@@ -39,3 +39,8 @@ env_logger = "0.3"
 default = []
 verify  = []
 
+# Enable panic!()s if critical errors occur.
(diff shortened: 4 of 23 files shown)


matthiasbeyer:
Not sure why the binary tests fail. @mario-kr would you like to test? I will continue fixing the library tests, but would be awesome if you could test the binary tests.

matthiasbeyer:
So as it seems, the headers are screwed up somehow. `imag-store create` does not create an entry anymore, but the debug output does not provide a error message, so I guess there is some serious bug in here.

---

One thing that should be done before merging this is moving the free functions out of the public scope here. We introduce the public module `toml_ext` from the store, which contains public utility functions (because we need them in `store` and `toml_ext`). We should move them to a `util` module.

matthiasbeyer:
@TheNeikos The above failed - but when we tested it, it didn't. That's because travis here uses rustc 1.11 still, and we are on 1.14 ... hmmmm

mario-kr:
rebasing the branch might help considerung the `.travis.yml`.

matthiasbeyer:
@mario-kr that'd be a fix, yes. Anyways, I refactored the thing to use a custom macro internally, which should also work on rustc 1.11!

matthiasbeyer:
Rebased. @mario-kr was probably right ... still, I included the custom macro patch, so we either define `if_cfg_panic!()` as empty or as wrapper to `panic!()`. Could be of use later, too.

mario-kr:
Ah, the regex update is throwing sticks between your feet.

matthiasbeyer:
Jup. That would have caused a CI-failure after merge, so good that I rebased this!

matthiasbeyer:
Now things get funny! :smile:

matthiasbeyer:
@mario-kr rebased, fixupd, Ci succeeds... what do we want more? Feel free to review now! :smile:

libimagentrylink/src/external.rs
```diff
@@ -312,6 +313,8 @@ impl ExternalLinker for Entry {
 
     /// Set the external links for the implementor object
     fn set_external_links(&mut self, store: &Store, links: Vec<Url>) -> Result<()> {
+        use libimagstore::toml_ext::TomlValueExt;
```
mario-kr: is this not included at the top of the file already?

libimagentrytag/src/tagable.rs
```diff
@@ -76,6 +77,8 @@ impl Tagable for EntryHeader {
     }
 
     fn set_tags(&mut self, ts: &[Tag]) -> Result<()> {
+        use libimagstore::toml_ext::TomlValueExt;
```
mario-kr: same here? Am i missing a detail of rust here?

libimagref/src/reference.rs
```diff
@@ -390,6 +391,8 @@ impl<'a> Ref<'a> {
     /// Update the Ref by re-checking the file from FS using the passed Hasher instance
     /// This errors if the file is not present or cannot be read()
     pub fn update_ref_with_hasher<H: Hasher>(&mut self, h: &H) -> Result<()> {
+        use libimagstore::toml_ext::TomlValueExt;
```
mario-kr: same

libimagstore/Cargo.toml
```diff
@@ -39,3 +39,8 @@ env_logger = "0.3"
 default = []
 verify  = []
 
+# Enable panic!()s if critical errors occur.
+# Can be used to debug the store more intensly via the imag-store commandline
+# application
```
mario-kr: an example would be nice.

matthiasbeyer:
I should probably rewrite the history of this branch a bit more, so we can have a nice history.

matthiasbeyer:
Now I consider this history in a mergeable shape! :small_airplane:

libimagstore/src/store.rs
```diff
@@ -1147,11 +984,26 @@ impl Entry {
     pub fn new(loc: StoreId) -> Entry {
         Entry {
             location: loc,
-            header: EntryHeader::new(),
+            header: Entry::default_header(),
             content: EntryContent::new()
         }
     }
 
+    pub fn default_header() -> Value { // BTreeMap<String, Value>
```
mario-kr: further down (in `toml_ext` somewhere), you have a trait called `Header`, which has s function with the same name.
Is it possible, to also implement this trait for `Entry`? If not, could you make a call here to `Value::default_header()` to reduce duplicate code? It seems to do the exact same thing...

libimagstore/src/toml_ext.rs
```diff
@@ -353,10 +358,104 @@ impl Extract for Value {
...
+        use toml::Parser;
+
+        let mut parser = Parser::new(s);
+        parser.parse()
+            .ok_or(ParserErrorKind::TOMLParserErrors.into())
+            .and_then(verify_header_consistency)
+            .map(Value::Table)
+    }
+
+    fn default_header() -> Value {
```
mario-kr: see previous comment

libimagstore/src/util.rs
```diff
@@ -0,0 +1,35 @@
...
+macro_rules! if_cfg_panic {
+    ()                       => { panic!() };
+    ($msg:expr)              => { panic!($msg) };
+    ($fmt:expr, $($arg:tt)+) => { panic!($fmt, $($($arg),*)) };
+}
+
+#[cfg(not(feature = "early-panic"))]
+#[macro_export]
+macro_rules! if_cfg_panic {
+    ()                       => { };
```
mario-kr: is it not possbile to do just sth like `($i)* => {};`?

matthiasbeyer: I used the macro implementation of `panic!()` itself as reference here, so ... no, I won't change that!

mario-kr: ah, ok. was just wondering if there'd be an easier way.

libimagstore/src/store.rs
matthiasbeyer: Fixed up with 337a03b.

It is not possible to implement `Header` for `Entry`, as an Entry is not a Header. The `Header` trait is _only_ for extending the `toml::Value` type. But I fixed what you suggested by calling `Value::default_header()` in this function.

matthiasbeyer:
Oh yeah... food would be a nice thing now! Lets see whether CI says yes, then rebase (fixup) and then I'll grab some food to (procrastinating from learning for my exams... shame on me)!

PR #837, opened 2016-11-16
Add older compilers for CI jobs

matthiasbeyer:
Coming from https://github.com/kbknapp/clap-rs/pull/752, I suggest we also include old(er) compilers into our CI setup, so we do not end up with `?` syntax before all users have a compiler that understands that syntax.

1.11.0 could be removed maybe after 1.14.0 was released, 1.12.0 after 1.15.0 was released... this way we make sure that all/most users have at least 1.13.0.

---

@TheNeikos what do you think?


TheNeikos:
I see no problems with that. I do it similarly over at (`rustbreak`)[https://github.com/TheNeikos/rustbreak/blob/master/.travis.yml#L7]

You shouldn't need to add more than one though, since they are 'forwards compatible' so you usually only test against the oldest one (and the latest)

matthiasbeyer:
So what do you suggest? Using 1.12.1 or 1.0.0 ? :smile:

TheNeikos:
The earliest compatible

matthiasbeyer:
Lets see how far we can get... :smile:

matthiasbeyer:
@TheNeikos looks good, right? So I guess we can merge this as soon as all jobs succeed...

PR #840, opened 2016-12-05
libimagmail/bulk import

matthiasbeyer:
Add functionality to bulk-import mails.

This is a naive implementation.


matthiasbeyer:
Closed because this is too old for putting effort in it to making it mergeable. Rather rewrite.

PR #842, opened 2016-12-05
Fix: Only list internal links here, no external ones

### LINKED ISSUE #841: Listing internal links also lists external links (indirectly)
Doing something like
```rust
imag link internal --list <entry>
```
where entry does have external links, results in output like this:
```
0  : counter/c1
1  : links/external/<some lengthy hash>
```
So basically, when listing internal links, we also get links which are more related to external links.
IMO, this does not make sense.
I suggest filtering out every store entry in the `links` subdirectory, or at least in the `links/external` directory.

matthiasbeyer:
Funny side note:

As the commit history reveals, I wasn't active in this repository for
some time.

I first implemented a solution for this problem in about 50 lines of
code. Then I thought that this could be done way simpler, so I rewrote
it in about 20 lines.

Finally, I noticed that I already have infrastructure for this, so this
two-line patch should fix the problem.

Know your codebase, dude!

---

Closes #841 

@mario-kr could you test?

### DIFF
--- a/imag-link/src/main.rs
+++ b/imag-link/src/main.rs
@@ -84,7 +84,7 @@ fn main() {
 
 fn handle_internal_linking(rt: &Runtime) {
     use libimagentrylink::internal::InternalLinker;
-    use libimagentrylink::external::iter::NoExternalIter;
+    use libimagentrylink::external::is_external_link_storeid;
 
     debug!("Handle internal linking call");
     let cmd = rt.cli().subcommand_matches("internal").unwrap();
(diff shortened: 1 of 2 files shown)


mario-kr:
LGTM (on the surface, no fancy test cases).
You could introduce a flag to still list the external ones (for debugging purposes and similar)

matthiasbeyer:
> You could introduce a flag to still list the external ones

Open a feature request! :smile: 

Yes, I see that this could be helpful!

imag-link/src/main.rs
```diff
@@ -96,6 +97,13 @@ fn handle_internal_linking(rt: &Runtime) {
                 match get_entry_by_name(rt, entry) {
                     Ok(Some(e)) => {
                         e.get_internal_links()
+                            .map(|iter| {
+                                if cmd.is_present("list-externals-too") {
+                                    iter.filter(|_| true)
```
mario-kr: I suggest 
```rust
iter.filter(|id| cmd.is_present("list-externals-too") || !is_external_link_storeid(&id))
```

matthiasbeyer:
@mario-kr can be merged, right?

mario-kr:
ACK

Issue #768, opened 2016-09-26
Properly document link semantics


neithernut:
The discussion on #765 showed that the semantics behind links are not as clear as they should be. The documentation should be expanded to explain these.

matthiasbeyer:
The preparation is done. Feel free to file your PR.

matthiasbeyer:
@neithernut Do you want to file a PR on this, or shall I write up some documentation for this?

neithernut:
I would defer this until the link functionality was expanded by an (optional) name as suggested in #765.

matthiasbeyer:
AFAICR we decided (in #765) to not alter the link functionality, but rather use intermediate entries with custom data in the header for adding semantics to links.

neithernut:
Not exactly. We concluded that the *option* to name links (or associations) would be beneficial or even necessary for some (future) modules featuring item functionality dealing with some items' relations.

mario-kr:
@matthiasbeyer I just came across something I wouldn't expect this way:
When I `imag link internal --list <some_entry>`, the external links are listed indirectly.
While I think it's perfectly fine to implement external links in such a way, I didn't really expect the internal link to the external-link-object to be listed.

matthiasbeyer:
@mario-kr Could you please file a new issue about that?

mario-kr:
the external-link being listed, or the inconsistency between external/internal? (or both?)

matthiasbeyer:
Both!

mario-kr:
But this certainly belongs here:
I just noticed that `imag link external --remove` does not take an argument; How do I know which external link I am deleting? May there only be one? When deleting is there some logic within external links to not delete external links, if they are pointed to by another entry?

matthiasbeyer:
> I just noticed that imag link external --remove does not take an argument;

This is a bug, indeed.

matthiasbeyer:
I'm marking this as done, assuming that it is documented enough already.

PR #843, opened 2016-12-05
Fix: We need an argument here

matthiasbeyer:
Closes a bug reported in https://github.com/matthiasbeyer/imag/issues/768#issuecomment-264851895

Note that we still need to specify the full URL here.


matthiasbeyer:
@mario-kr Feel free to test.

Issue #846, opened 2017-01-17
Focus-Shift: imag.rb


matthiasbeyer:
This is mainly the discussion-platform and description of the [0.3.0 milestone](https://github.com/matthiasbeyer/imag/milestones/3).

---

# Focus-shift: imag.rb

Why: [Summed up here](https://beyermatthias.de/blog/2017/01/17/whats-coming-up-in-imag-22/)

TL;DR: Remove own integrations, build Ruby-Gem out of core libraries,
because bootstrapping tools/integrations in Ruby is faster.

- [x] Remove:
    - [x] imag-bookmark
    - [x] imag-counter
    - [x] imag-diary
    - [x] imag-mail
    - [x] imag-todo
    - [x] libimagbookmark
    - [x] libimagcounter
    - [x] libimagdiary
    - [x] libimagmail
    - [x] libimagnotification
    - [x] libimagtodo

- [x] Keep
    - [x] imag-link
    - [x] imag-ref
    - [x] imag-store
    - [x] imag-tag
    - [x] imag-view (though a Ruby tool might come up with way better functionality)
    - [x] libimagentrytag
    - [x] libimagentryview
    - [x] libimagerror
    - [x] libimagref
    - [x] libimagrt
    - [x] libimagstore
    - [x] libimagstorestdhook
    - [x] libimagutil

- [ ] Keep maybe (though there might be a better/more flexible impl in Ruby at some point):
    - [ ] imag-notes
    - [ ] libimagnotes

- [ ] Add build infrastructure for Ruby
  - [ ] Testing of the Ruby-Gem
    - [ ] rspec?
    - [x] Unit tests?
  - [x] CI infrastructure

- [ ] Remove maybe (if not used elsewhere):
  - [ ] libimagtimeui
  - [ ] libimaginteraction

## And then...

Build `imag.rb`, a Gem to talk from Ruby to `libimagstore` (+`libimagerror`
and other foundation-libs) and later also to `libimagentrytag` and
`libimagentrylink`).

After that is done, we have to
- [ ] re-check all issues in the upcoming milestones whether they still apply to the codebase.

mario-kr:
I'd like to suggest/note a few things here:
* You forgot `imag-todo`, which probably goes onto the "to-delete"-list
* Maybe you'd want to keep a tool like imag-counter and/or  imag-notes as an example implementation of a backend in Rust
* Probably a restructuring of the releases/milestones. (which you also kind of mentioned)
[...]

PR #847, opened 2017-01-18
Imag ruby

matthiasbeyer:
Part of #846 

Closes #371 as outdated.

This PR will introduce `imag-ruby`, a Ruby gem to script imag, build with [ruru](https://github.com/d-unseductable/ruru).

This very first PR will only build the interface for `libimagstore` and the required core libraries (error, runtime).

---

# Todo

- [x] Translate to Ruby classes:
    - [x] FileLockEntry
    - [x] Store
    - [x] StoreId
    - [x] ~~Runtime~~ (Not needed as we create the `Store` object directly)
    - [x] ~~Configuration~~ (Not needed as we create the `Store` object directly)
    - [x] ~~Store Error types (would be convenience)~~ (Will be done in another PR)
    - [x] ~~Runtime Error types (would be convenience)~~ (Not needed)
    - [x] imag logging functionality, to be able to print logs in imag format
- [x] Make everything work
    - [x] Getting and setting headers of entries
    - [x] Getting and setting content of entries
- [x] Build abstractions around the core functionality to have nice things

---

The most complex part is the ownership-model we implemented in the imag store library... the Ruby code "owns" the `FileLockEntry` objects while it is running, though I'm not sure whether we can model this correctly (with `update()` on `drop()`, as Ruby has no destructors, [just finalizers](https://stackoverflow.com/questions/5956067/ruby-destructors)).


matthiasbeyer:
I just managed to build a ruby extension which contains the `StoreId` type as `RStoreId` and exports it to Ruby! AWESOME :tada:

matthiasbeyer:
The latest travis-ci test will fail as travis does not have ruby 2.3 yet. Building ruby on travis is no option, so I will push a commit which disables these tests with the next push.

matthiasbeyer:
The more code I write the more I notice that we need a translation functionality for `toml <-> ruru`.

@mario-kr do you want to get freaky? :smile:

Makefile
```diff
@@ -67,6 +67,9 @@ lib: $(LIB_TARGETS)
 
 lib-test: $(LIB_TARGETS_TEST)
 
+lib-imag-ruby-test:
```
mario-kr: Is this not covered by `$(LIB_TARGETS_TEST)`?

mario-kr:
No.

Also, you should consider expanding a bit on the matter, I dont really get what you actually want to have.

matthiasbeyer:
@mario-kr See [here](https://github.com/matthiasbeyer/imag/pull/847/files#diff-f5d7d38ea5eda7b43cc9b505ba7406c3) for example :smile:

Makefile
matthiasbeyer: Nope, it does something else. It runs ruby files. But as travis does not have ruby 2.3, which is the minimum version to build with "ruru", we cannot have travis-CI builds for this :cry:

matthiasbeyer:
@mario-kr Have a loooooook: https://github.com/matthiasbeyer/imag/pull/847/files#diff-a448437bf6e37dfd5c4b165dc5ff78e7

It woooooorks!

matthiasbeyer:
As far as I can see right now, I only need the `Runtime` object and then we can start scripting imag in Ruby - or am I missing something?

matthiasbeyer:
@mario-kr It works! I can talk to the Store now! I can actually create multiple store. Note that there must be a `imagrc.toml` in the runtimepath and it must alow implicit store creation if the `/store` directory does not exist in the runtimepath.

But besides that, the tests scripts in this very branch work!

There is a lot of things that might not yet work, especially actually _working_ with the entries from the store... 

---

I'd love if you could go ahead and script some Ruby scripts to test what works (I will do, too)!

mario-kr:
I'll have to pass on this one, I've never used ruby before and am currently not willing to learn an additional language (even though i heard that ruby's somewhat easy)

matthiasbeyer:
Ah, kay. I thought you used Ruby before... well, maybe @TheNeikos is interested? :smile:

Makefile
malept: > But as travis does not have ruby 2.3

Hi, could you elaborate on this? In my [rusty_blank](https://github.com/malept/rusty_blank/blob/master/.travis.yml) example extension, I install Ruby successfully on a Travis Rust image.

matthiasbeyer: Yes, you can install it on travis, but travis does not ship with Ruby.

I'm not sure what overhead it takes to install Ruby. Right now we have build times of ~3 minutes, which is really awesome. I don't want to give that up. But well, we need testing of the Ruby script interface, I clearly see that.

I guess my plan will be to include Ruby script interface testing as soon as we can talk of something stable. So I will merge _this_ branch this week hopefully and then start my work on bringing the other libs into the Ruby scope. As soon as I got something one can work with, I will enable tests with Ruby on travis, so we can work towards a release!

Thanks for your input!

malept: > I'm not sure what overhead it takes to install Ruby.

It depends on the OS and what's already installed on the given Travis worker image. Currently on the Trusty containers, at least 2.3.1 is installed (so it's only necessary to switch to the appropriate Ruby, the cost of which is negligent). Not sure what's installed on the OSX images, but it seems to take ~42 seconds to install the 2.3.1 binary.

matthiasbeyer: Okay. Sounds good. As we do not build for OSX right now, this should be not our concern here.

Maybe we should switch to trusty containers then! :+1:

malept:
Out of curiosity, have you considered using [Thermite](https://github.com/malept/thermite) to build/distribute the Rust binaries for the gem, so you don't have to require that users of the gem install Cargo/Rust?

matthiasbeyer:
Not yet. To be fair, this whole Ruby thing is (as of now) really experimental. Feel free to file a few issues on yuor thoughts about the Ruby infrastructure we need to set up Would be awesome! If you can, point to the right documentation to get it right, because I havn't touched the Ruby environment in ... years!

matthiasbeyer:
@mario-kr I think all TODOs from the intro-comment of this PR are done. Shall we merge this? :octocat:

mario-kr:
I dont know ruby, so I'm not the one to ask 😄 
but have an ACK anyway.

PR #848, opened 2017-01-18
Remove top-lvl Cargo.toml to work with rust beta

matthiasbeyer:
It seems like this causes a build-failure on master ...


matthiasbeyer:
@mario-kr any objections? I would love to have a PR from you that actually fixes the problem instead of working around it by removing the `Cargo.toml` - do you know how to fix this properly?

mario-kr:
```toml
[[bin]]
name ="imag"
path ="bin/src/main.rs"
```
might do the trick

mario-kr:
or remove it, if it doesnt fulfill some purpose

PR #852, opened 2017-01-19
Add workspace support

mario-kr:
This PR introduces proper cargo workspaces. Changes:
* The `Cargo.toml` removed in an earlier PR of @matthiasbeyer is readded.
Furthermore, missing workspace members are added
* The `[profile.dev] codegen-units =  2` configuration is put into the top-level `Cargo.toml`, as there are warnings that profile definitions are ignored if they are put in lower-level ones.
* The required minimum version of rust is changed to `1.13.0`, as this one seems to be the first to support workspaces.

Changed behaviour:
* Calling `make -j <threads>` to have more things build parallel does not yield any effect now, as the build directory now is a common `target` directory in the workspace, and cargo locks it when building something.
* `make bin-test` as well as `make lib-test` now have a runtime (on travis) of approximately 4 minutes each. (less than half of what it was before)
* `make clean` now deletes the top-level `target`-directory - as many times as there are crates.
There is no per-crate cleaning now (at least none which would make sense).


matthiasbeyer:
@mario-kr You really should fix your git setup... it lists you as "mario-kr" in the commits. Care to re-do all the commits?

If you do, please rewrite the history to `git revert` bce217794f531e8a0d7de84c1d4be504caafeffa properly and then base your on this reverted commit.
Results in a way cleaner history.

Cargo.toml
```diff
@@ -0,0 +1,58 @@
...
+    "libimagstore",
+    "libimagstorestdhook",
+    "libimagtimeui",
+    "libimagtodo",
+    "libimagutil",
+]
+
+[[bin]]
+name = "imag-bin"
+path = "bin/src/main.rs"
```
matthiasbeyer: Do we need the `[[bin]]` section? As far as I can tell we do,... maybe we should ask the cargo people whether there is a way to not have it...

mario-kr: yes, we need it

mario-kr:
```sh
> git config --global --get user.name
mario
```
I think this "mario-kr" is from github. and yes, "mario" is what i want in my commits for now (or mario-kr, theres not much difference to me)

Cargo.toml
mario-kr: btw, i dont think workspaces are usually used like that...
I figured I'd put the `imag`-binary in there, as it is with what you in the end would call imag related binaries.

matthiasbeyer:
Okay. So should I merge or do you want to clean up what I mentioned above:

> If you do, please rewrite the history to git revert bce2177 properly and then base your on this reverted commit.
> Results in a way cleaner history.

mario-kr:
I'm not really confident in my rebasing skills, so I'd go with "merge".
Though I would also like to suggest to run the build once more to confirm the results.

matthiasbeyer:
Builds should be reproducible, but well... I restarted the build.

matthiasbeyer:
Rebuild succeeded.

PR #853, opened 2017-01-21
Make logger types public

matthiasbeyer:
This commit makes the logging functionality types from libimagrt public,
so the libimagruby implementation can re-use the ImagLogger type.

---

Extracted from #847 - @mario-kr I guess this is okay, isn't it? I'm not sure. We expose the logging utilities of libimagrt - still, Rust binaries should initialize the logger via the `Runtime` type.
This is just done so I can use the `ImagLogger` type in the Ruby library to initialize a logger from Ruby code.

### DIFF
--- a/libimagrt/src/lib.rs
+++ b/libimagrt/src/lib.rs
@@ -49,9 +49,9 @@ extern crate libimagutil;
 #[macro_use] extern crate libimagerror;
 
 mod configuration;
-mod logger;
 
 pub mod error;
+pub mod logger;
 pub mod runtime;
 pub mod setup;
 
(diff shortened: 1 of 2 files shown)


mario-kr:
Dunno.
I thought you'd provide the Runtime in rubyor something like that. Having a public ImagLogger doesnt sound "evil" or "wrong" to me, so it's okay imo.

matthiasbeyer:
I will provide the `Runtime` object for Ruby, yes. Though I will not make all functions accessible for Ruby, as this would be _way_ to complex IMHO.

Okay, I'll merge then.

matthiasbeyer:
BTW.: Awesome how fast travis finishes! :heart:

mario-kr:
> BTW.: Awesome how fast travis finishes! ❤️

Deleting stuff certainly helps with that 😄

PR #860, opened 2017-01-25
Travis opt

matthiasbeyer:
Lets see what happens if we move some things out of our travis configuration.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,16 +5,6 @@ rust:
 - beta
 - stable
 cache: cargo
-before_install:
-- |
-  c=$(git diff $(git merge-base master $TRAVIS_COMMIT)..$TRAVIS_COMMIT --name-only | cut -d "/" -f 1 | uniq)
-  if [[ "$c" == "doc" ]]; then
-      echo "Only changes in DOC, exiting 0"
-      exit 0
-  else
-    echo "Changes in other directories than ./doc"
-    echo "continue build..."
-  fi
 before_script:
 - |
   pip install 'travis-cargo<0.2' --user &&
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,12 +11,7 @@ before_script:
   export PATH=$HOME/.local/bin:$PATH
 script:
 - |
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -5,28 +5,13 @@ rust:
...
   export PATH=$HOME/.local/bin:$PATH
 script:
 - |
-  export RUSTFLAGS="-C codegen-units=2"
-  if [[ "$TEST_SUITE" == "binaries" ]]; then
-    make bin-test
-  else # $TEST_SUITE == "libraries"
-    make lib-test
-  fi
+  make
```
mario-kr: this doesnt call any tests afaik. you probably want them, though you might as well
```sh
make
make bin-test
make lib-test
```
build time for the last 2 should be near to 0.

btw, does calling `make bin-test` even build imag-bin? It should, but does it?

mario-kr:
[this](https://github.com/rust-lang/rust/blob/master/RELEASES.md#cargo) should explain our troubles with beta 😄

.travis.yml
matthiasbeyer: Good point(s)!

> btw, does calling `make bin-test` even build imag-bin? It should, but does it?

It should, yes.

matthiasbeyer: Nah, thinking about it again - it shouldn't. It should do `make imag-bin-test`, but there is no such thing, so ... everything is fine!

mario-kr:
it's kinda infuriating that you dont even `cargo check` before pushing stuff to travis.
btw, the reason it didnt fail on the `fix imag bin build script` branch is that `make bin-test` doesnt build `imag`, which you said was wanted (no `imag-bin-test` target is specified was your point I think)

matthiasbeyer:
> it's kinda infuriating that you dont even cargo check before pushing stuff to travis.

If someone else can compute things for me, why should I do it? :smile:

matthiasbeyer:
@mario-kr can I have an ACK or NACK?

mario-kr:
ACK

PR #870, opened 2017-01-25
Ruby tests

matthiasbeyer:
This branch adds tests for the `libimagruby` gem.

WIP.


matthiasbeyer:
Maybe @malept could provide review and/or suggestions here as well? :smile: 
I'm not sure whether the way I do the test setup is even sane...

matthiasbeyer:
Should wait on #871

malept:
Personally, I would go with unit tests instead of rspec, because:

* Less to install (minitest is bundled with Ruby)
* They're closer to what the built-in Rust testing framework looks like

But obviously, you should use whatever you're more comfortable with.

Regarding how the tests themselves should be structured, I would write them in the following way, to make sure that the Ruru glue code works as you expect:
* Can I use method X? (as opposed to "does it exist")
* Does method X translate the arguments/return value to/from the Rust function correctly?

_(This assumes that there are sufficient Rust unit tests for the Rust logic.)_

matthiasbeyer:
This is superceeded by #889 - closing therefor.

PR #871, opened 2017-01-25
Ruby build setup

### LINKED ISSUE #868: Add build setup for imag-ruby gem.
Add [thermite](https://github.com/malept/thermite) based build setup for imag ruby gem.

Thank you, @malept, for pointing me to thermite!

matthiasbeyer:
This PR introduces the build setup for the `libimagruby` Ruby gem.

@malept can you provide review (waiting on https://github.com/malept/thermite/issues/29 though) whether I'm doing sane things here?

---

Closes #868


matthiasbeyer:
Hmm. I'm not sure where build jobs 5 and 6 are coming from?

Also I don't know why my build times got from ~3 minutes up to 15 minutes. Just because of Ruby? Can't believe that!

.travis.yml
```diff
@@ -23,6 +35,7 @@ addons:
     - libzzip-dev
     - make
     - tree
+    - ruby
```
malept: I don't think this is necessary since you're using RVM to install.

.travis.yml
```diff
@@ -39,7 +52,17 @@ notifications:
     - chat.freenode.net#imag
     template:
     - "%{repository_name} (%{branch} @ %{commit} by %{author}): %{result}"
+
+matrix:
+  include:
+    - env: RBLANK_RUBY_VERSION=2.3.3
```
malept: I would suggest naming the environment variable IMAG_RUBY_VERSION 😄

matthiasbeyer: :smile:

matthiasbeyer:
I found a bug: Somehow I managed to make the header funny from the ruby scripts. Could be a bug in the store itself.

matthiasbeyer:
Okay, this works now as intended, besides the termite bug that doesn't recognize the cargo workspaces.

malept:
I built locally with my (now fixed) cargo-workspaces branch. Thermite builds a file named `lib/liblibimagruby.so` on Linux. 😰

matthiasbeyer:
:smile: well, if you build with cargo only it is named `liblibimagruby.so`, so that's fine... but the initial `lib/` is ... strange!

matthiasbeyer:
@malept now it works on travis, as far as I can see... at least what is done yet. I'm not sure how to proceed, what is to do now? Testing setup? Setup to build the gem? I'm not sure how to do this, can you guide me?

malept:
> Setup to build the gem?

I think you already do that when you call `rake thermite:tarball`. Or do you mean deploying to GitHub Releases?

In general I would investigate the test failure, merge this PR, rebase the test PR, and continue there.

malept:
Well, I should probably merge & release the cargo workspaces changes before you merge this, so you can get rid of the `Gemfile` change.

matthiasbeyer:
@malept okay. I wonder why it still fails with rust beta and ruby 2.3.3., though.

malept:
Yeah [this](https://travis-ci.org/matthiasbeyer/imag/jobs/196679217#L1878-L1885) doesn't make any sense, unless Rust/Cargo beta renamed the file for some reason.

matthiasbeyer:
this is caused by my `/libimagruby/test/` ruby files, which are manual tests.

I should probably remove them from the setup and use the infrastructure of bundler to test my stuff.

malept:
Yeah, I would be happy to consider adding more support around test running in Thermite if it's generic enough (there's currently `rake thermite:test` to run `cargo test` in the correct context).

matthiasbeyer:
> this is caused by my /libimagruby/test/ ruby files, which are manual tests.

It is still funny that it only caused a build failure on the combination rust beta + ruby 2.3.3 ... we'll see with the next build!

libimagruby/.gitignore
```diff
@@ -1,3 +1,5 @@
 .bundle
 Gemfile.lock
+lib
```
malept: Oh, I completely missed this. _(I should not be reading code this early in the morning...)_

You should have a file like https://github.com/malept/rusty_blank/blob/master/lib/rusty_blank.rb so you can load the Rust extension as a regular gem :smile:

matthiasbeyer: Sure I need this? I was able to load the binary itself from Ruby without thermite or fiddle (as in firing up an `irb` and simply `require`ing the `.so` file).

matthiasbeyer: Ah, okay.

matthiasbeyer:
Maybe this is because cargo-0.16-nighly stores the binary elsewhere?

malept:
Does it? News to me. Where's that?

matthiasbeyer:
I don't really know, but I assume because of the failing test...

malept:
I would do some debugging to see where the library actually got placed.

matthiasbeyer:
@malept It _seems_ like the build succeeds now, including the `@$(BUNDLE) exec rake thermite:tarball` part from the `Makefile` - can you confirm?

matthiasbeyer:
I guess the CI takes so long now because thermite builds in release mode, which rebuilds `libimagruby` and all dependencies.

@malept is there a way to ask thermite for a `debug` build? Because rebuilding all the things in CI only because thermite wants a release build is ... time consuming! I just re-checked what `thermite:build` does, it also builds in release mode as far as I can tell...

matthiasbeyer:
Note to self: I really should rebase and fixup the history here before merging!

malept:
> is there a way to ask thermite for a debug build? Because rebuilding all the things in CI only because thermite wants a release build is ... time consuming! I just re-checked what thermite:build does, it also builds in release mode as far as I can tell...

It should be in the rake task docs:

```shell
$ bundle exec rake -T | grep thermite:build
rake thermite:build    # Build or download the Rust shared library: CARGO_TARGET controls Cargo target
```

`CARGO_TARGET` defaults to `release`.

matthiasbeyer:
I will merge this now and we'll see how we can do the testing setup over in #889

PR #887, opened 2017-02-04
Remove unused imports

matthiasbeyer:
The beta compiler checks this... nice!

### DIFF
--- a/imag-link/src/main.rs
+++ b/imag-link/src/main.rs
@@ -34,7 +34,7 @@
 
 #[macro_use] extern crate log;
 extern crate clap;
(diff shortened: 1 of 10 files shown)


matthiasbeyer:
I wondered why master didn't fail on travis, but we didn't build master within the last 4 days, so we didn't run into this issue. Now, with Rust 1.16 beeing beta, we run into this "issue" ... well, I don't think it is an issue, we clean up our codebase here, I guess! At least in concern of imports and dependencies... so, that's really nice IMHO.

Bit of a pain to do, though, as I do not even have Rust 1.15 locally... let alone 1.16!

PR #888, opened 2017-02-04
Virtual workspaces do only have the workspace key

matthiasbeyer:
CC @mario-kr


matthiasbeyer:
This might also fix some problems with the Ruby build setup.

PR #892, opened 2017-02-05
Remove warnings

matthiasbeyer:
Remove warnings from a few crates.


matthiasbeyer:
I think the `itself` warnings are caused by ruru and I do not have any influence on them, as it seems. So I'm merging now.

PR #797, opened 2016-10-12
Simplify implementation of FoldResult and add tests

matthew-piziak:
@matthiasbeyer
#777

Issue #777, opened 2016-10-02
We use too much .fold(), there is libimagutil::iter::FoldResult!


matthiasbeyer:
The documentation has to be improved for `FoldResult`: https://matthiasbeyer.github.io/imag/libimagutil/iter/trait.FoldResult.html#tymethod.fold_result

But then we should _really_ use it:

``` bash
$ ag "fold\(" -l
libimagentrylink/src/internal.rs
libimagentrylist/src/listers/table.rs
libimagentrylist/src/listers/core.rs
bin/src/main.rs
libimagutil/src/iter.rs
libimaginteraction/src/ui.rs
libimagref/src/lister.rs
```

I assume almost _all_ cases can be refactored with `FoldResult`.

---

This is a beginner-task, but not if you're a complete rust beginner.

matthiasbeyer:
Note to self: Submitted to this-week-in-rust as easy issue that might be interesting for contributors.

matthew-piziak:
@matthiasbeyer Thanks for identifying easy issues and posting them to this-week-in-rust. I'll take on this refactor. Next update within 48h.

matthew-piziak:
@matthiasbeyer I have to admit I'm quite confused by the behavior of the `FoldResult` trait. It throws away the accumulator while `fold`ing, which makes it a degenerate fold, with the behavior (in Haskell) of `last . mapM`. In contrast, [`itertools::fold_results`](https://docs.rs/itertools/*/itertools/trait.Itertools.html#method.fold_results) is a true fold over `Result`s. In plain English, `FoldResult` returns the first error, unless there are none in which case the last success, unless there are none in which case the default.

I do not believe it is the case that all uses of `fold` in the codebase have vacuous accumulators that can be refactored with `FoldResult`.

matthew-piziak:
I wrote the following test cases while trying to understand `FoldResult`:

```rust
#[test]
fn test_fold_result_success() {
    let v = vec![1, 2, 3];
    let result: Result<usize, &str> = v.iter().fold_result(5, |item| Ok(*item));
    assert_eq!(result, Ok(3));
}

#[test]
fn test_fold_result_failure() {
    let v = vec![1, 2, 3];
    let result: Result<usize, &str> = v.iter().fold_result(5, |_| Err("failure"));
    assert_eq!(result, Err("failure"));
}

#[test]
fn test_fold_result_default() {
    let v: Vec<usize> = vec![];
[...]

PR #898, opened 2017-02-19
Simplify implementation of FoldResult and add tests

matthew-piziak:
Replacing https://github.com/matthiasbeyer/imag/pull/797

Fixing https://github.com/matthiasbeyer/imag/issues/777


matthew-piziak:
I will need to fix some things for CI to pass.

matthiasbeyer:
Indeed. Can you squash your two commits? The first one adds the imag-bookmark and the second one removes it again.

PR #913, opened 2017-03-02
Makefile use workspaces

mario-kr:
With the introduction of workspaces, some things became sort-of wrong in the Makefile. E.g. running `make clean` would clean `$IMAG/target` as many times as crates are specified in the the `members` section in `$IMAG/Cargo.toml`.
AFAIK, the same applies to `cargo outdated` and `cargo update`.
Please review/test to make sure I'm not wrong 😄


matthiasbeyer:
Nice. Although I would sit out what is about the new github terms of service now... because this project is LGPL licensed, and I wouldn't like to give away my rights.

matthiasbeyer:
I guess we're good to go.

PR #919, opened 2017-04-22
Prepare 0.3.0

matthiasbeyer:
Prep branch for the 0.3.0 release


matthiasbeyer:
As this is only a prepare branch (and we can have multiple prep branches of course), I'll merge this now.

PR #920, opened 2017-04-22
imag-mail/init

matthiasbeyer:
We can write things in Ruby now.

This will be one of my first targets to check whether this whole idea works out as intended.


matthiasbeyer:
After https://github.com/matthiasbeyer/imag/pull/946 is merged, we have to reconsider this PR.

The infrastructure which is already there is nice, although it uses `mailparse`, which is not that nice. We should re-use this PR to implement `email` as a library for `libimagemail`.

PR #925, opened 2017-04-28
Try to fix travis AGAIN

matthiasbeyer:
I'm getting really annoyed by this travis bullshit!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ before_install:
   rvm install "$IMAG_RUBY_VERSION"
   rvm use "$IMAG_RUBY_VERSION"
   ruby --version
-  pushd libimagruby
-  bundle install --jobs=3 --retry=3 --path=$TRAVIS_BUILD_DIR/libimagruby/vendor/bundle
-  popd
 
 before_script:
 - |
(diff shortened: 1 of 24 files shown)


matthiasbeyer:
@malept can you tell what's going on here? Somehow the update to rustc 1.17 and cargo 0.18 broke the mechanism that found the `libruby.so.2.3.3` before... that's fixed now but now I'm running into thie error shown here: https://travis-ci.org/matthiasbeyer/imag/jobs/226836806

I squashed the history now. Maybe https://github.com/matthiasbeyer/imag/pull/925/commits/6252f9b2dfb7ffc4eb38b9f53af0813f50032ed7 should be reverted to get this fixed... I don't know (will push a revert shortly, then we'll see whether this fixes the build)

malept:
I can take a look at your PRs this weekend.

matthiasbeyer:
You are awesome, thank you a lot!

matthiasbeyer:
@malept careful *ping* :heart:

matthiasbeyer:
We're moving away from the idea having a Ruby API, but instead we're exporting things via stdio, so we can have ruby scripts in the call-chain.

This gets closed therefor.

PR #926, opened 2017-05-02
libimagstore: StoreId tests

matthiasbeyer:
Because why not... :smile:


matthiasbeyer:
This is ready, but as the ruby stuff does not work on master, I cannot merge it. Waiting for #925

PR #930, opened 2017-05-03
Dependency: toml: 0.2.* -> 0.4.*

matthiasbeyer:
Update the `toml` dependency.


matthiasbeyer:
Ah, right... `toml-rs` now uses the serde framework... I guess this branch will live a whole lot longer than...

matthiasbeyer:
Doc: Had to bump the minimal rust version to 1.15 as toml-rs uses 1.15 as minimal version.

Issue #938, opened 2017-05-15
libimagentrydatetime


matthiasbeyer:
A library to set a header inside an entry with a date/time spec would be awesome.

The header could simply be:

```toml
[entry.date]
start = "2105-01-12"
end = "2015-02-12"
```

or something. 

### Desired features

- [x] Simple specification
    - [x] Date
    - [x] Date + Hour
    - [x] Date + Hour + Minute
    - [x] Date + Hour + Minute + Second
- [x] Range specification
    - [x] Date
    - [x] Date + Hour
    - [x] Date + Hour + Minute
    - [x] Date + Hour + Minute + Second
- [x] CRUD functionality for the entries (probably via the `chrono` crate)

---

I would go with the `entry` namespace in the header here, as we're probably developing even more such extension libraries for the `FileLockEntry` type, so we can keep these generic things in one namespace.

PR #941, opened 2017-05-19
libimagentrydate/init

matthiasbeyer:
Pushed so we can track the changes.

---

Missing:

- [x] Range support


matthiasbeyer:
We should also add some tests for the range support.

matthiasbeyer:
Related: https://github.com/matthiasbeyer/imag/issues/938

matthiasbeyer:
Should rename to `libimagentrydatetime`.

PR #942, opened 2017-05-21
Libimagapi/init

matthiasbeyer:
API library for writing APIs.

Basically, the intended architecture is this:

```
+-----------+
| $language |
+-----------+
      |
+------------------+
| libimag$language |  # Still required for translating data types from Rust <-> $language
+------------------+
      |
+------------+
| libimagapi |
+------------+
        |   |
        |   +-----------------+
        |                     |
+---------------+             |
| libimagentry* |             |
+---------------+             |
            |                 |
            |                 |
+------------------------------+
|          libimagstore        |
+------------------------------+
```

So, `libimagapi` attaches to the store and later also some (or all?) `libimagentry*` libs for building a interface for other programming languages.

This is needed because other programming languages do not have the concept of ownership and borrowing, so we cannot hand out the real objects to them.

This library adds caches and gives out handles to cache entries, therefor the real values are never passed to the external source.

This way, Ruby, Python, Lua, Lisp and also C, C++ or other programming languages can be attached to the imag core (which is also the plan, of course).


matthiasbeyer:
We're moving away from language-level bindings. We're offering a JSON API instead, which is way simpler implementation-wise.

Hence, this gets closed.

PR #943, opened 2017-05-21
libimagentrycategory/init

### LINKED ISSUE #939: libimagentrycategory
A nice thing would be a library to add a "category" header field to an entry.

```toml
[entry]
category = "food"
```

### Desired Features

- [x] CRUD functionality

### Possible features to be implemented later

- [ ] Find all categories of all entries which have one, compare the textual difference and suggest addaptions (spelling correction etc) interactively.

---

I would go with the `entry` namespace in the header here, as we're probably developing even more such extension libraries for the `FileLockEntry` type, so we can keep these generic things in one namespace.

matthiasbeyer:
Because we might need this.

Closes #939

### DIFF
--- a/libimagentrycategory/src/category.rs
+++ b/libimagentrycategory/src/category.rs
@@ -28,6 +28,7 @@ use libimagerror::into::IntoError;
 use error::CategoryErrorKind as CEK;
 use error::MapErrInto;
 use result::Result;
+use register::CategoryRegister;
 
 #[derive(Clone, Debug, Eq, PartialEq, Ord, PartialOrd)]
 pub struct Category(String);
@@ -49,6 +50,9 @@ impl Into<String> for Category {
 pub trait EntryCategory {
 
     fn set_category(&mut self, s: Category) -> Result<()>;
+
+    fn set_category_checked(&mut self, register: &CategoryRegister, s: Category) -> Result<()>;
+
     fn get_category(&self) -> Result<Option<Category>>;
 
     fn has_category(&self) -> Result<bool>;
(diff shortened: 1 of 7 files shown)


matthiasbeyer:
@TheNeikos @asuivelentine @mario-kr review pls!

matthiasbeyer:
btw, here we see the awesomeness of abstractions! With `toml-query` and `libimagerror` this library is only 200 lines of code, ... much less than if we'd need to write all the things by hand!

libimagentrycategory/src/category.rs
```diff
@@ -0,0 +1,100 @@
...
+        self.get_header_mut()
+            .insert(&String::from("category.value"), Value::String(s.into()))
+            .map_err_into(CEK::HeaderWriteError)
+            .map(|_| ())
+    }
+
+    fn get_category(&self) -> Result<Option<Category>> {
+        let res = self.get_header().read(&String::from("category.value"));
+        if res.is_err() {
+            let res = res.unwrap_err();
```
asuivelentine: Why you don't use if let?

matthiasbeyer: Because I'm dump. Thanks for the good catch!

matthiasbeyer:
Some nice ideas:

A category-register where categories are created by the user and the library checks before applying a category to an entry.

For this, the library should be like this:

* Creating a `CategorySetter` object, which would set a category "foo"
* The object checks whether this is a registered category, if not checks for similar categories (optionally of course) and proposes them to the user of the lib
* The user of the lib can now decide whether to do this or not and `entry.set_category(&category)` or whether to drop the Category object

* Reading from an Entry should be `entry.get_category() -> Result<Option<Category>>`

matthiasbeyer:
The above commits can now be used as a base to implement:

> For this, the library should be like this:
> 
> Creating a CategorySetter object, which would set a category "foo"
> 
> The object checks whether this is a registered category, if not checks for similar categories (optionally of course) and proposes them to the user of the lib
> 
> The user of the lib can now decide whether to do this or not (CategoryBuilder.build() -> Category) and entry.set_category(&category) or whether to drop the Category object
> 
> Reading from an Entry should be entry.get_category() -> Result<Option<Category>>

matthiasbeyer:
I am not sure whether the above idea gives us any improvements over the current interface.

Issue #830, opened 2016-10-31
Rewrite bash-script tests in Rust


matthiasbeyer:
As we have the test infrastructure in the imag-store, we can rewrite the tests in rust, IMO.

The following files (as of master at ) have to be rewritten:

```
./imag-link/tests/link-test.sh
./imag-link/tests/utils.sh
./imag-store/tests/001-create_test.sh
./imag-store/tests/utils.sh
./imag-store/tests/003-delete_test.sh
./imag-store/tests/002-retrieve_test.sh
./tests/utils.sh
```

I add the "easy" tag here, I think this is rather easy to do for a not-complete-rust-beginner.

matthiasbeyer:
The Makefile can be adapted as well, then.

irobert91:
Hi! I would like to work on this.

I have a question though:
I believe that it is expected for the tests to be pretty much just translated from bash to Rust. If so, what would be a good way for the prep and cleanup phases to be invoked for each test? I am referring to these:
https://github.com/matthiasbeyer/imag/blob/master/tests/utils.sh#L82
https://github.com/matthiasbeyer/imag/blob/master/tests/utils.sh#L49

I was thinking about using similar logic, of creating a `fn` for each test and then invoking them all, like `invoke_tests` does, but this doesn't feel Rusty. Would it be fine to just call Rust versions of `prepare_store_directory` and `reset_store` at the beginning and end of each test?

matthiasbeyer:
Hi. Awesome that you want to tackle this! :heart: 

First of, I'm pretty sure we don't want to use rust to call an imag binary, so that'd be a major difference from the bash tests. We want the tests to be in the binary itself, though I don't know whether you can have `#[test]`s in a rust _binary_.

The second thing:

You actually don't need `prepare_store_directory` and `reset_store` to be translated to Rust because the Store backend itself does automatically use an in-memory filesystem instead of the real filesystem when running the tests. So there won't be a possibility to test the hooks (for example the git hook), but I'm even not sure whether the hooks was a good idea. Maybe for debugging and simple assertions, but not for things like git tracking. So no problem leaving this out, actually.

Feel free to ask more questions! :+1:

irobert91:
I see. I thought that you would still want to have integration tests, similar to the current ones, so I was going to use `std::process::Command` to call the `imag-link` and `imag-store` binaries.
But as I understand now, you would rather prefer to have unit tests? Would the new tests cover the same parts as the bash tests and only that?
[...]

PR #947, opened 2017-05-30
imag-link/rewrite-tests WIP

irobert91:


irobert91:
@matthiasbeyer This would be the first step.
Naming/location of file and types might be way off. Please suggest better ones if needed.

Continuing the discussion from #830:
> We could refactor the Runtime::new function to not use the App::get_matches fn, but a wrapper which can be replaced (for testing purposes) by a mockup.

Where would be a good place for the mock `impl` of the (currently named) `CliSpec` trait? In the test mod of `imag-link/src/main.rs`?

> Having a file for the configuration is not that great, IMO. We could include a TOML directly in the build artifact for this, though. We would then need to refactor the config-loading process out.

Could you please go into a bit more detail? I don't know how this would be achieved.

matthiasbeyer:
So far I like what I see!

> Where would be a good place for the mock impl of the (currently named) CliSpec trait? In the test mod of imag-link/src/main.rs?

I guess that'd be the best idea: To mock the implementation of the `CliSpec` in every `mod tests {}` in each `imag-*` implementation. Maybe we can extract common functionality into a `libimagtestutils` later... but for now I'd go down the road of having mocks in each `imag-*/src/main.rs`.

> > TOML directly in the build artifact
>
> Could you please go into a bit more detail? I don't know how this would be achieved.

Is [this](https://doc.rust-lang.org/std/macro.include_bytes.html) enough explanation :smile: ?

If not:

We can have toml config files in the source tree but include them via the [`include_bytes!()` macro](https://doc.rust-lang.org/std/macro.include_bytes.html) into the build artifact. This can then be parsed before the tests and used by each test - or (if it isn't possible in another way) each test parses the toml again to use it.

irobert91:
It seems like configuration is currently [loaded from a file](https://www.rust-lang.org) all the time. Should I apply the same strategy as with the `CliSpec`? This would mean that a second parameter would be added to `Runtime::new`. Is there another way?

irobert91:
That would also eliminate the need for an actual toml file.
On the other hand, now that I look at [`FileAbstraction`](https://github.com/matthiasbeyer/imag/blob/master/libimagstore/src/file_abstraction.rs#L53), I see that I could `write_file_content`, but even if this would be done only once somehow, the `Configuration` would be loaded and parsed with each `Runtime::new`, unless we use only one `Runtime` for all tests, defined with `lazy_static` maybe. Is this doable or even a good solution?

irobert91:
I believe that the last part (making use of `FileAbstraction` directly) would be possible by making the [`mod`](https://github.com/matthiasbeyer/imag/blob/master/libimagstore/src/lib.rs#L58) public. Do we have a reason not to? Is it supposed to be used only inside `libimagstore`?
What do you think is the best way to approach this?

Sorry for adding several inconsistent comments. I am writing my thought process as I'm looking through the code.

matthiasbeyer:
> It seems like configuration is currently loaded from a file all the time

You linked to the rust-lang website here - I guess that's not what you intended to do, right? :smile:

>  Should I apply the same strategy as with the CliSpec? This would mean that a second parameter would be added to Runtime::new. Is there another way?

I'd say yes, although adding another parameter to the Runtime is not really nice. Is there a way to do this via a template parameter which is filled with a default type? For example [the rust stdlib HashMap](https://doc.rust-lang.org/std/collections/struct.HashMap.html) implementation has a default parameter `RandomState` - we could use this technique to have a standard template parameter which loads the default configuration files (basically what the current implementation of the `Runtime` type does). In the tests, we can override this parameter to load another config,... possibly from in-memory TOML. Do you understand what I mean?

> I see that I could `write_file_content` but even if this would be done only once somehow,

Accessing this store-backend function from a (rather high-level) test is a no-go, IMO.

> the Configuration would be loaded and parsed with each Runtime::new, unless we use only one Runtime for all tests, defined with lazy_static maybe. Is this doable or even a good solution?

No, having one Runtime for all tests is nt a good solution. Loading a new Runtime object for each test would be the best way as far as I can see.

> I believe that the last part (making use of FileAbstraction directly) would be possible by making the mod public. Do we have a reason not to? 

Yes, this is a backend module of the store. We shouldn't need to access these details from a high-level test.

> Sorry for adding several inconsistent comments. I am writing my thought process as I'm looking through the code.

That's completely fine and I'm more than happy to answer your questions, really! :smile: :+1: 

---

So from what I see we are discussing things here we really should not have to discuss.
By now, I have the following things in mind (and bear with me, I'm also not certain whether we want to implement these tests like this... I'm also just writing down thoughts).

* We want to emulate a commandline call and then check whether the functionality we called is executed correctly. That is, if we want to `imag link --from a --to b`, we have to 
    1. create two files (`a` and `b`) in the store
    2. call `imag-link` functionality with the above parameters 
    3. check whether the headers of the files look like we expect them to.

For this, we need to be able to access the `Store` object (which we can, if we have a `Runtime` object).
So we need to create a `Runtime` object. We need to pass a commandline call. This is implemented in this PR already - wonderful!

Now we need to be able to get a `libimagrt::configuration::Configuration` into the `Runtime` - as far as I can see, this could also be implemented into the `CliSpec` type you already implemented (which then should be renamed to something else, of course). Basically a function that returns a `Result<Configuration>` object. The implementation for the `App` from clap could then do what the current `Runtime` does, the mocked implementation could simply load a in-memory TOML and build the `Configuration` object from that. Please note that `Configuration::new` must be changed for that, but as that function is not supposed to be called by a user of `libimagrt` directly, I'm perfectly fine with it!

---

Please not that I'm on vacation from tomorrow, 02.06.2017 until 05.06.2017 (evening). So I cannot respond in this time. I hope maybe @mario-kr or @asuivelentine can help you with implementation details or interpretation of the above if you did not understand single aspects.

---

I would like to have a short :+1: or :-1: from @TheNeikos on the above writeup, whether this would be sane to be implemented like that or not.

irobert91:
>You linked to the rust-lang website here - I guess that's not what you intended to do, right? :smile:

Indeed, I wanted to link to `fetch_config`: https://github.com/matthiasbeyer/imag/blob/master/libimagrt/src/configuration.rs#L220

>>Should I apply the same strategy as with the CliSpec? This would mean that a second parameter would be added to Runtime::new.
>
>Is there a way to do this via a template parameter which is filled with a default type?

I have tried in [various ways](https://play.rust-lang.org/?gist=0c1052bf00ff2e46f5a83b5f60491af3&version=stable&backtrace=0), but it doesn't seem to work.

> as far as I can see, this could also be implemented into the CliSpec type you already implemented (which then should be renamed to something else, of course)

I think this is a good solution, having a "runtime building" object that provides both command line arguments and configuration. Although maybe it would be cleaner to add a separate trait for config creation and `impl` it for `clap::App` like you described.

On a different topic, when I format the code, rustfmt changes some parts of the old code as well, not only the new one that I'm writing. Is there a specific rustfmt config that I could use to avoid this? I couldn't find one in the repo.

irobert91:
I am getting this output when creating the `Runtime` mock (using `/tmp` as `--rtp`):
```
ERROR[   1]: Could not find repository from '/tmp/store'

ERROR[   1]: Could not find repository from '/tmp/store'

ERROR[   1]: Could not find repository from '/tmp/store'
[imag][WARN]: Registering git hook with store failed
[imag][WARN]: Registering git hook with store failed
[imag][WARN]: Registering git hook with store failed
```

I have tried disabling the hooks through config:
```toml
[store]
implicit-create = true

store-unload-hook-aspects  = []

pre-create-hook-aspects    = []
post-create-hook-aspects   = []

pre-move-hook-aspects      = []
post-move-hook-aspects     = []

pre-retrieve-hook-aspects  = []
post-retrieve-hook-aspects = []

pre-update-hook-aspects    = []
post-update-hook-aspects   = []

pre-delete-hook-aspects    = []
post-delete-hook-aspects   = []

[store.aspects.debug]
parallel = false

[store.aspects.vcs]
parallel = false

[store.hooks.stdhook_debug]
aspect = "vcs"

[store.hooks.stdhook_git_update]
aspect = "vcs"

enabled = false

[store.hooks.stdhook_git_update.commit]
enabled = false

[store.hooks.stdhook_git_delete]
aspect = "vcs"

enabled = false

[store.hooks.stdhook_git_delete.commit]
enabled = false

[store.hooks.stdhook_git_storeunload]
aspect = "vcs"

enabled = false

[store.hooks.stdhook_git_storeunload.commit]
enabled = false
```
But this has no effect. Is it doable through configuration?

Otherwise I guess we could have another method on the `cli_spec` parameter in `Runtime::new` to disable the hooks with a check somewhere around [these lines](https://github.com/matthiasbeyer/imag/blob/master/libimagrt/src/runtime.rs#L174).

Or should the hooks even be disabled? Is there another way?

@mario-kr, @asuivelentine could you give me some input on this?

irobert91:
Most probably it can be more idiomatic and cleaner, but this is what I have so far. @matthiasbeyer I'm looking forward to your feedback and suggestions!

libimagrt/src/configuration.rs
```diff
@@ -275,3 +274,23 @@ fn fetch_config(rtp: &PathBuf) -> Result<Value> {
         .map(|inner| Value::Table(inner.unwrap()))
         .ok_or(ConfigErrorKind::NoConfigFileFound.into())
 }
+
+pub trait GetConfiguration {
+    fn get_configuration(rtp: &PathBuf) -> Result<Configuration> {
+        fetch_config(rtp).map(Configuration::new)
```
matthiasbeyer: As far as I can see, we only need the `fetch_configuration` function in the impl of `GetConfiguration` - so we could move it here, right?

libimagrt/src/runtime.rs
```diff
@@ -74,20 +77,22 @@ impl<'a> Runtime<'a> {
...
 
-        let matches = cli_spec.clone().get_matches();
+        let matches = cli_app.clone().matches();
 
         let is_debugging = matches.is_present("debugging");
         let is_verbose   = matches.is_present("verbosity");
         let colored      = !matches.is_present("no-color-output");
 
-        Runtime::init_logger(is_debugging, is_verbose, colored);
+        if C::enable_logging() {
```
matthiasbeyer: Why `C::enable_logging()` and not `cli_app.enable_logging()`?

libimagrt/src/runtime.rs
```diff
@@ -111,7 +116,7 @@ impl<'a> Runtime<'a> {
         let configpath = matches.value_of("config")
                                 .map_or_else(|| rtp.clone(), PathBuf::from);
 
-        let cfg = match Configuration::new(&configpath) {
+        let cfg = match C::get_configuration(&configpath) {
```
matthiasbeyer: Same here: Why `C` and not `cli_app`?

libimagrt/src/runtime.rs
```diff
@@ -169,22 +174,26 @@ impl<'a> Runtime<'a> {
...
-
-            for (hook, aspectname, position) in hooks {
-                if let Err(e) = store.register_hook(position, &String::from(aspectname), hook) {
-                    if e.err_type() == StoreErrorKind::HookRegisterError {
-                        trace_error_dbg(&e);
-                        warn!("Registering git hook with store failed");
-                    } else {
-                        trace_error(&e);
-                    };
+            if C::enable_hooks() {
```
matthiasbeyer: ... and here. :smile:

libimagrt/src/spec.rs
```diff
@@ -0,0 +1,27 @@
+use std::io::Write;
+
+use clap::{App, ArgMatches, Shell};
+
+pub trait CliSpec<'a> {
```
matthiasbeyer: Could you add some documentation on the purpose of this trait?

imag-link/src/main.rs
```diff
@@ -310,3 +310,176 @@ fn list_links_for_entry(store: &Store, entry: &mut FileLockEntry) {
...
+            self.inner.get_matches_from(self.args)
+        }
+    }
+
+    impl<'a> GetConfiguration for MockLinkApp<'a> {
+        fn get_configuration(_: &PathBuf) -> ConfigResult<Configuration> {
+            ::toml::de::from_slice(TOML_CONFIG_FILE)
+                .map_err(|_| {
+                             ConfigErrorKind::TOMLParserError.into()
+                         })
```
matthiasbeyer: You don't need the scope `{...}` here! Please remove it and put the whole mapping into one line.

Also: Why aren't you using `map_err_into(ConfigErrorKind::TOMLParserError)` (maybe you didn't know it existed,... so here's your hint)? :smile:

imag-link/src/main.rs
```diff
@@ -310,3 +310,176 @@ fn list_links_for_entry(store: &Store, entry: &mut FileLockEntry) {
...
+        entry.get_content_mut().push_str(DEFAULT_ENTRY);
+
+        Ok(id)
+    }
+
+    fn get_entry_links<'a>(entry: &'a FileLockEntry<'a>) -> Option<&'a Value> {
+        entry
+            .get_header()
+            .get("imag")
+            .and_then(|ih| ih.get("links"))
```
matthiasbeyer: This _could_ be minified using `toml-query` (the crate) - if you don't mind...

matthiasbeyer:
> > > Should I apply the same strategy as with the CliSpec? This would mean that a second parameter would be added to Runtime::new.
> >
> > Is there a way to do this via a template parameter which is filled with a default type?
>
> I have tried in various ways, but it doesn't seem to work.

As far as I can see, the [ways you mentioned](https://play.rust-lang.org/?gist=0c1052bf00ff2e46f5a83b5f60491af3&version=stable&backtrace=0) worked? At least the linked playground compiles and runs for me...

> I think this is a good solution, having a "runtime building" object that provides both command line arguments and configuration. Although maybe it would be cleaner to add a separate trait for config creation and impl it for clap::App like you described.

The way it is implemented right now looks really nice!

> On a different topic, when I format the code, rustfmt changes some parts of the old code as well, not only the new one that I'm writing. Is there a specific rustfmt config that I could use to avoid this? I couldn't find one in the repo.

You don't need to format the code. Just try to keep lines below the 100 chars range and don't do crazy unreadable things like one-liners whereever possible...

```rust
if foo.bar().baz() { bla.call(Foo::new()) } else { blubb.blah() }
```

is ugly. Prefer readable code and I'm fine (the current patchset looks really good IMO).

> Config, Hooks, things don't work

Hm. That's because the store expects a git repository initialized where the `/tmp/store` is. I'm working on a patchset to remove the hook support entirely, because it is rather useless (targeting #950 with this patchset then).

If you'd wait with your patchset, I guess mine (removal of the hook support) should be ready within the next days,... you would need to rebase this PR then and maybe remove some things, if you don't mind that...

---

So far: :heart: :smile: :+1:

libimagrt/src/configuration.rs
irobert91: You mean replace `get_configuration` with `fetch_config(uration)`? Or keep the name and just move the function body?

libimagrt/src/runtime.rs
irobert91: When I first wrote them, all the methods weren't static, but then I thought that they have nothing to do with `self`, so I made them static. I will change them if you prefer them that way.

irobert91:
>As far as I can see, the ways you mentioned worked? At least the linked playground compiles and runs for me...

I linked to a version that compiles, but `fn main` body looks like this:
```rust
    // this won't work
    // let conf = Runtime::new();
    let conf = Runtime::<Config>::new();
    let another_conf = Runtime::<AnotherConfig>::new();

    println!("{:?}", conf);
    println!("{:?}", another_conf);
```

If the second line, the one that doesn't explicitly set the type parameter, is uncommented, the code won't compile. Anyway, I think that the current implementation is good enough and we don't need to pursue this track.

> If you'd wait with your patchset, I guess mine (removal of the hook support) should be ready within the next days,... you would need to rebase this PR then and maybe remove some things, if you don't mind that...

Sure! Then I could remove the `enable_hooks` method from `InternalConfiguration`. I will have to rebase the code when everything is done anyway.

libimagrt/src/configuration.rs
matthiasbeyer: Replace the body of `get_configuration` with the body of `fetch_config` and remove the latter therefor.

libimagrt/src/runtime.rs
matthiasbeyer: Yes, that'd be nice... I guess it does not make a difference for the compiler, but I find the `cli_app.enable_logging()` variant much more readable!

matthiasbeyer:
I guess we're overthinking something here. Why won't we use [this](https://gist.github.com/anonymous/ebd0ef92a2b18abfe39da779273be458) instead of default template parameters?

> Sure! Then I could remove the enable_hooks method from InternalConfiguration. I will have to rebase when everything is done anyway.

Awesome, ... I'm [in the process of getting things ready](https://github.com/matthiasbeyer/imag/pull/951).

But not today anymore... tomorrow! :night_with_stars:

matthiasbeyer:
:heart: This looks so awesome! I can't wait to merge this! I'm really looking forward writing tests with this setup! Finally removing the bash stuff, yay!

matthiasbeyer:
FYI: #951 has landed.

irobert91:
>I guess we're overthinking something here. Why won't we use this instead of default template parameters?

I agree, another constructor that sets the config feels like a cleaner solution.
Will make the needed changes.

matthiasbeyer:
I'm sorry to tell you that the travis reports "all good" because the binaries are not yet in the test execution. To fix this, feel free to cherry-pick 21cc4e97750315142f49eab01c2be2ed95044b82 onto your branch.

I guess we're running in more problems, anyways. I'm in the process of writing #943 and I just found out that running tests in a lib crate results in the `libimagstore` build with the FS-backend, not the in-memory one as I expected. I guess this has to be sorted out first.

The backends are currently switched based on compiletime `[cfg(test)]` and `[cfg(not(test))]` - which works well when running `libimagstore` tests, but not when running tests from outside of this very crate.
The backends should be setable with dependency-injection, I guess. This can even be hidden behind a feature-flag, I guess.

irobert91:
I encountered a little issue while implementing tests with `Runtime::with_configuration`. `Configuration` fields are all private, so it isn't possible to create a literal one like
```rust
Configuration {
    verbosity: false,
    editor: None,
    editor_opts: "".to_owned(),
    config: value,
}
```
I also reverted the `Configuration::new` signature to the original
```rust
pub fn new(rtp: &PathBuf) -> Result<Configuration>
```
because the new version wasn't needed anymore given the new `Runtime` constructor.
So now the first thing I can think of is to create a new constructor for `Configuration` also:
```rust
pub fn with_values(toml_config: Value, verbosity: bool, editor: Option<String>, editor_opts: String) {
    ...
}
```
Do you have a better idea?

matthiasbeyer:
I do not see a better way right now.

irobert91:
Do you think it would be enough to only set the value of the `config` field through the constructor and have empty defaults for the rest?

matthiasbeyer:
I guess the best idea would be to have a constructor which takes the config toml object and more builders like this: `with_verbosity(self, b: boolean) -> Self` to be able to adjust them test-wise.

imag-link/src/main.rs
```diff
@@ -368,20 +365,18 @@ version = \"0.3.0\"\
...
-    }
-
     impl<'a> InternalConfiguration for MockLinkApp<'a> {
         fn enable_logging(&self) -> bool {
             false
         }
     }
 
+    fn generate_test_config() -> Option<Configuration> {
+        ::toml::de::from_str("[store]\nimplicit-create=true")
```
irobert91: I thought that we don't need an actual file

libimagrt/src/configuration.rs
```diff
@@ -70,21 +70,33 @@ impl Configuration {
...
+                config: cfg,
+                verbosity: verbosity,
+                editor: editor,
+                editor_opts: editor_opts,
+            }
+        })
+    }
+
+    /// Get a new configuration object built from the given toml value.
+    pub fn with_value(value: Value) -> Configuration {
```
irobert91: Ended up just adding a constructor that takes a `toml::Value` because the other fields can be extracted out of it.

libimagrt/src/runtime.rs
```diff
@@ -56,26 +56,61 @@ impl<'a> Runtime<'a> {
...
+
+    /// Builds the Runtime object using the given `config`.
+    pub fn with_configuration<C>(mut cli_app: C, config: Option<Configuration>) -> Result<Runtime<'a>, RuntimeError>
+        where C: Clone + CliSpec<'a> + InternalConfiguration
+    {
+        use std::io::stdout;
+
+        use clap::Shell;
+
+        let matches = cli_app.clone().matches();
```
irobert91: `matches` will be parsed twice when this method is called from `Runtime::new`, which I believe is not ideal. Let me know if you have a better solution.

irobert91: Maybe we could have a third, private constructor taking `ArgMatches` as an argument, called by the other 2 existing ones. What do you think @matthiasbeyer?

matthiasbeyer: Yes, this seems to be the way to go.

irobert91:
In order to make use of the nice changes you made in #962...
I was thinking to add a new method to `InternalConfiguration`: `use_inmemory_fs` and use a different backend to instantiate the `Runtime` store inside `Runtime::_new`, but as far as I can tell this means that `libimagstore::file_abstraction` needs to be made public. Am I missing something? Is there another way?

matthiasbeyer:
The types from the `file_abstraction` module are exported in the `store` module!

irobert91:
I'm not sure if I'm doing something wrong, but if I run the tests using the in-memory backend, this error occurs:
```
thread 'tests::test_link_modificates' panicked at 'called `Result::unwrap()` on an `Err` value: StoreError { err_type: GetCallError, cause: Some(StoreError { err_type: RetrieveCallError, cause: Some(StoreError { err_type: MalformedEntry, cause: None, custom_data: None }), custom_data: None }), custom_data: None }'
```

Whereas if I use the file system, everything goes well.
Are there some differences between the two that I have to take into account when testing?

matthiasbeyer:
That might be a bug in the backend. You should only need to set the backend when writing the tests, the rest of the code should be exactly the same as with the filesystem backend. I'm on mobile right now and therefor cannot investigate. I hope I have time for this tomorrow!

matthiasbeyer:
And [I have another patch for you](https://github.com/matthiasbeyer/imag/commit/d3c612c97771997364f55027532d4782a44bf6b2) which fixups [my broken commit](1f95c6a620386bd3c5df85fbbb15327db1726cae) as there was a typo in it.

Sorry for the hassle!

matthiasbeyer:
Any progress recently? :smiley:

irobert91:
Sorry for the delayed reply. I believe I made all the (close to final) changes.
Please take a look and let me know if anything still needs to be modified.

matthiasbeyer:
Hm, this errors but not because of your changes AFAICS. I will investigate later.

matthiasbeyer:
I guess with d2b1d82015925b0ba8c92c7abf2da626961e8bd6 included this should finally compile... 

If you're okay with it I can include and cleanup this PR for you!

irobert91:
Sorry, I haven't had much time lately.
I think I can make some time in the near future for final edits if you'll tell me what is needed, but of course, feel free to make changes yourself.

matthiasbeyer:
@irobert91 this is completely okay! I am in exams phase, so I also have not much time. I will see whether I can find some time for the final cleanup! Thank you very much for your effort! :heart:

matthiasbeyer:
@irobert91 I [updated and fixed the branch](https://github.com/matthiasbeyer/imag/tree/pr/947) and travis runs the tests currently: https://travis-ci.org/matthiasbeyer/imag/builds/251428319

I'm not sure how to push to this PR - did you allow edits from maintainers in the initial message for this PR? If not, please do so, so I can re-use this PR for the cleanup without having to open another PR to superceed this one! :smile:

libimagrt/src/runtime.rs
```diff
@@ -54,94 +56,118 @@ impl<'a> Runtime<'a> {
...
                                 .map_or_else(|| rtp.clone(), PathBuf::from);
 
-        debug!("RTP path    = {:?}", rtp);
-        debug!("Store path  = {:?}", storepath);
         debug!("Config path = {:?}", configpath);
 
-        let cfg = match Configuration::new(&configpath) {
-            Err(e) => if e.err_type() != ConfigErrorKind::NoConfigFileFound {
+        let config = match Configuration::new(&configpath) {
+            Err(e) => if e.err_type() != ConfigErrorKind:: NoConfigFileFound {
```
matthiasbeyer: Note to self: whitespace fail in this line.

matthiasbeyer:
Another note to self: we can update the toml-query dependency here before merging or after. I have no strong feelings.

I hope to get the final cleanup done next weekend or shortly after (exams are done on Friday, but party is until Sunday :-P )

PR #952, opened 2017-06-05
Remove hooks impl link consistency check in libimagentrylink

matthiasbeyer:
Part of #951


matthiasbeyer:
This was just hacked together, I don't think this will compile :smile:

PR #962, opened 2017-06-07
Libimagstore/fs memory backend as dependency injection

matthiasbeyer:
As described in https://github.com/matthiasbeyer/imag/pull/947 and discovered in the process of implementing https://github.com/matthiasbeyer/imag/pull/943, we have a problem:

(from #947:)

> I'm in the process of writing #943 and I just found out that running tests in a lib crate results in the libimagstore build with the FS-backend, not the in-memory one as I expected. I guess this has to be sorted out first.
> 
> The backends are currently switched based on compiletime [cfg(test)] and [cfg(not(test))] - which works well when running libimagstore tests, but not when running tests from outside of this very crate.
The backends should be setable with dependency-injection, I guess. This can even be hidden behind a feature-flag, I guess.

This tries to solve this problem by implementing a dependency-injection for the backend.

### DIFF
--- a/libimagstore/src/file_abstraction.rs
+++ b/libimagstore/src/file_abstraction.rs
@@ -134,8 +134,9 @@ mod fs {
                         .map_err_into(SEK::FileNotSeeked));
 
                     let mut s = String::new();
-                    f.read_to_string(&mut s);
-                    Ok(s)
+                    f.read_to_string(&mut s)
+                        .map_err_into(SEK::IoError)
+                        .map(|_| s)
                 },
                 FSFileAbstractionInstance::Absent(ref p) =>
                     (try!(open_file(p).map_err_into(SEK::FileNotFound)), p.clone()),
@@ -143,10 +144,12 @@ mod fs {
             *self = FSFileAbstractionInstance::File(file, path);
             if let FSFileAbstractionInstance::File(ref mut f, _) = *self {
                 let mut s = String::new();
-                f.read_to_string(&mut s);
-                return Ok(s);
+                f.read_to_string(&mut s)
+                    .map_err_into(SEK::IoError)
+                    .map(|_| s)
+            } else {
+                unreachable!()
             }
-            unreachable!()
         }
 
         /**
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
If this works, we need a HUUUUGE commit message and documentation on how this works.

It does not yet, obviousely. @TheNeikos I'd love to hear your suggestions!

matthiasbeyer:
This implementation now compiles, but it does not work. I do not know why and I fail to debug it. @TheNeikos any idea? Please?

matthiasbeyer:
The fun thing is: The code does not even get into the FS backend codebase when running the tests that fail here.

matthiasbeyer:
HOLY SHIT IT WORKS :fireworks: :fireworks: :fireworks: 

@TheNeikos please, please, please review this for me! I'd love to hear your opinion and suggestions!

matthiasbeyer:
Of course the commit messages have to be rewritten and documentation is missing - I'm doing this tomorrow. The tests running on my machine will keep me awake all night long I guess :smile:

matthiasbeyer:
Ah, @irobert91 I'd love to read your review on this as well! :smile: 

And maybe also @asuivelentine and @mario-kr ! The more review, the better. Especially with this code!

libimagstore/src/file_abstraction.rs
```diff
@@ -17,112 +17,200 @@
...
+use error::StoreError as SE;
+
+pub use self::fs::FSFileAbstraction;
+pub use self::fs::FSFileAbstractionInstance;
+pub use self::inmemory::InMemoryFileAbstraction;
+pub use self::inmemory::InMemoryFileAbstractionInstance;
 
 // TODO:
 // This whole thing can be written better with a trait based mechanism that is embedded into the
 // store. However it would mean rewriting most things to be generic which can be a pain in the ass.
```
irobert91: You've already merged it, but isn't this `TODO` solved with this PR?

matthiasbeyer: Whups, thanks for pointing out!

PR #966, opened 2017-06-09
imag-timetrack / libimagtimetrack

### LINKED ISSUE #937: imag-timetrack / libimagtimetrack
# libimagevent

Library to track events

We want to add dates / times to entries. We should do this as extension-trait on `FileLockEntry` to add the appropriate headers.

```rust
pub trait EventEntry {
    fn make_event_now(&mut self, accuracy) -> Result<()>;
    fn make_event(&mut self, timespec, accuracy) -> Result<()>;
    // ... ?
}
```

# imag-event

We want a frontend where we can specify:

* Accuracy (day, hour, minute, second)
* Comment (free text)
* Specifying the `when` - now or `<time spec>`
* Creating events (stored in `/event/<time>-<hash>.ext` probably - we need a decision here)
* start- and end-time (ranges)

## Example usage 

```bash
imag event start +talk +work # "+talk" and "+work" are tags - possible without "+" prefix?
imag event stop +talk # started tags are now "work"
imag event start +programm # started tags are now "work", "programm"
imag event stop # no started tags
imag event continue # started tags are now "work", "programm"
```

## Data layout

The store-internal data layout is important for the flexibility here.

The basic idea is that "tags" are normal tags in the `libimagentrytag` sense.
When creating an event (`start`ing it) with a tag `"exampletag"`, a new entry would be created in `/event/2017/06/01/13:34:12.23456-exampletag.ext` containing the following header:

```toml
tags = [ "exampletag" ]
event.start = "2017-06-01T13:34:12.23456"
# event.end not set
```

When closing the tag `"exampletag"` (`stop`ing it), the `event.end` field would be set to the current timestamp.

### Advantages of data layout

* Each event is in one entry in the store, this enables
    * More meta-data can be added (category, text)
    * Entries can be linked

### Disadvantages of data layout

* "Tagging" events is not possible, as they work with normal imag-tags.

### Alternatives for the data layout

* The path could also include the tag name as a top-level collection: `/event/exampletag/2017/06/01/13:34:12.23456.ext`
    * Would improve search-times for specific tags
    * Would improve "stopping" tags
    * Would worsen the search times for "continue the last stopped thing"
* The "event tags" could be stored in a sub-field of the `[event]` section, so it wouldn't infer with the normal tagging system
    * Could be abstracted and controlled by a setting, or even a prefix for the tags (`+tag` for event-tag, `tag` for imag-tag - if clap gives us the power to have this)

## Future features

Future features that may be implemented on this (everything via configuration ofc):

* auto-denying of certain tags on weekends, holidays, etc.
* time-chart/burndown functionality
* denying certain tags at certain times (no work tags after 18:00 for example)

matthiasbeyer:
Closes #937

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ members = [
     "libimagref",
     "libimagrt",
     "libimagstore",
+    "libimagtimetrack",
     "libimagtimeui",
     "libimagtodo",
     "libimagutil",
--- a/doc/src/05100-lib-timetrack.md
+++ b/doc/src/05100-lib-timetrack.md
@@ -50,3 +50,17 @@ The `end` field is, of course, only set if the event already ended.
 The library uses the `libimagentrydatetime::datepath::DatePathBuilder` for
 building `StoreId` objects.
 
+The library offers two central traits:
+
+* `TimeTrackStore`, which extends a `Store` object with functionality to
+  create `FileLockEntry` objects with a certain setting that is used for
+  time-tracking, and
+* `TimeTracking`, which extends `Entry` with helper functions to query the
+  entry-metadata that is used for the time tracking functionality
+
+The library does _not_ provide functionality to implement `imag-timetrack` or
+so, as the core functionality is already given and the commandline application
+can implement the missing bits in few lines of code.
+
+Aggregating functionality might be provided at a later point in time.
+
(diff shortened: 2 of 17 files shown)


doc/src/05100-lib-entryevent.md
```diff
@@ -0,0 +1,39 @@
...
+* a start date
+* an end date
+* a tag
+
+by default and might be extended with more header fields as one likes.
+
+The header of a event "work" entry looks like this:
+
+```toml
+tags = [ "work" ]
```
matthiasbeyer: This should be in the 'event' namespace in the header!

PR #969, opened 2017-06-17
Add iterator helpers in libimagstore

matthiasbeyer:
Some iterator types for the store.

Should be carefully evaluated whether this is actually necessary.


matthiasbeyer:
I thought I'd need this for another lib, but it turns out I can compile the other library with this patch, too. So I'll leave this unmerged.

PR #976, opened 2017-06-18
Imag store dump

matthiasbeyer:
This uses the new backend-replacing mechanism to dump the store to stdout.

It does currently not work as expected.

To reproduce:

```bash
../target/debug/imag-store --config ../imagrc.toml --override-config store.implicit-create=true --rtp /tmp/ dump
```

Outputs:

```
Config file parser error:
ERROR[   1]: a custom error
[imag][WARN]: No config file found.
[imag][WARN]: Continuing without configuration file

ERROR[   1]: StoreId 'id' part is absolute (starts with '/') which is not allowed

ERROR[   1]: StoreId 'id' part is absolute (starts with '/') which is not allowed
{"version":"0.3.0","store":{}}
ERROR[   1]: JSON error
ERROR[   2]: IO Error -- caused:
ERROR[   3]: Could not instantiate
[imag][ERROR]: Loading Store IO backend failed
```

So the store gets actually dumped to stdout as expected, though the errors (and the exit code) are not as expected.

@mario-kr You are always enthusiastic in debugging such things - do you care? :smile:

### DIFF
--- a/imag-store/src/dump.rs
+++ b/imag-store/src/dump.rs
@@ -39,7 +39,7 @@ pub fn dump(rt: &mut Runtime) {
         });
 
     if let Ok(_) = cachingres {
-        if let Err(_) = rt.store_backend_to_stdio().map_err_trace() {
+        if let Err(_) = rt.store_backend_to_stdout().map_err_trace() {
             error!("Loading Store IO backend failed");
             exit(1);
         }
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@mario-kr I fixed the errors and submitted them in #977 and #978 as cherry-picks - after these two PRs are merged I will rebase this branch which will result in some commits vanishing.

This works now, artificially tested.

PR #988, opened 2017-07-04
Imag link/fix panic

### LINKED ISSUE #981: imag-link remove panics on unwrap
All following have the following options set after `imag`:

```
--rtp /tmp/ --config ./imagrc.toml --override-config store.implicit-create=true 
```

```
$ ./target/release/imag-store create -i example entry -c "EXAMPLE!"
$ ./target/release/imag-store create -i example-1 entry -c "EXAMPLE!"
$ ./target/release/imag-link internal add --from example --to example-1
$ ./target/release/imag-link internal remove -f example -t example-1
# panics
```

matthiasbeyer:
Closes #981 

Although this fixes the panic, we have another issue now. The entry looks like this after the removal of the link:

```
---
[imag]
links = []
version = "0.3.0"
---
EXAMPLE!--
EXAMPLE!
```

which is not valid.

I'm not sure whether this is a bug in the store...

### DIFF
--- a/imag-link/src/main.rs
+++ b/imag-link/src/main.rs
@@ -55,6 +55,7 @@ use libimagstore::store::FileLockEntry;
 use libimagstore::store::Store;
 use libimagerror::trace::{MapErrTrace, trace_error, trace_error_exit};
 use libimagentrylink::external::ExternalLinker;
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
It was a bug in the store which is addressed in #989

PR #997, opened 2017-07-15
Fix wrong lib name

matthiasbeyer:
The library was not named correctly.


matthiasbeyer:
Maybe we should also rework the module names here.

PR #1002, opened 2017-07-19
libimagentrylink/testing

matthiasbeyer:
Testing for libimagentrylink, including a quickcheck-based test which might run too long. Lets see what travis says.


matthiasbeyer:
Unfortunately, quickcheck pushes the build time (on beta in this case) from 12 minutes to 19, which is _not nice_.

@TheNeikos Do you have an idea how to disable a certain test in the test execution? The quickcheck test just takes too long, but having one is really nice (and maybe we will have more later).

matthiasbeyer:
Rebased, with the quickcheck in the history. Ready for merging.

Issue #1003, opened 2017-07-23
store::GlobStoreIdIterator returns Item = StoreId - but should return Result


matthiasbeyer:
Currently, the `GlobStoreIdIterator` returns `Item = StoreId`.

But because it does some transformation inside the `next()` function, we should really return `Result<StoreId>` instead. If we refactor the code to return this, the `GlobStoreIdIterator` can not be transformed into a `StoreIdIterator` anymore, which is a problem.

Maybe we should transform the `StoreIdIterator` also to return `Result<StoreId>` therefor, but this would cause us a lot of code to be re-written. We can use `libimagerror::iter::TraceIterator::trace_unwrap()` to unwrap or trace in the client code.

This is a major overhaul of one of the `libimagstore` APIs and should go _after_ the 0.3.0 release, of course.

matthiasbeyer:
Now that I think about it, all iterators in imag should return `Result<_>`, because all iterators somehow depend on the `StoreIdIterator` (if not implementing a wrapping type around some iterator from an external dependency).

We should takle this right after 0.3.0, because this is a major design flaw IMO. @asuivelentine If you have nothing to do in your free time, feel free to tackle this :smile: . I cannot help you, though.

PR #1004, opened 2017-07-23
Reimplement GlobStoreIdIterator

matthiasbeyer:
Related to #1003 

Does not close the issue mentioned above, because this only reimplements the iterator to fix buggy behaviour.

This PR is ready, though.

### DIFF
--- a/libimagstore/src/store.rs
+++ b/libimagstore/src/store.rs
@@ -1037,6 +1037,15 @@ mod glob_store_iter {
 
     use libimagerror::trace::trace_error;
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Now that I think about it, all iterators in imag should return `Result<_>`, because all iterators somehow depend on the `StoreIdIterator` (if not implementing a wrapping type around some iterator from an external dependency).

We should takle this right after 0.3.0, because this is a major design flaw IMO. @asuivelentine If you have nothing to do in your free time, feel free to tackle this :smile: . I cannot help you, though.

PR #1008, opened 2017-08-25
Reorganize code in subdirs

matthiasbeyer:
A porposal to move things into subdirs, to lessen the amount of things in the root repository.

@TheNeikos I'd love to have comments from you on this one!

PR #1019, opened 2017-08-26
Fix warnings by not allowing them anymore

matthiasbeyer:
This might get a bit lengthy. In this first patch, we deny all warnings, hopefully getting to a point where a whole imag build can be done with not one warning.

Patches for fixing the (now) compile errors will follow, feel free to contribute!

### DIFF
--- a/lib/core/libimagerror/src/lib.rs
+++ b/lib/core/libimagerror/src/lib.rs
@@ -18,6 +18,7 @@
 //
 
 #![deny(
+    dead_code,
     non_camel_case_types,
     non_snake_case,
     path_statements,
@@ -26,6 +27,7 @@
     unused_allocation,
     unused_import_braces,
     unused_imports,
+    unused_must_use,
     unused_mut,
     unused_qualifications,
     while_true,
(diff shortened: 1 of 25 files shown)


matthiasbeyer:
The travis error is not reproducible on my machine when building the corrosponding crate. @TheNeikos and @mario-kr can you reproduce this?

libimagtodo/src/task.rs
```diff
@@ -48,7 +48,7 @@ impl<'a> Task<'a> {
 
     pub fn import<R: BufRead>(store: &'a Store, mut r: R) -> Result<(Task<'a>, String, Uuid)> {
         let mut line = String::new();
-        r.read_line(&mut line);
+        try!(r.read_line(&mut line).map_err_into(TodoErrorKind::UTF8Error));
```
mario-kr: I reproduced the build fail.
line 35 in this file:
```
use error::{TodoErrorKind as TEK, MapErrInto};
```

Therefore, it should be like this (on all occasions):
```
try!(r.read_line(&mut line).map_err_into(TEK::UTF8Error));
```

matthiasbeyer: Thanks. Funny that I couldn't reproduce.

matthiasbeyer:
Will rebase after #1008 is merged, as this PR is easier to fix than #1008 IMO.

PR #1021, opened 2017-08-26
Use cargo feature to specify both version and path

matthiasbeyer:
Thanks to reddit user /u/TimNN for suggesting this.

---

I think whether setting it to "0.4.0" would actually make sense or not. @TheNeikos or @mario-kr -
 suggestions?

### DIFF
--- a/.imag-documentation/Cargo.toml
+++ b/.imag-documentation/Cargo.toml
@@ -15,75 +15,27 @@ homepage      = "http://imag-pim.org"
 
 [dependencies]
 
-[dependencies.libimagannotation]
-path = "../libimagannotation"
-
-[dependencies.libimagbookmark]
-path = "../libimagbookmark"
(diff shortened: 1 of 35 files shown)


mario-kr:
Definitely a nice change, I guess my prediction was somewhat right ;)

IMHO: I would not make a follow-up release for this change alone. (at least not "0.4.0").

matthiasbeyer:
No, no follow up release. Have a look at the diff. Using the 0.3.0 version string (aka the already released stuff) or the next release?

mario-kr:
If I understand the comment on reddit correctly, there's no need to bump the version to 0.4.0 unless you are about to publish.

matthiasbeyer:
Turns out we actually have to set "0.4.0" so cargo recognizes the internal dependencies.

PR #1029, opened 2017-08-28
All extensions as traits

matthiasbeyer:
Tracking PR for migrating all crates to use the extension-trait approach rather than wrapping store types.


matthiasbeyer:
This looks really good and seems to have no merge conflicts. I guess we can merge this as soon as CI succeeds!

PR #1031, opened 2017-08-29
libimagdiary: extension as trait

matthiasbeyer:
Everything as trait in libimagdiary, pls!

WIP.


matthiasbeyer:
Funny. First, rustc 1.21 tells me that "Serde" is not longer required and now all jobs fail because "Serde" is missing.

matthiasbeyer:
I guess we have to import #1032 here. Damn. I guess I will import #1032 to #1029 and rebase this one, once #1032 is finished.

PR #1037, opened 2017-08-31
Libimagerror/based on error chain

matthiasbeyer:
Tracking branch for the next major refactoring: Base error types on `error_chain` crate rather than own logic to be more powerful and also easily pass context information (which was a hack with our approach, but gets really easy and nice with `error_chain`).


matthiasbeyer:
Pushed as backup, not entirely ported yet.

matthiasbeyer:
I guess it may happen that rebasing this branch after a merge of #1029 will be... hard. But the other way round (integrating these changes into #1029) seems not better at all ... so I guess merging #1029 before this one is the easier way...

matthiasbeyer:
Integration of the branch compiles and tests run properly on my machine. Hopefully I did not screw up when resolving the conflicts.

Lets see what CI tells us ... and we'll merge as soon as CI succeeds.

PR #1038, opened 2017-09-01
Changelog

matthiasbeyer:
I'd like to see a short review from @asuivelentine and @mario-kr, if that's okay :smile:


mario-kr:
lgtm

PR #1041, opened 2017-09-01
Add type for representing recurrence

matthiasbeyer:
We need such a thing.

Or maybe @TheNeikos or @mario-kr knows a library that does such a thing?


matthiasbeyer:
I guess this thing should be based on `chrono::Duration` and we could build some iterators for increasing a basedate with them. I'm in the process of thinking this through.

matthiasbeyer:
This will be very well covered by [kairos](https://github.com/matthiasbeyer/kairos).

Issue #948, opened 2017-05-31
imag-habit: Habit tracker


matthiasbeyer:
Another idea for an imag-tool: A habit tracker.

- [x] Defining habits
  - [x] Iteration: Daily, weekly, monthly
  - [ ] With a "habit-question": "Did you drink enough today?" (user-defined)
- [x] One can "check off" a habit
- [ ] Generating of statistics 

[more to come]

---

Inspired by [Loop Habit Tracker](https://f-droid.org/repository/browse/?fdfilter=loop&fdid=org.isoron.uhabits)

PR #1042, opened 2017-09-01
libimaghabit: init

matthiasbeyer:
Partly solves #948

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ members = [
     "lib/domain/libimagbookmark",
     "lib/domain/libimagcontact",
     "lib/domain/libimagdiary",
+    "lib/domain/libimaghabit",
     "lib/domain/libimagmail",
     "lib/domain/libimagnotes",
     "lib/domain/libimagtimetrack",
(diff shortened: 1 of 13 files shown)


matthiasbeyer:
I cannot continue here before we have a decent way to calculate dates like "today and then every 7 days" or something like that.

Hopefully [kairos](https:://github.com/matthiasbeyer/kairos) will be in a usable shape soonish, so we can use the endless iterators from there.

matthiasbeyer:
Still missing:

- [ ] The `Habit` has a basedate which is used as base for incrementation. But the naming is a bit wrong here.
- [ ] The `Habit` has no increment member yet. We should add that
- [ ] Maybe `Habit` should get a function from where one can get the next instance which is not in the store yet. So the function basically gets the iterator from the basedate and the increment and finds the first date from the iterator where no store entry exists. Should be do-able, though a bit of a complex one.

matthiasbeyer:
This has stalled, but mostly because I want to get kairos ready because I would heavily rely on it in this tool.

matthiasbeyer:
Could be considered ready, IMO.

PR #1043, opened 2017-09-01
imag-link: positional args

matthiasbeyer:
Imag-link should work with positional arguments after this.


matthiasbeyer:
Waiting for #1044 and #1045

PR #1046, opened 2017-09-02
All extensions as traits imag todo

mario-kr:
Not really sure if this is what you wanted... but here it is.
Instead of a `Task` type which wrapped a `FileLockEntry`, it's a `Task` trait implemented for FLE.
The naming scheme has become rather verbose.

Nothing is tested, but it builds.


mario-kr:
wrong target branch

matthiasbeyer:
I could have moved it to another base.. but as you already opened a new PR,.. here we go.

PR #1047, opened 2017-09-02
imag-todo/libimagtodo: All extensions as traits

mario-kr:
Not really sure if this is what you wanted... but here it is.
Instead of a `Task` type which wrapped a `FileLockEntry`, it's a `Task` trait implemented for FLE.
The naming scheme has become rather verbose.

Nothing is tested, but it builds.

### DIFF
--- a/bin/domain/imag-todo/src/main.rs
+++ b/bin/domain/imag-todo/src/main.rs
@@ -61,7 +61,9 @@ fn tw_hook(rt: &Runtime) {
     let subcmd = rt.cli().subcommand_matches("tw-hook").unwrap();
     if subcmd.is_present("add") {
         let stdin = stdin();
-        let stdin = stdin.lock(); // implements BufRead which is required for `FileLockEntry::import_task_from_reader()`
+
+        // implements BufRead which is required for `Store::import_task_from_reader()`
+        let stdin = stdin.lock();
 
         match rt.store().import_task_from_reader(stdin) {
             Ok((_, line, uuid)) => println!("{}\nTask {} stored in imag", line, uuid),
@@ -101,11 +103,12 @@ fn list(rt: &Runtime) {
                         match fle.get_header().read(&String::from("todo.uuid")) {
                             Ok(Some(&Value::String(ref u))) => Some(u.clone()),
                             Ok(Some(_)) => {
-                                warn!("Header type error");
+                                error!("Header type error, expected String at 'todo.uuid' in {}",
+                                       fle.get_location());
                                 None
                             },
                             Ok(None) => {
-                                warn!("Header missing field");
+                                error!("Header missing field in {}", fle.get_location());
                                 None
                             },
                             Err(e) => {
--- a/bin/domain/imag-todo/src/main.rs
+++ b/bin/domain/imag-todo/src/main.rs
@@ -25,7 +25,6 @@ extern crate toml_query;
 #[macro_use] extern crate version;
 
 extern crate libimagrt;
-extern crate libimagstore;
 extern crate libimagerror;
 extern crate libimagtodo;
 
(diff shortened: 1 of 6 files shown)


lib/domain/libimagtodo/src/task.rs
```diff
@@ -29,31 +29,43 @@ use task_hookrs::task::Task as TTask;
...
 use error::{TodoErrorKind as TEK, MapErrInto};
 use result::Result;
 
 /// Task struct containing a `FileLockEntry`
-#[derive(Debug)]
-pub struct Task<'a>(FileLockEntry<'a>);
-
-impl<'a> Task<'a> {
+pub trait Task<'a> {
+    fn import_task_from_reader<R: BufRead>(store: &'a Store, r: R) -> Result<(Self, String, Uuid)>
```
matthiasbeyer: Could be moved to a `TaskStore` trait which is implemented on `Store`.

matthiasbeyer: In fact, all functions here could be moved to a `TaskStore` trait implemented on `Store`.

Then we can introduce a `Task` trait and functions for getting task-data in it (like `get_start_time()` and such) and implement it on FLE.

matthiasbeyer:
@mario-kr you are ready here?

mario-kr:
Yes, the task trait was made into a taskstore trait. the task trait as you have suggested it would be a bit overkill, as only the uuid is saved in the store, so querying a field via imag would require to start a taskwarrior process.

mario-kr:
Oh i guess someone should test if it still works before merging.

matthiasbeyer:
Could you `pull --ff-only` from https://github.com/matthiasbeyer/imag/tree/all-extensions-as-traits_imag-todo please?

mario-kr:
done.
Sidenote: As you have implemented the method `get_uuid`, you might as well use it in `imag-todo`.

mario-kr:
could we work with gh pull requests here?

matthiasbeyer:
You mean I create a PR for you to merge? Or you create a new PR for the binary? Either way: Yes, sure! But I'm off until tomorrow, so not today! :smile: :sleeping:

mario-kr:
There are already changes in the binary in this branch, so I was suggesting that you make a PR which I can then accept onto this one.

bin/domain/imag-todo/src/main.rs
```diff
@@ -61,17 +61,17 @@ fn tw_hook(rt: &Runtime) {
     let subcmd = rt.cli().subcommand_matches("tw-hook").unwrap();
     if subcmd.is_present("add") {
         let stdin = stdin();
-        let stdin = stdin.lock(); // implements BufRead which is required for `Task::import()`
+        let stdin = stdin.lock(); // implements BufRead which is required for `FileLockEntry::import_task_from_reader()`
```
matthiasbeyer: Documentation wrong.

bin/domain/imag-todo/src/main.rs
```diff
@@ -92,30 +92,34 @@ fn list(rt: &Runtime) {
...
-                    Ok(None) => {
-                        warn!("Header missing field");
-                        None
+            let uuids : Vec<_> = iter.filter_map(|storeid| {
+                match rt.store().retrieve(storeid) {
+                    Ok(fle) => {
+                        match fle.get_header().read(&String::from("todo.uuid")) {
+                            Ok(Some(&Value::String(ref u))) => Some(u.clone()),
+                            Ok(Some(_)) => {
+                                warn!("Header type error");
```
matthiasbeyer: We should also put in the error message which entry triggered the warning.

Also: Make them `error!()`s!

bin/domain/imag-todo/src/main.rs
```diff
@@ -92,30 +92,34 @@ fn list(rt: &Runtime) {
...
+                                None
+                            },
+                            Ok(None) => {
+                                warn!("Header missing field");
```
matthiasbeyer: same here.

matthiasbeyer:
@mario-kr please cherry-pick https://github.com/matthiasbeyer/imag/tree/all-extensions-as-traits_imag-todo onto your PR. No need for a new PR, I guess.

mario-kr:
I `git pull`ed your branch and pushed it back here.

PR #1061, opened 2017-09-08
Implement aliases for imag binary

matthiasbeyer:
Compiles, but not yet tested.

Anyone volunteering? @mario-kr ? :smile:


matthiasbeyer:
Works as far as I can see...

PR #1064, opened 2017-09-14
imag-diary: per diary timed config

matthiasbeyer:
The config file can be used to set a diary to create timed or non-timed entries.

Compiles and tested. But I would like to test it a bit more before merging, especially whether cli-overrides-config works.


matthiasbeyer:
> Compiles and tested. But I would like to test it a bit more before merging, especially whether cli-overrides-config works.

Yes, I can override this setting, but I cannot disable the setting from the commandline, which should be the way this should work IMO.

PR #1080, opened 2017-09-16
Do not use "imag" namespace in header

### LINKED ISSUE #1077: libimagentrytag uses wrong header location

matthiasbeyer:
Closes #1077


matthiasbeyer:
So basically we're running into a `EntryAlreadyBorrowed` error here. I wonder why that is.

PR #1111, opened 2017-09-23
libimagentrylink: Add helper to delete links based on predicate

matthiasbeyer:


matthiasbeyer:
It seems like we don't actually need this functionality, so why keep it?

PR #1121, opened 2017-09-30
Initial import of imag-mv

matthiasbeyer:


matthiasbeyer:
Moving basically works, though not into collections that do not exist yet.

Should be fixed before this gets merged.

PR #1130, opened 2017-10-12
Imag annotate/init

### LINKED ISSUE #687: imag-notes/libimagnotes: Annotations
Would be a good idea/feature to be able to

```
imag notes annotation --add example/entry
```

which does nothing more than create a note entry in the store, maybe named with a hash of the path of the entry to annotate. Then, the user can add a note (as usual) and afterwards, this note is linked to the `example/entry` via `libimagentrylink`.

Of course, other arguments for the `annotation` subcommand might be added:
- `--list` - list annotations for an entry (including some identifier for each, possibly the hash)
- `--delete` - delete annotations (by either selecting from a list or non-interactively by hash)
- `--show` - show annotations for an entry

possibly more...

---

The hash mentioned above should be unique for _each_ annotation. So using the `hash(path(file_to_annotate))` wouldn't be enough, as it would yield the same hash for two different annotations. The contents of the annotation is also not good, IMHO, as this would yield complex code when annotations should be immutable and an annotation should really be just the combination of `imag note` and `imag link`. So maybe the time when the annotation was created...

---

The args mentioned above might even be subcommands for the `annotation` subcommand themselves, each having arguments on their own.

---

The `annotation` subcommand is rather hard to type, so an alias should be provided. I would vote for `ann` or `anno`.

---

I guess this is not that hard to implement, as almost the complete infrastructure for this is already there, but it is not done in an eyeblink either.

---

@asuivelentine wanna have something not-so-small to work on? This could be something for you, if you want!

matthiasbeyer:
Waiting for #1129 to be merged, as this one is based on #1129.

Closes #687

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,7 @@
 [workspace]
 members = [
     "bin/core/imag",
+    "bin/core/imag-annotate",
     "bin/core/imag-gps",
     "bin/core/imag-grep",
     "bin/core/imag-link",
--- a/bin/core/imag-annotate/Cargo.toml
+++ b/bin/core/imag-annotate/Cargo.toml
@@ -1,0 +1,30 @@
+[package]
+name = "imag-annotate"
+version = "0.5.0"
+authors = ["Matthias Beyer <mail@beyermatthias.de>"]
+
+description = "Part of the imag core distribution: imag-annotate command"
+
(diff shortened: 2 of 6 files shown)


matthiasbeyer:
Tested on the commandline, guess that's enough for the first step. Ready to merge.

PR #1110, opened 2017-09-23
Libimagcontact/init

matthiasbeyer:
Initial draft for `libimagcontact`, for implementing `imag-contact` later.

PR #1132, opened 2017-10-12
Imag contact/init

matthiasbeyer:
Based  on #1110 and still WIP. I have not even tested it yet, but I'm ready to "announce" that I'm working on this module ...


matthiasbeyer:
This works so far, though the hash ref stuff does not yet work as intended. The content hash gets printed but the path hash is required to be able to fetch the entry later on.

matthiasbeyer:
Needs rebase after #1141

matthiasbeyer:
This works now.

@mario-kr and @asuivelentine would you like to play with it? I have not yet added an option to add a new contact, because I am not sure how that should work, especially where to place the generated `vcf` file. We do not create it in the store, of course... I guess the best idea would be to add a config option where to place new `vcf` files... maybe even a list of folders and a commandline switch which one to choose.

What do you think?

matthiasbeyer:
Should wait for #1146

matthiasbeyer:
Rebased and squashed everything.

PR #1163, opened 2017-11-25
libimagstore: More powerful iterator extensions

matthiasbeyer:
Implements more powerful iterators in `libimagstore` which extend the base `StoreIdIterator` with (for example) getter iterators.

Now these iterators are implemented on all `Iterator<Item = StoreId>` as well as on all `Iterator<Item = Result<StoreId, E>>` (`E` generic).

Thanks to [Shepmaster on stackoverflow](https://stackoverflow.com/questions/47487756/extending-all-iteratorsitem-resulttype-e-to-transform-type) for helping me with this.

---

**On hold** until 0.5.0 is released.

Lets see what the CI tells us about this PR.

### DIFF
--- a/lib/core/libimagstore/src/iter.rs
+++ b/lib/core/libimagstore/src/iter.rs
@@ -27,14 +27,13 @@ macro_rules! mk_iterator_mod {
         fun       = $fun:expr
     } => {
         pub mod $modname {
-            use storeid::StoreIdIterator;
             use storeid::StoreId;
             #[allow(unused_imports)]
             use store::FileLockEntry;
             use store::Store;
             use error::Result;
 
-            pub struct $itername<'a>(StoreIdIterator, &'a Store);
+            pub struct $itername<'a>(Box<Iterator<Item = StoreId>>, &'a Store);
 
             impl<'a> Iterator for $itername<'a> {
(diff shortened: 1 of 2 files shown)


lib/core/libimagstore/src/iter.rs
```diff
@@ -90,6 +176,50 @@ mk_iterator_mod! {
...
+#[cfg(test)]
+mod compile_test {
+
+    // This module contains code to check whether this actually compiles the way we would like it to
+    // compile
+
+    use store::Store;
+    use storeid::StoreId;
+
+    #[allow(dead_code)]
```
TheNeikos: You can add this to the whole `mod` declaration

lib/core/libimagstore/src/iter.rs
```diff
@@ -17,41 +17,121 @@
...
     {
         modname   = $modname:ident,
         itername  = $itername:ident,
         iteryield = $yield:ty,
         extname   = $extname:ident,
         extfnname = $extfnname:ident,
         fun       = $fun:expr
+    } => {
+        use storeid::StoreId;
+        #[allow(unused_imports)]
```
TheNeikos: Why not delete?

PR #1166, opened 2017-12-06
Dependency update: log 0.4.0

matthiasbeyer:
**This is work-in-progress.**

We will follow the release candidates of the `log` crate here until `0.4.0` is released. We will merge the changes _after_ our `0.5.0` release.

Lets see how the `0.4.0-rc.1` release works with our codebase.


matthiasbeyer:
Depending on when env logger will be released in 0.5.0 (see https://github.com/sebasmagri/env_logger/issues/43) we can merge or not this soonish or not... or so.

I'm too drunk for this right now.

PR #1182, opened 2017-12-24
libimagwiki: init

matthiasbeyer:
WIP


matthiasbeyer:
Removed the notion of an "mainpage" and implicitely create "index" therefore.

PR #1184, opened 2017-12-24
imag-log: init

matthiasbeyer:
depends on #1183 

WIP.


matthiasbeyer:
The `show()` implementation is still missing.

PR #1191, opened 2018-01-03
CLI interfaces redesign

matthiasbeyer:
This  is an effort to move the CLI interface implementation to the respective `libimag*` and minify the implementation of the related `imag-*` binary.

The idea is that we can use this to integrate functionality from different `imag-*` commands into eachother, for example:

```
imag diary create foobar and-link --to anotherentry and-annotate --text 'Hello world'
```

This effort is highly experimental, of course. Thus, I'm not adding this to a milestone. I'm not even sure whether we want such functionality, as the complexity of the crates increases (behind feature gates) and it might be complex to actually use... the tradeoff is

```
imag diary create foobar
imag link diary/foobar anotherentry
imag annotate diary/foobar --text 'Hello world'
```

vs. the above call.


matthiasbeyer:
This was a nice experiment, but I don't think we should integrate these patches/this functionality.

PR #1193, opened 2018-01-03
Rustc update

matthiasbeyer:
I guess this will fail now, because rustc 1.23.0 will be released tomorrow...


matthiasbeyer:
Builds stopped, will trigger build tomorrow or the day after tomorrow.

PR #1204, opened 2018-01-06
Add script to add a new crate

matthiasbeyer:
@manuelmessner could you please have a look? I don't see the error:

```
scripts/new-crate.sh: line 105: syntax error: unexpected end of file
```


scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
+#/usr/bin/env bash
```
mario-kr: not a shebang

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo -e ""
+    echo -e "Execute _only_ from the top level of the repository"
+    exit 0
+fi
+
+crate_type="$1"
+crate_location="$2"
+
+exists_cmd() {
+    command -v $1 || { echo "No $0 found"; exit 1; }
```
mario-kr: `$0` does not seem right, maybe `$1`?

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+}
+
+GIT=$(exists_cmd "git")
+CARGO=$(exists_cmd "cargo")
+
+{ cat ./Cargo.toml 2>/dev/null | head -n 1 | grep "[workspace]" >/dev/null; } || {
```
mario-kr: use `grep -q` instead of `/dev/null`

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+
+crate_type="$1"
+crate_location="$2"
+
+exists_cmd() {
+    command -v $1 || { echo "No $0 found"; exit 1; }
+}
+
+GIT=$(exists_cmd "git")
+CARGO=$(exists_cmd "cargo")
```
mario-kr: assignment not necessary, exists_cmd exits on failure, so you can just use the normal cargo and git commands.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+    exit 1;
+}
+
+if [[ -e "$crate_location" ]]; then
+    echo "Crate exists: $crate_location"
+    exit 1;
+fi
+
+IFS=/ read -ra crate_name_parts <<< "$crate_location"
+crate_name=$(echo ${crate_name_parts[-1]})
```
mario-kr: i'm pretty sure you dont need echo, just assign it.
```
crate_name="${crate_name_parts[-1]}"
```

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+    crate_description="Library for the imag core distribution"
+else if [[ "$crate_type" == "bin" ]]; then
+    crate_description="Part of the imag core distribution: $crate_name command"
+fi
+
+git_name="$($GIT config user.name)"
+git_email="$($GIT config user.email)"
+
+store="lib/core/libimagstore/Cargo.toml"
+crate_version=$(cat $store | grep version | head -n 1 | cut -d '"' -f 2)
```
mario-kr: use
```
grep version $store
```
instead of cat.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+
+echo "Crate type:        $crate_type"
+echo "Crate location:    $crate_location"
+echo "Crate name:        $crate_name"
+echo "Crate version:     $crate_version"
+echo "Crate description: $crate_description"
+echo "Crate author:      $git_name <$git_email>"
+
+exit 1
```
mario-kr: early exit? script is aborted this way.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+
+pushd "$(dirname $crate_location)"
```
mario-kr: you dont know if $crate_location exists yet, so a mkdir should be done first

mario-kr: nvm, dirname strips it, though the underlying path might not exist yet.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+[dependencies]
+
+EOS
+
+echo "Cargo.toml written. Please make sure that the README has the right path!"
+popd
+
+$GIT add ${crate_location}/*
+
+sed -i "$ s/]/    \"${crate_location}\",\n]/" Cargo.toml
```
mario-kr: this might be kinda dangerous/Cargo.toml breaking. And if you have to sort the list manually either way... lt the user handle it.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+
+# 1. Creates a new crate
+# 2. Adds the required crate meta information
+# 3. Sets the version of the crate to the same version as libimagstore
+# 4. Adds the crate to the top-level workspace
+
+if [[ "$1" == "-h" || "$1" == "--help" ]];
+then
+    echo -e "$0 [bin|lib] ./path/to/new/crate"
+    echo -e ""
```
mario-kr: i dont think `-e` is required anywhere here, also you can just use echo without any arguments for a newline.

scripts/new-crate.sh
```diff
@@ -0,0 +1,104 @@
...
+    echo -e "Execute _only_ from the top level of the repository"
+    exit 0
+fi
+
+crate_type="$1"
+crate_location="$2"
```
mario-kr: you should check if you actually got two arguments from the caller.

scripts/new-crate.sh
mario-kr: the flag `-m 1` will give you only the first match, so head is not necessary either.

mario-kr: if youre willing to use `rg`, you can do it like this:
```
rg -Nm 1 'version = "(.*)"' $store --replace '$1'
```

matthiasbeyer: Awesome suggestions. I will not use `rg` because I won't assume it is on the users system, but the other suggestions are nice, thanks!

scripts/new-crate.sh
matthiasbeyer: Yes, as it is still WIP I added this line so the script does nothing but print what it calculated. Will be removed later on.

scripts/new-crate.sh
matthiasbeyer: I'm planning on implementing the sorting some day.

It might break toml, but it is unlikely IMO, as the top-level Cargo.toml is really simple and we only append to it.

matthiasbeyer:
Development of this is integrated in #1219 ... will close this now in favor of the other PR.

PR #1213, opened 2018-01-11
Update dependency: toml-query -> 0.6

matthiasbeyer:
Not ready yet, as we really should use the new typed-reading functions to reduce code size on our side.


matthiasbeyer:
I'm more than sure that this will break the build now...

PR #1219, opened 2018-01-15
Minor patches

matthiasbeyer:
Long-living minor patches (mostly infrastructure patches) branch. Integrates #1204 for example.

To be merged right before the 0.6.0 release.


matthiasbeyer:
It is about time to merge this, there are already too much patches in here to keep it longer out of master.

PR #1224, opened 2018-01-31
Git version in version

matthiasbeyer:
Just an experiment to see how this affects build time...


matthiasbeyer:
@TheNeikos can you please have a short look? I fear that this blows up when releasing the next version (and includes the commit hash in the released binary, which I don't want)...

So I'd appreciate some review here.

matthiasbeyer:
Maybe we should use [built](https://crates.io/crates/built) for that!

matthiasbeyer:
Not maybe. We _should_ use the "built" crate for that - because `git describe` gives us what we want and thus we can fetch the imag version directly from the git version information!

matthiasbeyer:
This does not even work. I don't know why.

PR #1232, opened 2018-02-05
Doc overhaul

### LINKED ISSUE #1230: Introduction>The Problem: wrong statement related to the underlying situation
Hi,

On https://github.com/matthiasbeyer/imag/blob/master/doc/src/01000-intro.md the project states that:

> The problem imag wants to solve is rather simple. When the project was initiated, there was no PIM-Suite available which
> 
> - was for this domain of users ("power-users", "commandline users")
> - uses plain text as storage format
> - was scriptable
> - was modular
> - contained functionality to link content

I would like to challenge this bold statement for being utterly wrong.

Most prominent and probably most used example of this kind of software is Emacs [Org-mode](https://orgmode.org). The fact that you don't like ["the other editor"](https://imag-pim.org/doc/web/0.5.0/) (which is [not just an editor](http://karl-voit.at/2015/10/23/Emacs-is-not-just-an-editor/) at all) does not mean that you have to ignore the fact that there **are** solutions out there which perfectly fit to the valid requirements you describe above.

A second project that would fit to this set is [Zim](http://zim-wiki.org/) or many other instances of the whole class of [Personal Wiki](https://en.wikipedia.org/wiki/Personal_wiki)s.

Don't get me wrong: your project seems to be very cool and please try to make the best PIM system there is. However, you should be honest about your project so that people stumbling over [pages like this](https://imag-pim.org/doc/web/0.5.0/) don't get the wrong idea that there is only imag which tries to address the requirements mentioned.

This issue might seem to be nitpicking. Although for the underlying motivation, the basic idea, the situation in general you should not start on wrong statements.

matthiasbeyer:
Closes #1230

### DIFF
--- a/doc/src/01000-intro.md
+++ b/doc/src/01000-intro.md
@@ -10,7 +10,7 @@ generally read with big grain of salt.**
 If you have any objections, suggestions for improvements, bugs, etc, please file
 them.
 A way to reach out to the imag project maintainer(s) is described in the
-CONTRIBUTING file of the repository or in this document, on the appropriate
+CONTRIBUTING file of the repository or in this document, in the appropriate
 section.
 
 ## The Problem {#sec:intro:problem}
--- a/doc/src/01000-intro.md
+++ b/doc/src/01000-intro.md
@@ -15,35 +15,35 @@ section.
 
 ## The Problem {#sec:intro:problem}
 
-The problem imag wants to solve is rather simple. When the project was
-initiated, there was no PIM-Suite available which
+The problem this project tries to solve is to provide a modular commandline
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
I'm not sure whether there's something missing in this PR.

matthiasbeyer:
@novoid would you approve here? Any more things to clearify? 😄

novoid:
The notion that imag is the only tool for the defined set of requirements is gone and this was my main criticism at this passage. So: thumbs up from my side ;-)

matthiasbeyer:
Awesome, thank you!

PR #1256, opened 2018-02-06
Add TraceIterator::trace_unwrap_exit()

matthiasbeyer:
This iterator exits on the first Err(_)

Will not be in 0.6.0 ... we need to also use this before merging it.


matthiasbeyer:
> we need to also use this before merging it.

yeah... I really should not do late-night merges.

PR #1260, opened 2018-02-06
imag-mv: Fix links after moving entry

### LINKED ISSUE #1240: `imag-mv` does not fix links
Moving an entry in the store with `imag-mv` does not automatically fix the links from/to this entry.

This is a bug.

matthiasbeyer:
Closes #1240 

Not tested yet


matthiasbeyer:
Tested.

PR #1263, opened 2018-02-06
Do not print stats when listing files with matches

### LINKED ISSUE #1252: `imag grep -l` should only print list of files
`imag grep -l` does currently output the status:

```
Processed 1 files, 0 matches, 1 nonmatches
```

which it should not do.

Also we should have a flag for printing the complete filepath (like `-L` for printing `/home/$USER/.imag/store/....`).

matthiasbeyer:
Closes #1252


matthiasbeyer:

PR #1267, opened 2018-02-06
Do not allow linking with entry itself

### LINKED ISSUE #1250: `imag link foo foo` fails
Linking an imag entry to itself with `imag link` fails with 

```
ERROR[   1]: relative URL without a base
link exited with non-zero exit code
```

Which should not happen. 

---

This is misuse of the `imag-link` tool and I consider this not-so-severe, hence I'm postponing this bugfix to the next release rather than 0.6.0.

matthiasbeyer:
Without this check, linking an entry with itself yields the following
error:

    ERROR[   1]: Entry is already borrowed: StoreId { base: Some("/home/m/.imag/store"), id: "notes/test" }
    ERROR[   2]: Error when calling retrieve() -- caused:
    ERROR[   3]: Error when calling get()

Which is semantically correct, but the user may get confused by that.

Instead, we print a nice error message that the entry cannot be linked
to itself.

This is not fixed in libimagentrylink itself, because libimagentrylink
cannot be called for the same entry.
If this would be possible, we would pass two `Entry` objects
mutably to the link functionality routines. This is not possible with
Rusts borrow semantics and therefor yields above error.

---

Closes #1250

### DIFF
--- a/bin/core/imag-link/src/main.rs
+++ b/bin/core/imag-link/src/main.rs
@@ -61,6 +61,7 @@ use libimagrt::runtime::Runtime;
 use libimagrt::setup::generate_runtime_setup;
 use libimagstore::error::StoreError;
 use libimagstore::store::FileLockEntry;
+use libimagstore::storeid::StoreId;
 use libimagutil::warn_exit::warn_exit;
 use libimagutil::warn_result::*;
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:

matthiasbeyer:
tested. Works.

PR #1270, opened 2018-02-07
License header checker

### LINKED ISSUE #1269: CI script for checking license header
#1268 shows that it would be a good idea to add a script which is also executed by CI which checks all files for the license header and also whether the year numbers are there.

matthiasbeyer:
Closes #1269 

Should fail, but succeed when rebased with #1268 merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ addons:
 
 script:
   - |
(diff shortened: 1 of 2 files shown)


matthiasbeyer:

PR #1272, opened 2018-02-08
Prepare 0.6.0 release

matthiasbeyer:
* Add new libraries/binaries to release script


matthiasbeyer:
The v0.6.0 tag does not build on travis because of c229d7d which was necessary to release the crate. I will push a fix shortly (if possible).

PR #1274, opened 2018-02-10
libimagerror: Refactor

matthiasbeyer:
A complete refactoring of the `libimagerror` crate.


matthiasbeyer:
The changes in this branch introduces some inconvenience in the binaries, but reduces the size if libimagerror and simplifies a lot in there.

If the error output is better after these changes, I'd rather merge this and have some inconveniences in the binaries (which are rather trivial themselves, so it does not hurt so much).

matthiasbeyer:
Waiting for #1282 to land

PR #1276, opened 2018-02-10
libimagerror: weightloss

matthiasbeyer:
Refactor libimagerror with some weightloss. Might conflict with #1274 and maybe we can integrate these patches into #1274


matthiasbeyer:
#1274 reduced this down to one patch. This patch was moved to #1279, so this can be closed.

PR #1283, opened 2018-02-11
imag-timetrack: Add kairos support in "list" subcommand

matthiasbeyer:
This patch adds kairos support in the "list" subcommand for the "-f" and
"-t" parameters which limit the entries to show.

Something like

    imag timetrack list --from yesterday

is now possible.


matthiasbeyer:
Just tested this. Works.

PR #1285, opened 2018-02-13
libimagentryref: Rewrite

### LINKED ISSUE #1284: Rewrite libimagentryref
We need to rewrite `libimagentryref` because its purpose changed slightly in the past months of development.

# Purpose / Incentive

`libimagentryref` provides "references" to files on the filesystem. A reference might be used by a domain library to refer to data outside of the store (like libimagcontact refers to vcard files without putting them into the store). A reference might not be only a "reference entry" in the store but rather a more complex entry which _contains_ a reference. Thus, libimag_entry_ref is the right name for this library and it should only provide a building block for the actual entry.

# API Requirements

- [x] A way to generate reference names (for example a SHA1 hash) which can then be used to build an actual `StoreId` from, under a custom collection. So basically a `UniqueFileIdGenerator`.
- [ ] A way to re-find files via a custom `FileFinder` that implements its own filesystem traversal (and can thus filter for certain filetypes or even whether something is actually a file or a directory)
- [ ] A way to get the actual file with automatic integrity checking (whether it still matches the `UniqueFileId` which was stored for the file).

# Impact

Currently the following crates depend on `libimagentryref` and need probably to be rewritten for this:

- [ ] `imag-ref`
- [x] `libimagcontact`
- [ ] `imag-contact`
- [x] `libimagmail`
- [x] `libimagentrymarkdown`

It will also influence how `libimagcalendar` will be implemented.

matthiasbeyer:
The branch is named "refactor", but it is rather a complete rewrite of `libimagentryref`.

Closes #1284


matthiasbeyer:
Things are more complicated than I thought. I guess the libraries using this library will need a rewrite themselves because of these changes here. I guess this branch will become a rather long-living thing.

I will implement the refactorings for the libraries using this functionality as PRs onto this PR because of its scope.

PR #1289, opened 2018-02-14
libimagentryref: Add default generators

matthiasbeyer:
Default generators which use Sha functions to hash contents of files referenced.


lib/entry/libimagentryref/src/lib.rs
```diff
@@ -52,3 +52,16 @@ pub mod error;
...
+    feature = "generators-sha1",
+    feature = "generators-sha224",
+    feature = "generators-sha256",
+    feature = "generators-sha384",
+    feature = "generators-sha512",
+    feature = "generators-sha3",
+))]
+extern crate crypto;
+
+#[cfg(feature = "generators")]
```
matthiasbeyer: We should check for

```
    feature = "generators-sha1",
    feature = "generators-sha224",
    feature = "generators-sha256",
    feature = "generators-sha384",
    feature = "generators-sha512",
    feature = "generators-sha3",
```

here as well.

PR #1292, opened 2018-02-14
Fix broken pipes

### LINKED ISSUE #1291: Piping imag output panics
We use `print!()` and `println!()` a lot.

But that panics when piping, for example:

```
$ imag store ids  | head -n 1
thread 'main' panicked at 'failed printing to stdout: Broken pipe (os error 32)', src/libstd/io/stdio.rs:692:8
note: Run with `RUST_BACKTRACE=1` for a backtrace.
imag-store exited with non-zero exit code
```

Thus, we should rewrite the appropriate parts to use `write!()` and handle errors properly.

matthiasbeyer:
Closes #1291

Some refactoring involved.

### DIFF
--- a/bin/core/imag-annotate/src/main.rs
+++ b/bin/core/imag-annotate/src/main.rs
@@ -43,6 +43,7 @@ extern crate libimagerror;
 extern crate libimagstore;
 extern crate libimagutil;
 
+use std::io::Write;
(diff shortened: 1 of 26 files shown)


matthiasbeyer:
The "core" commands are not yet fixed.

PR #1295, opened 2018-02-19
libimagcalendar: init

matthiasbeyer:


matthiasbeyer:
depends on #1313

matthiasbeyer:
Note for myself:

The `Calendar` trait now provides functionality to get `Calendar::entries()` which automatically creates `calendar/entry/<UID>` files (to be tested, also if there is no UID, what happens and how to improve?).

These can be used via the `Entry` trait.

It must be tested whether it works as expected because the `Entry` stores the UID of the event it refers to and parses the file on `Entry::get_description()` (for example).

matthiasbeyer:

PR #1310, opened 2018-02-24
Update uuid to 0.6

Dylan-DPC:
Hey

Updating this crate to 0.6 since the new release is out. There are no further changes required.

### DIFF
--- a/lib/domain/libimagtodo/Cargo.toml
+++ b/lib/domain/libimagtodo/Cargo.toml
@@ -20,8 +20,8 @@ is-it-maintained-open-issues      = { repository = "matthiasbeyer/imag" }
 maintenance                       = { status     = "actively-developed" }
 
 [dependencies]
-task-hookrs = "0.4"
-uuid = "0.5"
+task-hookrs = { git="https://github.com/matthiasbeyer/task-hookrs.git", branch="master" }
+uuid = "0.6"
 toml = "0.4"
 toml-query = "0.6"
 log = "0.4.0"


matthiasbeyer:
Hey, thanks for you PR, this is awesome!

Unfortunately this breaks the build as `task-hookrs` uses uuid `0.5`. I will update `task-hookrs` [with this PR](https://github.com/matthiasbeyer/task-hookrs/pull/55) and after that we can push a commit here which updates the `task-hookrs` dependency.

matthiasbeyer:
Hm, I don't want to release task-hookrs 0.5.0 just because of a dependency update. But I guess its okay to depend on task-hookrs master branch, at least until we get the 0.7.0 release of imag ready, so if you could push a commit which changes that dependency, I guess that would fix this PR. :+1:

Dylan-DPC:
Alternatively you could release a 4.1 since I don't see any breaking change.

matthiasbeyer:
A dependency update were types from the dependency are exported is a breaking change.

Dylan-DPC:
Agreed. but if you are using your dependency and the way you use it doesn't change then it can't be considered breaking. However it is difficult to consider whether it is breaking or not its fine I will hook it to master but hopefully no change in master breaks the crate

lib/domain/libimagtodo/Cargo.toml
```diff
@@ -20,7 +20,7 @@ is-it-maintained-open-issues      = { repository = "matthiasbeyer/imag" }
 maintenance                       = { status     = "actively-developed" }
 
 [dependencies]
-task-hookrs = "0.4"
+task-hookrs = { git="https://github.com/matthiasbeyer/task-hookrs.git" branch="master" }
```
matthiasbeyer: There's a `,` missing before the `branch` key.

matthiasbeyer:
Awesome! Feel free to squash all your changes into one commit! If you do this, please provide a commit message that explains your changes! <3

Dylan-DPC:
@matthiasbeyer all good :+1:

PR #1311, opened 2018-02-25
Impl IntoStoreId for everything that implements Into<String>

matthiasbeyer:
@TheNeikos is this idomatic Rust? I'm not that confident with this patch...


TheNeikos:
The style itself is idiomatic. So I'd say you are fine on that regard

matthiasbeyer:
Yes, but semantically that might be a bad idea, right? Because not everything that can be a `String` can be a `StoreId`...

matthiasbeyer:
Naaah, this breaks existing interfaces too much and is semantically not completely nice. I let it be.

PR #1313, opened 2018-02-26
libimagstore: Add WithStoreIterator trait

matthiasbeyer:
This patch changes the store interface quite a bit:

Before, we had the possibility to create a `StoreIdIteratorWithStore`
by calling `Store::entries()`.

This was nice. The `StoreIdIteratorWithStore` had functions to transform
it into a `StoreCreateIterator` for example.

With this change, we have an abstraction so that _every_ iterator which
holds a reference to the `Store` can be transformed into a
`StoreCreateIterator` (for example). The iterator object has to
implement the `WithStoreIterator` trait for this, which is quite simple.

The `StoreIdIteratorWithStore` iterator type was changed to not provide
the `into_create_iter()` (for example) functions via the type but the
trait,... which changes the interface of `libimagstore`.

This change was made because we ran into the following problem: We
wanted to wrap a `StoreIdIterator` for filtering for a specific
collection in the store ("/calendar", to be specific).
This was possible, of course. But we cannot use this new iterator type
(`CalendarIter`) and call `into_get_iter()` on it.
We would need to implement it directly on the type.
With this change, we get this functionality by implementing the much
less complicated `WithStoreIterator` trait.


matthiasbeyer:
@TheNeikos please have a look... I cannot figure out this lifetime error!

lib/core/libimagstore/src/iter.rs
```diff
@@ -222,3 +222,81 @@ mod compile_test {
...
+            let store = self.store();
+            StoreCreateIterator::new(Box::new(self), store)
+        }
+    }
+
+    pub trait IntoDeleteIter<'a> {
+        fn into_delete_iter(self) -> StoreDeleteIterator<'a>;
+    }
+
+    impl<'a, I: 'static> IntoDeleteIter<'a> for I
```
matthiasbeyer: From what I understand, the problem is that `I` must have a `'static` lifetime because the compiler thinks that it lives too long.

I don't see why that is, because `I` is packed into a `StoreCreateIterator` object and thus may not outlive anything, right? @TheNeikos any ideas?

matthiasbeyer:
Stalled for too long and we can work without this ergonomic quite well. So I'm closing this now. Unmerged.

Issue #956, opened 2017-06-05
Imag command chaining


matthiasbeyer:
I'm heavily thinking about redesigning imag a bit to be able to do command-chaining (with pipes as one does with the shell), so that things like

```bash
imag entry new | imag tag add foo | imag category add baz | imag entry add-content --from /tmp/text | imag entry write --to /some/store/place
```

are possible.

I'm thinking how to implement this without having to change a lot of the codebase

# Store support for this

We could easily implement this with the store, as far as I know. This would then happen using a known structured data format for data passing. I would use JSON for this. 

## Store interface changes

The basic idea is to change the `create()`, `delete()`, `get()` and `update()` function to not get a `StoreId` but a `EntryIdentifier` which is

```rust
enum EntryIdentifier {
   Id(StoreId)
   StdIn // for `update` and `delete`
   StdErr // if needed
   StdOut // For `get` and `create`
```

`StoreId` would `impl Into<EntryIdentifier>`, so the current use of the store won't change at all (maybe a `into()` call here and there).

With a non-`EntryIdentifier::Id(_)` passed, the store would use the `stdin`/`stdout` to read/write entries.

The data format which is proposed below could be processed by the already implemented `imag-*` tools without much modification. The `Store::*()` calls would be altered a bit. A tool can also easily check whether it should read from stdin/write to stdout: if the `--id`/`--path` argument is not passed (the argument which defines which store entry should be altered), the data is passed via pipe.

## JSON for passing objects around

The JSON format would be pretty simple:

```json
{
    "version": "0.3.0",
    "pipe": [ {
            "module": "libimagentry",
            "call": "imag-entry create"
        },
        {
            "module": "libimagentry",
            "call": "imag-entry create"
        }
    ],

    "entries": [
        {
            "header": "// header goes here. TOML <=> JSON is possible AFAIK",
            "content": "// content... "
        }
    ],
}
```

### Explanation of the fields

The `version` field could be used for knowing which version of the store is required to be able to process this structure correctly.

The `pipe` field just popped into my mind when writing this - it could be used so we know what transformed the entries already.

The `entries` field then contains a list of entries as JSON objects.

### Intermediate store

A variant of the above data format would be that the `entries` field does not contain a JSON-Array, but a JSON-object, where the keys are (baseless) `StoreId`s:

```json
{
    "entries": {
        "/some/id": { "header": "", "content": "" },
        "/some/other": { "header": "", "content": "" }
    }
}
```

[...]

PR #1315, opened 2018-02-26
libimagrt: provide io

matthiasbeyer:
First step for #956

### DIFF
--- a/bin/core/imag-annotate/src/main.rs
+++ b/bin/core/imag-annotate/src/main.rs
@@ -135,7 +135,6 @@ fn remove(rt: &Runtime) {
 }
 
 fn list(rt: &Runtime) {
-    let mut out     = ::std::io::stdout();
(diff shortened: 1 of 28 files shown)


matthiasbeyer:
I wonder whether it would be a good idea to implement `Write` on `Runtime` so that one can `write!(rt, "foo")`. Otherwise we need a way to get a `&mut Write` from `&Runtime`. @TheNeikos any ideas?

matthiasbeyer:
Maybe one idea would be to provide a proxy object which either holds a `Stderr` or a `Stdout` and implements `Write` itself... then we return a proxy object from `Runtime::stdout()` and create it, depending on `atty::is(Stdout)` with `Stdout` or `Stderr`... so that the `Runtime` object itself never holds a `Stdout` itself.

Would result in more runtime overhead, though.

PR #1327, opened 2018-03-09
Add util macro for extending types with traits without boilerplate

matthiasbeyer:


matthiasbeyer:
@TheNeikos do you see anything bogus here?

(WIP because we should actually use it)

matthiasbeyer:
I'm against this after some more thinking. Its just to complex to wrap ones head around, from a newcomer point of view.

PR #1328, opened 2018-03-09
libimagstore internal cache redesign

matthiasbeyer:
Ideas and hopefully also implementations related to #956 

@TheNeikos This is complicated, but maybe you have some ideas?


matthiasbeyer:
### Problem

Right now we have the notion of a "cache" and a "backend" in the store.

### Approach

I think that has to be rewritten first: We don't need both. The "backend" can do the caching on its own, the `Store` may never know of "caching".

The `backend` should be loadable from stdin and be convertible into a backend which writes to FS (and the other way round).

The `backend` may do caching of `Entry` objects. I guess we have to rewrite the whole `file_abstraction` module in `libimagstore` for this.

matthiasbeyer:
These first patches refactored the store so that the `Cache` actually holds the `Backend` implementation and all things are accessed via the `Cache`.

matthiasbeyer:
Note to self: The latest CI error is due to borrowing inside the store. With `Cache::insert` we create a `Borrowed` entry, but sometimes it needs to be created `Present` and sometimes `Borrowed`  ... we should add a `Cache::mark_entry_borrowed_status(StoreId, PresentStatus)` or something like that.

matthiasbeyer:
Ummm... why does this fail?

```
error: test failed, to rerun pass '--bin imag-link'
```

Tells me nothing...?

matthiasbeyer:
Okay, this works now.

But I doubt the whole approach. The whole Store/Cache/Backend structure is just so _messy_. Maybe we should rewrite the whole thing (as in everything below the `Store` interface).

The `Cache` should hold `StoreId -> Entry`, that would make it easily de/serializeable. But right now it holds `StoreId -> StoreEntry`, which makes the whole thing messy as hell to be de/ser.

matthiasbeyer:
This whole thing does not work. I tried to implement piping in #1384 but it does not work as expected and the code is ugly as hell.

I'll let it be.

PR #1331, opened 2018-03-10
libimagentryref: content ref

matthiasbeyer:
## Problem

Right now we have the problem that we cannot refer to a single entry in a file. For example, if we have a "icalendar" file which contains 2 events, a reference points to the _file_ rather to the actual entry.

## Idea

The idea is that we create a "content ref", a reference which point to an specific "entry" inside the file.
The normal reference would be created like normal, but with the `ContentRef` we create more information.

## Implementation details

The following implementation details are not yet figured out:

### How to store

Maybe we can create a `{ref}-{content-ref}` filename from it, but this would imply that neither the "ref" nor the "content-ref" contains the seperator-character.
What would be more flexible: the "ref" part will be the directory and the "content-ref" part will be the filename.
But then we cannot create a ref to the file without any content ref (or we create "{ref}/file", but that would imply that "content-ref" can never be "file".

Maybe the way to go is:

* If there is only a "ref", we create a file at "{ref}/file"
* If there is a "ref" and a content-ref, we create a file at "{ref}/content-{content-ref}"

which would be the simplest way to go.

---

This is required for `imag-calendar` to work.

### DIFF
--- a/lib/entry/libimagentryref/Cargo.toml
+++ b/lib/entry/libimagentryref/Cargo.toml
@@ -25,6 +25,7 @@ log = "0.4.0"
 toml = "0.4"
 toml-query = "0.6"
 error-chain = "0.11"
+either = "1.4"
 
 libimagstore     = { version = "0.8.0", path = "../../../lib/core/libimagstore" }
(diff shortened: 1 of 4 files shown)


matthiasbeyer:

PR #1333, opened 2018-03-12
Fix: use Path::file_stem() instead of Path::file_name()

matthiasbeyer:
Because we don't want to know the extension of the filename here.


matthiasbeyer:
No backport required as this particular bug was not in 0.6.0.

PR #1348, opened 2018-03-19
Provide Ref::make_ref() for making a ref out of an existing entry

matthiasbeyer:
Extracted from #1295 

Blocked until we tested #1295 and know it works as expected.


matthiasbeyer:
Changelog added in minor branch.

PR #1355, opened 2018-03-22
Fix: Store::create() should fail if the entry exists

### LINKED ISSUE #1352: Store::create overrides existing entries.
`imag habit create` executed with the same parameters on another date results in the latter overriding the former call.

matthiasbeyer:
Closes #1352


matthiasbeyer:
Changelog added to minor branch.

matthiasbeyer:
IMO ready, but needs testing.

PR #1357, opened 2018-03-22
Libimaghabit/create vs retrieve

matthiasbeyer:
Semantics were bogus here.


matthiasbeyer:
We should properly test this (on the CLI) before merging.

matthiasbeyer:
Works so far.

PR #1326, opened 2018-03-09
Doc

matthiasbeyer:
Documentation cleanup and adding of a "how to write imag modules"-guide.

PR #1359, opened 2018-03-22
Add guide on how to write modules

matthiasbeyer:
Extracted from #1326 because this needs more work.


matthiasbeyer:
I fear that we don't continue much here, so rather merge as-is rather than let this be stalled for too long.

PR #1361, opened 2018-03-22
Remove "build" job to speed up travis builds

matthiasbeyer:
I'm not entirely sure, but I _think_ that actually doing `cargo test --all --all-features` is enough... we don't need to do `cargo build` as well.

Still waiting on more insight from [here](https://users.rust-lang.org/t/is-it-safe-to-do-only-cargo-test-in-my-travis-jobs/16369/3)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,6 @@ script:
   - |
     bash ./scripts/find-dead-symlinks || exit 1
     bash ./scripts/license-headers-updated || exit 1
-    cargo build --all --all-features -j 1 || exit 1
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Not merging this because of discussions in linked thread.

matthiasbeyer:
As @steveklabnik explained in the linked users.rust-lang.org thread, `cargo test` should be enough, so I reopened this PR.

I would say this is a rather small improvement in build time, but it is an improvement (about ~2 minutes compared to an PR filed at the same time). Either way, it does not hurt!

PR #1368, opened 2018-03-23
imag-link: Do table output by default

### LINKED ISSUE #1243: `imag-link list` output refactoring
I think we should have a `prettytable-rs` output from `imag-link list` and an option (`--plain`) to list links without counter.

matthiasbeyer:
Closes #1243


matthiasbeyer:
Changelog added to minor branch, tested.

PR #1373, opened 2018-04-03
libimagrt: Handle unknown subcommand

matthiasbeyer:
Not yet integrated.


matthiasbeyer:
Tested with an executable bash script `imag-store-foobar` in `$PATH` and calling `./target/bin/imag-store foobar`.

Works.

PR #1379, opened 2018-04-07
libimagstore: Make iterator lifetimes less restricting

matthiasbeyer:


matthiasbeyer:
I guess this will break everything else.

PR #1381, opened 2018-04-07
imag-calendar: init

matthiasbeyer:
This is the first step for a `imag-calendar` command.

Reading calendars should already work.

---

This is currently based on `libimagcalendar` (#1295), which is not yet merged.


matthiasbeyer:
I just tried to import my personal calendars and I found that the `vobject` crate is _far_ from complete, as it does not even handle recurring events correctly.

I guess `vobject` needs a complete rewrite first.

matthiasbeyer:

PR #1382, opened 2018-04-08
Fix: Editor commands should be split at whitespace

matthiasbeyer:
This fixes the following problem:

If the editor setting was "vim " instead of "vim", the editor was called
with `"vim" " "`, which resulted in unexpected behaviour.

The patch fixes this.


matthiasbeyer:
Changelog in minor branch.

PR #1398, opened 2018-04-14
Add shell completion via build script (again)

### LINKED ISSUE #1211: Shell completion for imag binary itself
Currently, the bash completion for the `imag` binary itself only knows of the flags created via clap.

We have no way to complete the following:

```
imag --rtp /tmp diary <tab>
```

because this would need the script to call the completion script of `imag-diary` in this case.

Maybe @mario-kr or even @manuelmessner (our shell/bash guru) knows a solution for this.

Completion scripts can be generated via `<imag tool> --generate-commandline-completion <shell>` where `shell` is one of `bash`, `fish`, or `zsh`.

mario-kr:
closes #1211


mario-kr:
PR #1011 should be considered here, as the original build script was removed there for reasons which might still be relevant

PR #1418, opened 2018-04-19
Rewrite libimagentryview interface

### LINKED ISSUE #1390: libimagentryview uses println
We still use `println!()` in viewers.

That should change asap, but I do not consider it a real bug, so it can wait until after the next release.

matthiasbeyer:
In the previous versions, the sink (where the entries should be written
to) was not passed.

This did conflict with the libimagrt holding the stdout/stderr handles,
because it automatically writes to stdout (which we don't want to do in
some cases).

Passing the sink is way nicer. This patch changes libimagentryview so
that the sink is passed to the viewer.

---

Closes #1390

### DIFF
--- a/bin/core/imag-view/src/main.rs
+++ b/bin/core/imag-view/src/main.rs
@@ -185,6 +185,9 @@ fn main() {
             viewer.wrap_at(width);
         }
 
+        let output      = rt.stdout();
+        let mut lockout = output.lock();
+
         entry_ids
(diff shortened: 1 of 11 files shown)


matthiasbeyer:
Before merge: Outsource c7bf036 to another PR.

matthiasbeyer:
Contains the patch which was submitted in #1443 as well, waiting for that.

PR #1442, opened 2018-04-24
imag-view: markdown compile

### LINKED ISSUE #1439: imag-view should be able to compile markdown
We could use `mdcat` to print markdown in the terminal properly formatted, if `imag view compile` is called.

This is merely an idea.

matthiasbeyer:
Closes #1439


matthiasbeyer:
Should wait on #1418 and be rebased and merged afterwards.

PR #1443, opened 2018-04-24
Impl AsRef<Entry> for FileLockEntry

matthiasbeyer:
@TheNeikos does this make sense? Semantically I mean?


matthiasbeyer:
After some discussion with TheNeikos, we think this does not make sense.

PR #1450, opened 2018-04-25
libimagcontact: not based on libimagentryref

### LINKED ISSUE #1448: imag contact pathes are not necessarily correct
When importing contacts into imag, imag-contact creates a header field "path", which points to the file containing the vcf data.

But this filepath is not necessarily correct on all devices. For example, when syncing the contact file to another device using `vdirsyncer`, it gets recreated with another filename on that other device.

So the `path` is not correct on the other device.

---

~~Possible fix~~

~~Let `path` only point to the directory where the file is located and find the right file using the `UID` of the contact, which we also store in the imag entry. This results in more runtime processing, but I see no other way right now.~~

---

Further: As we use `libimagentryref` below `libimagcontact`, which creates the `path` header variable, we cannot fix this in `libimagcontact`.

matthiasbeyer:
This rewrites `libimagcontact` to be not based on `libimagentryref`.

And it rewrites `imag-contact` to work with the new interface.

---

Closes #1448


matthiasbeyer:
From my manual testing everything seems to work.

PR #1457, opened 2018-04-27
Minor

### LINKED ISSUE #1455: Add an example of use to the readme
Currently, the readme tells me the idea behind the tool which is great, but doesn't show me what using the tool looks like.

matthiasbeyer:
Closes #1455


matthiasbeyer:
That mutt helper needs an fix for contacts with multiple email adresses: It seems that mutt wants (as a third parameter from the TSV output) the vcard-email-parameters (like "HOME", "PREF").

That is not yet handled by the jq-mutt helper, but support for this was merged in #1459 .

matthiasbeyer:
I guess we should merge this now.

PR #1471, opened 2018-04-30
libimagstore: Do not collect in Iterators

matthiasbeyer:
Refactors the iterator API (and changes its interface).

The iterators did some `collect()` calls internally. With these patches, they do not do this anymore.

Ready, but the binaries are not yet refactored for this and thus might fail to compile.


matthiasbeyer:
dd22509 is also cherry-picked to the minor branch.

matthiasbeyer:
Note to self: Remove the latest two commits before merging this. The debug output adding commit has nothing to do with this PR, the other commit is in the minor branch.

matthiasbeyer:
Well, we still need the "Remove Store::reset_backend()" patch because fixing this first and then removing it is too much stress for me.

PR #1478, opened 2018-04-30
imag-diagnostics: Add a progressbar to the command

matthiasbeyer:


matthiasbeyer:
Hitting https://github.com/Amanieu/parking_lot/issues/72 here, need to wait for rustc 1.26, so we support 1.24+1.25+1.26 then.

PR #1502, opened 2018-08-11
Optimize backend impl to not hold open files

matthiasbeyer:
This patch changes the filesystem-backend implementation of libimagstore
to open files on each read/write rather than holding the file handle in
memory at all times.

Whenever a lot of imag store entries are read into memory, the imag
process may ran out of file descriptors. With this patch applied, a
`Store::get()` call on an entry which is not yet in the store cache
would cause the file to be read, but the FD being dropped after that.
Likewise, a `Store::update()` (which is also called if the imag entry is
dropped) would re-open the file on the filesystem and write the contents
from the imag store cache back to the file.

With this patch, opening hundrets or thousands of imag entries should be
no problem anymore, only the available memory should be a limit then.

Signed-off-by: Matthias Beyer <mail@beyermatthias.de>

---

@mario-kr and @asuivelentine can you please build one or two imag commands with this patch and test them (CLI)? Just want to make sure I don't have forgotten some details about this patch :smile: 

Suggestion: Build `imag-ids`, even patch out the store cache flushing (see below) and test it on a store with more than 100 entries.

If you report success, I will apply the patch below with "Tested-by"-trailer to this PR.

```diff
diff --git a/bin/core/imag-ids/src/main.rs b/bin/core/imag-ids/src/main.rs
index c8da3ceff..ec2297367 100644
--- a/bin/core/imag-ids/src/main.rs
+++ b/bin/core/imag-ids/src/main.rs
@@ -92,13 +92,6 @@ fn main() {
         .map_err_trace_exit_unwrap(1)
         .trace_unwrap_exit(1)
         .enumerate()
-        .map(|(i, e)| {
-            if i % 100 == 0 {
-                let _ = rt.store().flush_cache();
-            }
-
-            e
-        })
         .filter(|id| collection_filter.filter(id))
         .filter(|id| match query_filter.as_ref() {
             None     => true,
```

### DIFF
--- a/lib/core/libimagstore/src/file_abstraction/fs.rs
+++ b/lib/core/libimagstore/src/file_abstraction/fs.rs
@@ -32,10 +32,7 @@ use storeid::StoreId;
 use file_abstraction::iter::PathIterator;
 
 #[derive(Debug)]
-pub enum FSFileAbstractionInstance {
-    Absent(PathBuf),
-    File(File, PathBuf)
-}
+pub struct FSFileAbstractionInstance(PathBuf);
 
 impl FileAbstractionInstance for FSFileAbstractionInstance {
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
I will merge this now, we can always roll back...

